PR #8, opened 2022-02-21
Fixes IQ Skill Restriction table length

theCapypara:
According to SkyTemple's previous XML, this field should be 0x89 in size. Are you sure 0x8C is correct? If it is, what are the rest of the bytes for?

### DIFF
--- a/symbols/arm9.yml
+++ b/symbols/arm9.yml
@@ -804,8 +804,9 @@ arm9:
         NA: 0x20A1A5C
         EU: 0x20A1FE0
       length:
-        NA: 0x89
-        EU: 0x89
+        NA: 0x8A
+        EU: 0x8A
+      description: "Table of 2-byte values for each IQ skill that represent a group. IQ skills in the same group can not be enabled at the same time."
     - name: SECONDARY_TERRAIN_TYPES
       address:
         NA: 0x20A1AE8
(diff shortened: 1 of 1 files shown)


UsernameFodder:
I think I just assumed the last bit was padding to a 4-byte boundary, since it's right next to `SECONDARY_TERRAIN_TYPES`. What's the format of these table entries? Is it sensible that the last bytes could be considered a null-terminator entry?

theCapypara:
Yeah that's what I guessed could be the case. 

An entry is one 16-bit signed int restriction group). Each entry is supposed to have a matching entry in IQ_SKILLS, but the length of IQ_SKILLS seems to be "correct" in the YML files (those are 32-bit signed ints determining the required IQ amount).

Some sample code that uses it is here: https://github.com/SkyTemple/skytemple-files/tree/master/skytemple_files/hardcoded/iq.py#L153

I could just change it to ignore the zero padding. But it's kinda weird those two table lengths aren't consistent here?

theCapypara:
From Discord:

UsernameFodder:
> I saw your comments over email. If you change the length to 8A instead of 89, does that still fix the issue? 8A makes more sense if the entries are int16
> Also if you want to do me a favor, can you add descriptions to the IQ skill symbols? The length issue was mainly me not knowing what they meant and making bad assumptions

Updated that and added the descriptions.

UsernameFodder (approved):
Thanks for the fix!

PR #11, opened 2022-02-26
Document the WTE format and its applications

EleventhNov:
The end goal should be to hopefully have a fully documented API of where and how the format is used, which considering how small it is, should be rather feasible. It might be so simple in fact that it could serve as a subject for a tutorial / guide! But that's to be discussed and resolved later

### DIFF
--- a/headers/types/common/common.h
+++ b/headers/types/common/common.h
@@ -78,16 +78,16 @@ struct wte_texture_params {
     enum texture_format format : 3;
     bool flip_x : 1;
     bool flip_y : 1;
-    uint8_t unusedD;
+    uint8_t unusedD : 3;
 };
 ASSERT_SIZE(struct wte_texture_params, 2);
 
 struct wte_header {
-    char signature[4]; // 0x0: Signature bytes (must be "\x57\x54\x45\x00")
-    void* texture; // 0x4
-    uint32_t texture_size; // 0x8
+    char signature[4];                // 0x0: Signature bytes (must be "\x57\x54\x45\x00")
+    void* texture;                    // 0x4
+    uint32_t texture_size;            // 0x8
     struct wte_texture_params params; // 0xC
-    uint16_t _padding_0xe; 
+    uint16_t _padding_0xe;
     /*  These bounds are NOT used by the game, but they prove useful to extract the texture out
         of the file. The offsets are redundant and should be zero
 
@@ -96,8 +96,8 @@ struct wte_header {
         needs to be a power of 2 in the range of 8..1024. The actual texture can have a lower
         height, but not a lower width, as the width is required to properly read the image */
     struct rect16_xywh texture_bounds; // 0x10
-    struct rgb* palette; // 0x18
-    uint16_t color_amt; // 0x1C: How many colors are stored in the palette
+    struct rgb* palette;               // 0x18
+    uint16_t color_amt;                // 0x1C: How many colors are stored in the palette
     uint16_t _padding_0x1e;
 };
 ASSERT_SIZE(struct wte_header, 32);
--- a/headers/types/common/common.h
+++ b/headers/types/common/common.h
@@ -82,7 +82,7 @@ struct wte_header {
     struct wte_texture_params params;
     /*  These bounds are NOT used by the game, but they prove useful to extract the texture out
         of the file. The offsets are redundant and should be zero
-        
+
         The width specified here should always be the same as the one specified in the texture
         params, but the height may be lower. The reason is that just like the width, the height
         needs to be a power of 2 in the range of 8..1024. The actual texture can have a lower
--- a/headers/types/common/common.h
+++ b/headers/types/common/common.h
@@ -22,13 +22,11 @@ struct rect16_xywh {
 ASSERT_SIZE(struct rect16_xywh, 8);
 
 // The fact the 4th byte is never used hints to the fact the struct is packed
-#pragma pack(push, 4)
 struct rgb {
-    int8_t r;
-    int8_t g;
-    int8_t b;
+    int32_t r : 8;
+    int32_t g : 8;
+    int32_t b : 8;
 };
-#pragma pack(pop)
 ASSERT_SIZE(struct rgb, 4);
 
 // A structure that represents a file stream for file I/O.
@@ -57,12 +55,12 @@ ASSERT_SIZE(struct file_stream, 72);
 // Struct used to load, store and unload WTE files
 struct wte_handle {
     void* content; // Pointer to the heap-allocated WTE data. Only stored for freeing the data
-    wte_header* header;
+    struct wte_header* header;
 };
 ASSERT_SIZE(struct wte_handle, 8);
 
 // These arguments are almost directly passed to the TEXIMAGE_PARAM register, just rearranged
-// For more information see 
+// For more information see:
 // https://problemkaputt.de/gbatek.htm#ds3dtextureattributes
 struct wte_texture_params {
     uint8_t texture_smult : 3;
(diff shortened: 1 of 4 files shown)


headers/types/common/common.h
```diff
@@ -36,6 +53,47 @@ struct file_stream {
...
+// For more information see:
+// https://problemkaputt.de/gbatek.htm#ds3dtextureattributes
+struct wte_texture_params {
+    uint8_t texture_smult : 3;
+    uint8_t texture_tmult : 3;
+    uint8_t unused6 : 2;
+    enum texture_format format : 3;
+    bool flip_x : 1;
+    bool flip_y : 1;
+};
```
UsernameFodder, line 73: Are the last 3 bits unused? Can we add a `uint8_t unused13 : 3` at the end for explicitness?

headers/types/common/common.h
```diff
@@ -36,6 +53,47 @@ struct file_stream {
...
+ASSERT_SIZE(struct wte_texture_params, 2);
+
+// To accurately describe the structure of the header we need to enable packing
+#pragma pack(push, 4)
```
UsernameFodder, line 77: There's already a global `#pragma pack(1)` active in all headers, so you shouldn't need to mess with the packing. I think it's also easier to reason about structure if padding is stated explicitly. So as defined, this should be (assuming this is the packing you intended):

```c
struct wte_header {
    char signature[4]; // 0x0: Signature bytes (must be "\x57\x54\x45\x00")
    void* texture; // 0x4
    uint32_t texture_size; // 0x8
    struct wte_texture_params params; // 0xC
    uint16_t _padding_0xe;
    /*  These bounds are NOT used by the game, but they prove useful to extract the texture out
        of the file. The offsets are redundant and should be zero
        The width specified here should always be the same as the one specified in the texture
        params, but the height may be lower. The reason is that just like the width, the height
        needs to be a power of 2 in the range of 8..1024. The actual texture can have a lower
        height, but not a lower width, as the width is required to properly read the image */
    struct rect16_xywh texture_bounds; // 0x10
    struct rgbx* palette; // 0x18
    uint16_t color_amt; // 0x1C: How many colors are stored in the palette
    uint16_t _padding_0x1e;
};
ASSERT_SIZE(struct wte_header, 32);
```

Also, this is a fairly large struct so let's put the field offsets in the comments as above. They're convenient to have when looking in a RAM viewer.

headers/types/common/common.h
```diff
@@ -36,6 +53,47 @@ struct file_stream {
 };
 ASSERT_SIZE(struct file_stream, 72);
 
+// Struct used to load, store and unload WTE files
```
UsernameFodder, line 56: I'm not really sure what a WTE file is. Could you add some cursory explanation to this comment about what these files are?

UsernameFodder (changes requested):

UsernameFodder:
> It might be so simple in fact that it could serve as a subject for a tutorial / guide!

I'll never say no to more documentation 😄

headers/types/common/common.h
```diff
@@ -13,6 +13,23 @@ struct iovec {
...
+    int16_t h;
+};
+ASSERT_SIZE(struct rect16_xywh, 8);
+
+// The fact the 4th byte is never used hints to the fact the struct is packed
+struct rgbx {
+    uint8_t r;
+    uint8_t b;
+    uint8_t g;
+    uint8_t x;
```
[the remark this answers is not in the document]
EleventhNov: That'd be smart, although I mostly called it RGBX because that's how many of the existing tech documents call it. There's also a chance that the value might be used in some scenarios (most of the time, x = 0x80), although for WTE I'm pretty certain it's not used at all

headers/types/common/common.h
EleventhNov: About the padding, Yeah I can see how explicit padding would be nicer (but it makes harder to imagine that what you have right there was the struct used in the games almost 15 years ago :P)

I've actually never seen field offsets interspersed in a struct. I wonder if this will always be a requirement for big structs?

headers/types/common/common.h
EleventhNov: I'd be delighted!

headers/types/common/common.h
UsernameFodder: > (but it makes harder to imagine that what you have right there was the struct used in the games almost 15 years ago :P)

Yeah that's probably true. I think it's more useful for us from a pragmatic reverse engineering standpoint though. When you have the full source code as a dev you don't usually need to worry about raw struct layouts in the first place, but we do :P.

> I've actually never seen field offsets interspersed in a struct. I wonder if this will always be a requirement for big structs?

Well if it's too annoying then you don't have to. Maybe it's not so useful in this case. For stuff like `struct dungeon` it's really nice to have as a reference when live-debugging in an emulator, since you can't really import struct definitions into a memory viewer, and it lets the C types serve the same purpose as, e.g., something like [my RAM notes](https://docs.google.com/document/d/1_Q_7BGmNx5wJtJ9iJEwlK1WITjiCcEQxE9C82RECJbg).

headers/types/common/common.h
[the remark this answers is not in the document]
UsernameFodder: Oh okay, I assumed it would always be 0 or something. If there's already a convention about it we can just leave it as `rgbx` then.

headers/types/common/common.h
EleventhNov: > When you have the full source code as a dev you don't usually need to worry about raw struct layouts in the first place, but we do :P.
>
Heh, I can agree with that
> since you can't really import struct definitions into a memory viewer, and it lets the C types serve the same purpose as, e.g., something like [my RAM notes](https://docs.google.com/document/d/1_Q_7BGmNx5wJtJ9iJEwlK1WITjiCcEQxE9C82RECJbg).
>
Actually, there might be a way to use them in an emulator. As you might be aware, Desmume has a Lua scripting engine. With it I've been able to make a small type subsystem so you can bind defined types (including structs!) to addresses and read/write their contents in a typed manner. It's still in a pretty early stage, as I didn't expect to share it with anyone, but maybe we could use this concept here. It'd require parsing the structs (or just add them manually which is what I've been doing), but I'm pretty sure it should be doable (and IMO, it's way more handy than a RAM viewer). I wonder what are your thoughts on this?

UsernameFodder: > I've been able to make a small type subsystem so you can bind defined types (including structs!) to addresses and read/write their contents in a typed manner

That sounds really interesting. I would love to see how that looks and how easy it is to use.

For this repo, I still think field offsets have a lower barrier to entry to use, since it sounds like your method would require a person to set up Lua scripting on DeSmuME (which I believe is only supported on Windows). Minimizing environment setup was something I strived for with this repo, as you can maybe tell from how I set up all this CI stuff so people wouldn't need to set up a dev environment.

But! I would rather have some contribution than none, so if you find labeling field offsets unnecessary or annoying, feel free to just not do it :P. We're not actually losing information so they can always be added later if needed.

EleventhNov: Yeah, It's a shame that Lua scripting and Desmume in general don't work for Linux (I couldn't get it to work at least), being the best emulator for debugging there is. I'd be glad to share the prototype, but I might need some advice regarding how should I share it...

> Minimizing environment setup was something I strived for with this repo
>
Hah, that's actually a very nice thought, setting environments is a pain isn't it? >_< But in the end, it'd only be a lil' upgrade and wouldn't provide anything more than doing it the old fashioned way (other than some convenience :P)

UsernameFodder: > I'd be glad to share the prototype, but I might need some advice regarding how should I share it...

Is it just a Lua script? I think putting it up on GitHub would be a natural way to share it.

EleventhNov: Hm alright... I put it on the fork of this repo, I'm not sure if that's what you meant
You should be able to download the files and test it for yourself. It comes with a lil' example, although there's some documentation missing. Also it might be worth it to move this conversation elsewhere :P (where to, I'm not really sure)

PR #12, opened 2022-02-26
Add info from pmd2scriptdata.xml in skytemple_files

UsernameFodder:
Port over what's possible. Also a bit of additional, related research by me.

@theCapypara could you check that the new data symbol descriptions and accompanying struct comments are accurate? I barely have any scripting experience so a lot of these are educated guesses.

BTW I don't think these are a suitable replacement for `pmd2scriptdata.xml`, but I still wanted to transfer most of the info over.


headers/types/ground_mode/enums.h
```diff
@@ -0,0 +1,1708 @@
+// Enums used primarily in the context of ground mode
+
+#ifndef HEADERS_TYPES_GROUND_MODE_ENUMS_H_
```
theCapypara, line 3: If you want to be game-accurate it would probably be better to replicate the entire tables with all their settings (as an array of structs?)

You could keep the enums for easier lookup.

theCapypara: Ah, I see you do have structs defined for the data. Don't know if you also want to consider adding arrays with the concrete data from the games in it or if that's out of scope.

headers/types/ground_mode/version_dep_enums.h
```diff
@@ -0,0 +1,2476 @@
+// Version-dependent enums for ground mode
+
+#ifndef HEADERS_TYPES_GROUND_MODE_VERSION_DEP_ENUMS_H_
+#define HEADERS_TYPES_GROUND_MODE_VERSION_DEP_ENUMS_H_
+
+// North American version
```
theCapypara, line 6: Idea: You could use a `define`d constant to check the region and set the enums just for the selected region.

theCapypara (approved):
LGTM

headers/types/ground_mode/enums.h
UsernameFodder: Do you mean like recording actual field values from the binaries? Yeah I think that's out of scope (also why I think this isn't a viable replacement for `pmd2scriptdata.xml`. I'd like to keep things at struct definitions without declaring any variables. If people want to know the actual data they can look in the binaries using the layout definitions, or look at `pmd2scriptdata.xml`.

headers/types/ground_mode/version_dep_enums.h
UsernameFodder: I thought about that, but decided against it since that would make these headers slightly harder to use. You'd either have to modify the region def every time manually before importing, pass it as a preprocessor arg, or I'd have to generate near-identical versions of the headers for each region. I'm hoping I don't run into any region-dependent struct layouts that require that sort of thing and can't be solved with unions 😅.

PR #14, opened 2022-03-02
Document camera target in dungeon struct

EleventhNov:
Small change that I needed so here it be :P


UsernameFodder (approved):
Looks good. Incremental improvements are nice :).

PR #15, opened 2022-03-02
Symbols for changing window palettes

EleventhNov:
This one took way more tries than I'd like to admit to get right, but here it is n-n

### DIFF
--- a/headers/functions/overlay29.h
+++ b/headers/functions/overlay29.h
@@ -253,5 +253,6 @@ void LogMessageWithPopup(struct entity* user, char* message);
 void LogMessage(struct entity* user, char* message, bool show_popup);
 void LogMessageById(struct entity* user, int message_id, bool show_popup);
 bool RunDungeonMode(undefined4* param_1, undefined4 param_2);
+void SetBothScreensWindowColorToDefault(void);
 
 #endif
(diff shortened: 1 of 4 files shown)


headers/functions/overlay29.h
```diff
@@ -252,5 +252,6 @@ void LogMessageWithPopup(struct entity* user, char* message);
 void LogMessage(struct entity* user, char* message, bool show_popup);
 void LogMessageById(struct entity* user, int message_id, bool show_popup);
 bool RunDungeonMode(undefined4* param_1, undefined4 param_2);
+void SetBothScreensWindowColorToDefault();
```
UsernameFodder, line 255: Add `void` to the param list, i.e., `void SetBothScreensWindowColorToDefault(void);`

Unlike in C++, an empty param list in C means unspecified rather than explicitly no params.

UsernameFodder (changes requested):

headers/functions/overlay29.h
EleventhNov: Huh, that's interesting actually. Apparently it's something done to allow cross compilation with older programs? I had no idea!

EleventhNov:
My bad for that, I had made a little mistake. Hopefully it's fixed with this

UsernameFodder (approved):
Looks good to me!

PR #17, opened 2022-03-03
Add DungeonGoesUp function

EleventhNov:


UsernameFodder (approved):
Ah nice, we had the restrictions field already but not the function.

EleventhNov:
Yeah, I noticed that. I actually wasn't very sure if this was truly "DungeonGoesUp" or something more general until  I checked your struct thingy :P

UsernameFodder:
Truth be told I haven't verified it personally 😅; it's something I pulled from End's dungeon data doc. But if it turns out to be more general, it can always be updated later.

PR #21, opened 2022-03-22
Add Kaomado-related functions

EleventhNov:
These alone won't do to be able to use Kaomado's yourself, there's a few more functions of interest but I'll leave those for a future pull request


headers/functions/kaomado.h
```diff
@@ -0,0 +1,11 @@
+// All Kaomado loads use the same stream, so it's initialized once on game start!
```
UsernameFodder, line 1: This should go in the description in `arm9.yml`.

headers/functions/kaomado.h
```diff
@@ -0,0 +1,11 @@
+// All Kaomado loads use the same stream, so it's initialized once on game start!
+void InitStreamKaomado(void);
```
UsernameFodder, line 2: Can you put these in `arm9.h`? Making new function files will take a bit more work:
1. Putting `#include` in the right places so that it's reachable from `pmdsky.h`.
2. Modifying `symbol_check.py` to be more robust so it can cross-check different files against a single symbol file.

As I mentioned, I think it's a bit early to be doing this kind of micro-organization.

headers/types/common/common.h
```diff
@@ -136,14 +136,21 @@ struct rect16_xywh {
 };
 ASSERT_SIZE(struct rect16_xywh, 8);
 
-// The fact the 4th byte is never used hints to the fact the struct is packed
-struct rgb {
+struct rgb_24 {
```
UsernameFodder, line 139: How about we call this `rgb`, and the other one `rgbx`? It's clearer, and the `_<number>` suffix is a convention I'd like to reserve for the enum wrappers that #22 introduces.

symbols/arm9.yml
```diff
@@ -1472,6 +1548,21 @@ arm9:
...
         type: struct script_var_table
+    - name: KAOMADO_LAYOUTS
+      address:
+        NA: 0x209E014
+      length:
+        NA: 0xC0
+      description: |-
+        All the possible layouts a Kaomado portrait can be placed in by default
+        
+        type: {struct kaomado_layout}[32]
```
UsernameFodder, line 1559: Remove the curly braces to make it conventional C syntax: `struct kaomado_layout[32]`. I only use the curly brace hack for "inline" struct definitions that aren't in the C headers.

headers/types/common/kaomado.h
```diff
@@ -0,0 +1,54 @@
+// Represents the state of a portrait that wants to be displayed
+struct kaomado {
```
UsernameFodder, line 2: Could we either write a "kaomado 101" like we did with the WTE format, or just link to the exiting Project Pokémon doc? It'll help someone looking at this without prior knowledge.

headers/types/common/kaomado.h
```diff
@@ -0,0 +1,54 @@
+// Represents the state of a portrait that wants to be displayed
```
UsernameFodder, line 1: I guess I'm okay with adding a separate `kaomado.h` for structs, but there needs to be a `#include "kaomado.h"` at the bottom of `common.h` so that it's reachable from `pmdsky.h`.

This file also needs an include guard:

```c
#ifndef HEADERS_TYPES_COMMON_KAOMADO_H_
#define HEADERS_TYPES_COMMON_KAOMADO_H_

...

#endif
```

symbols/arm9.yml
```diff
@@ -813,6 +813,82 @@ arm9:
...
+      address:
+        NA: 0x204D8BC
+      description: |-
+        Tries to load the portrait data associated with the passed Kaomado struct.
+        Returns whether the operation was successful (the portrait could be found)
+        This function also modifies the flip fields in the passed Kaomado struct
+        
+        r0: pointer to Kaomado struct
+        r1: pointer to struct kaomado_buffer
+        return bool
```
UsernameFodder, line 891: Needs a colon

headers/functions/kaomado.h
```diff
@@ -0,0 +1,11 @@
+// All Kaomado loads use the same stream, so it's initialized once on game start!
+void InitStreamKaomado(void);
+
+void InitKaomado(struct kaomado*);
+void InitKaomadoSpecies(struct kaomado*, int16_t species_id);
+void SetKaomadoEmote(struct kaomado*, uint8_t emote_idx);
+void SetKaomadoLayout(struct kaomado*, uint8_t layout_idx);
+void SetKaomadoOffset(struct kaomado*, struct portrait_offset* offset);
+void AllowKaomadoDefault(struct kaomado*, bool allow);
+void IsValidKaomado(struct kaomado*);
```
UsernameFodder, line 10: `bool`?

UsernameFodder (changes requested):

headers/types/common/common.h
EleventhNov: Sure thingy ^^ I mostly avoided using that name because  there were problems with it at first

headers/functions/kaomado.h
EleventhNov: Bool?

EleventhNov: ...OH XD

headers/types/common/kaomado.h
EleventhNov: Oh yeah! Personally I'd prefer to write it, as there's a few aspects that aren't covered/mentioned there (how portraits are looked up, the role of negative pointers as size boundaries, how missing emote look-ups are resolved...) Although where would it go exactly?

UsernameFodder: I would probably put an overview of the format at the top of this file, or in the symbol description of one of the key functions. I would put nitty-gritty details related to a specific function or struct in the relevant place (the description of that function symbol, or as a comment over the relevant struct). The organization of the memory allocation stuff might be a good example (see `struct mem_*` in `common.h` and the `MemAlloc` family in `arm9.yml`).

headers/types/common/enums.h
```diff
@@ -3,6 +3,12 @@
 #ifndef HEADERS_TYPES_COMMON_ENUMS_H_
 #define HEADERS_TYPES_COMMON_ENUMS_H_
 
+// Represents one of the two screens the console has
+enum screen {
+    MAIN = 0,
```
UsernameFodder, line 8: Prefix enums with a "namespace". Otherwise no other enum can use the name, e.g., `MAIN`. Perhaps `SCREEN_` for `enum screen` and `CANVAS_` for `enum kaomado_canvas_state`.

headers/types/common/enums.h
```diff
@@ -3,6 +3,12 @@
 #ifndef HEADERS_TYPES_COMMON_ENUMS_H_
 #define HEADERS_TYPES_COMMON_ENUMS_H_
 
+// Represents one of the two screens the console has
```
UsernameFodder, line 6: What does this mean? Like the top screen and the touch screen? Can we use more descriptive names for the enum labels? It's not clear to me what "main" means in this context.

headers/types/common/kaomado.h
```diff
@@ -0,0 +1,60 @@
...
+};
+ASSERT_SIZE(struct kaomado_buffer, 0x350);
+
+// Stores data of a specialized Kaomado canvas for rendering the portrait into itself
+struct kaomado_canvas {
+    uint8_t canvas_handle;
+    uint8_t pad0x1;
+    uint8_t pad0x2;
+    uint8_t pad0x3;
+    enum kaomado_canvas_state state : 32;
```
UsernameFodder, line 36: #22 actually removes the `: 32` from enums, since they're already assumed to be 4 bytes, and Ghidra behaves better if we avoid bitfields when possible. Let's remove it here too.

symbols/arm9.yml
```diff
@@ -1472,6 +1548,21 @@ arm9:
...
+      description: |-
+        All the possible layouts a Kaomado portrait can be placed in by default
+        
+        type: struct kaomado_layout[32]
+    - name: KAOMADO_FILE
+      address:
+        NA: 0x209E0D4
+      length:
+        NA: 0x14
+      description: "Path of the file where all the portraits are stored. \"FONT/kaomado.kao\", padded with null to a multiple of 4"
```
UsernameFodder, line 1565: If I'm reading this right, we could put `type: char[20]` for courtesy?

symbols/ram.yml
```diff
@@ -191,6 +191,12 @@ arm9:
         2: Today's "Oh My Gosh"
         3: Here Comes Team Charm!
         4: In the Future of Darkness
+    - name: KAOMADO_STREAM
+      address:
+        NA: 0x22AB4B0
+      length:
+        NA: 0x48
+      description: The file stream utilized for all Kaomado portrait loads
```
UsernameFodder, line 199: Is this of type `struct file_stream`, using, e.g., `FileOpen` and such? If so we should be more precise and note that here.

headers/functions/arm9.h
```diff
@@ -91,6 +91,17 @@ void SetBothScreensWindowsColor(int palette_idx);
 void InitMainTeamAfterQuiz(void);
 void ScriptSpecialProcess0x3(void);
 void ScriptSpecialProcess0x4(void);
+
+void InitStreamKaomado(void);
+void InitKaomado(struct kaomado*);
+void InitKaomadoSpecies(struct kaomado*, int16_t species_id);
```
UsernameFodder, line 97: Should `species_id` be of type `enum monster_id`? The width shouldn't really matter here since it's a function param.

UsernameFodder (changes requested):

headers/types/common/kaomado.h
```diff
@@ -0,0 +1,60 @@
+#ifndef HEADERS_TYPES_COMMON_KAOMADO_H_
+#define HEADERS_TYPES_COMMON_KAOMADO_H_
+
+// Represents the state of a portrait that wants to be displayed
+struct kaomado {
+    monster_id_16 species_id; // The species id, or the set index inside kaomado.kao
```
UsernameFodder, line 6: #22 is merged now; this should be `struct monster_id_16`.

UsernameFodder (changes requested):

UsernameFodder:
It was pointed out in https://github.com/UsernameFodder/pmdsky-debug/pull/156 that with the current state of the repo, a lot of the things added in this PR adds overlap with existing symbols/structs, but there's a bunch of stuff here that wasn't already documented.

@EleventhNov Since this PR has been sitting for quite some time, I went ahead and updated this branch and merged stuff in a way I thought was reasonable. Feel free to glance at the new version if you feel like it. If there are no concerns, I plan to merge this within the next day or so.

UsernameFodder (approved):

PR #22, opened 2022-03-22
Don't pack structs by default

UsernameFodder:
The `#pragma pack` directive is the only way to pack structs that
Ghidra's C parser supports, but it prevents compilers from assuming
struct alignment, which causes them to emit a lot of boilerplate code
that can be safely run even with unaligned structs. This is horribly
inefficient for writing patches using the C headers.

Most of the C structs don't actually require packing. Change structs to
be naturally aligned by default, but also keep a build target that
switches back to the old implicit packing behavior. By compiling in both
modes simultaneously, the clarity benefits of forbidding implicit struct
padding is preserved (as long as structs are accompanied by an
ASSERT_SIZE), while giving compilers more reasonable alignment
guarantees in most cases.

As part of this migration, replace all 8-bit and 16-bit enum bitfields
with proper wrapper structs. This should improve decompiler output in
Ghidra (which still doesn't support bitfields), and also allow enum
types to be more readily applied to globals. The convenience macros
ENUM_8_BIT and ENUM_16_BIT are added to help standardize these wrapper
structs and make them easier to define.

### DIFF
--- a/.github/workflows/check-headers.yml
+++ b/.github/workflows/check-headers.yml
@@ -19,7 +19,15 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v3
       - name: Compile
-        run: make -C headers
+        run: make -C headers headers-aligned
+  implicit-padding-check:
+    runs-on: ubuntu-latest
+    needs: compile
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: Compile with implicit packing
+        run: make -C headers headers-packed
   format-check:
     runs-on: ubuntu-latest
     steps:
--- a/headers/Makefile
+++ b/headers/Makefile
@@ -9,10 +9,23 @@ else
 $(error C compiler not found)
 endif
 
+# Compile headers
+# Structs constrained by ASSERT_SIZE can only compile under both natural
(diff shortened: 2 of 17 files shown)


UsernameFodder:
@EleventhNov could you try this branch with whatever you were doing and check that it gives better compiler output?

EleventhNov:
> @EleventhNov could you try this branch with whatever you were doing and check that it gives better compiler output?

It does! Just one small thing, I'm using the arm-none-eabi-gcc compiler (for obvious reasons :p) and I don't think that one has a -m32 flag. However it does have a -mno-short-enums, and it won't work without that flag. I'm not sure if this is relevant, but figured I'd bring it up :P

UsernameFodder:
> I'm using the arm-none-eabi-gcc compiler (for obvious reasons :p) and I don't think that one has a -m32 flag. However it does have a -mno-short-enums, and it won't work without that flag (as all the static asserts will fire). I'm not sure if this is relevant, but figured I'd bring it up :P

That's fine, the compiler flags in this repo are just for size checks. C is C, so they can be compiled with whatever flags or cross-compilers suit your needs.

PR #25, opened 2022-04-20
Fall back to fs::copy on tempfile persist errors

UsernameFodder:
On Unix systems where the temp directory is mounted on a different
filesystem from the target path (which is the case for /tmp on some
systems), NamedTempFile::persist() will fail. Fall back to fs::copy() to
handle such failures.


UsernameFodder:
@theCapypara Does this change fix the issue for you?

theCapypara:
Yep, this works!

PR #29, opened 2022-05-01
Mark strings that are (probably) const as const

theCapypara:
This marks some strings for functions in the C-header files as const.
Now I didn't really test any of them, I just went with strings that will probably / should really be const / have no reason to really not be const.
I would consider this change experimental, but we can roll it back in parts if needed.

Rationale: This will make it easier to build a fast Rust FFI since Rust code would need to work with heap allocations to properly support passing non-const strings.


theCapypara:
Ah, I'll fix the formatting issues, one sec.

theCapypara:
Actually it just occured to me that I forgot that Rust and C-Strings are fundementally different it will require re-allocation anyway, welp. This is still easier in Rust to deal with than mutable C-strings.

UsernameFodder (changes requested):
I'm not all that concerned about things breaking 🤷. It won't have any effect for RE tools, so the only issues that could arise would be when trying to write patches, and hopefully it should be pretty obvious if something is wrong.

UsernameFodder:
@theCapypara There are a couple other things we might as well make `const char*`, and I think we ought to for standardization:

- `msg` in `AppendProgPos`
- `s` in `Mbtowc`
- `s` in `Strlen`
- `s1` and `s2` in `Strcmp` and `Strncmp`
- `string` in `Strchr`
- `string` and `stopset` in `Strcspn`

UsernameFodder:
LGTM, thanks.

UsernameFodder (approved):

PR #32, opened 2022-05-06
Relicense symbols and C headers under GPLv3-MIT dual license

UsernameFodder:
Projects like tech-ticks' c-of-time use the symbol tables and C headers
to generate actual ROM patches. However, the current GPLv3 licensing of
the symbols/headers are problematic for the distribution of these
patches, since they are inherently tied to the closed-source EoS ROMs
and thus incompatible with GPL.

Release the symbols and C headers under a dual GPLv3-MIT license to make
it easier for people to create and distribute patches derived from them.
Patch creators may choose to use the symbols and headers under the MIT
license.

Note that this commit does NOT change the license of the overall
pmdsky-debug repository, which is and remains GPLv3.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,3 +54,5 @@ This repository is an attempt to implement the first sharing method (machine-rea
 ### No$GBA with symbol names
 ![No$GBA with symbol names](docs/images/top-readme-nocash.png)
 
+## Licensing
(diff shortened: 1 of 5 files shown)


UsernameFodder:
@theCapypara @EleventhNov Are you alright with your contributions being relicensed under these new terms?

EleventhNov:
Yeah

theCapypara:
Yep

PR #33, opened 2022-05-06
Duplicated functions in overlay 29

Metalcape:
All the addresses I could find for ItemIsActive, DefenderAbilityIsActive, IsMonster and EntityIsValid. I did not exhaustively check if EntityIsValid is ever called on a different type of pointer, because there are really a lot of references to it, but for now it seems that it's only used for entities.


symbols/overlay29.yml
```diff
@@ -235,6 +377,65 @@ overlay29:
...
+          - 0x230B3B4
+          - 0x23125F4
+          - 0x23237CC
+          - 0x23290A0
+          - 0x232A980
+          - 0x232C840
+          - 0x232E860
+          - 0x233344C
+      description: |-
+        Checks if a defender has an active ability that isn't disabled by an attacker's Mold Breaker.
```
UsernameFodder, line 406: I just realized there are two variations of this function with different assembly. I think they're clearly the same function, but maybe the compiler optimized them differently. Can we add a line to the description mentioning this? Maybe something like:

```
Checks if a defender has an active ability that isn't disabled by an attacker's Mold Breaker.

This function has two variants, each with slightly different assembly. This is probably due to differences in compiler optimizations at different addresses.
```

UsernameFodder (changes requested):

headers/functions/overlay29.h
```diff
@@ -35,7 +35,6 @@ bool MonsterIsType(struct entity* entity, enum type_id type_id);
 bool IqSkillIsEnabled(struct entity* entity, enum iq_skill_id iq_id);
 int GetMovePower(struct entity* entity, struct move* move);
 void AddExpSpecial(struct entity* attacker, struct entity* defender, int base_exp);
-bool PptrIsValid(void** pptr);
 bool DefenderAbilityIsActive(struct entity* attacker, struct entity* defender,
```
UsernameFodder, line 38: Hmm, since we're reordering the symbols, we might as well reorder the function signatures as well to keep things nice.

UsernameFodder (changes requested):

UsernameFodder (changes requested):
Oh and also, feel free to add yourself to the top-level `CREDITS.txt` file if you want.

headers/functions/overlay29.h
Metalcape: Do you mean reordering the signatures to match the order of the functions in the symbol file?

UsernameFodder: Yep, that's what I meant.

symbols/overlay29.yml
```diff
@@ -235,6 +377,67 @@ overlay29:
...
+
```
UsernameFodder, line 407: Need an indent here 😛

UsernameFodder (changes requested):

UsernameFodder (approved):
All looks good now. Thanks!

PR #36, opened 2022-05-08
Missing EU symbols in arm9.bin and dungeon mode

Metalcape:
This should fill in the gap between NA and EU symbols, at least for dungeon mode. All the missing data fields have the same length in both versions as far as I know. I also added some hardcoded values to the description of their symbols so that they are easier to find in the JP version.


symbols/arm9.yml
```diff
@@ -686,7 +686,23 @@ arm9:
           - 0x2052418
           - 0x2054A60
           - 0x20609E8
-        EU: 0x200D6BC
+        EU:
+          - 0x200D6BC
```
UsernameFodder, line 690: `arm5find` is actually giving me 18 addresses (both for NA/EU), but there are only 16 here. Are some of the matches invalid?

symbols/arm9.yml
```diff
@@ -686,7 +686,23 @@ arm9:
           - 0x2052418
           - 0x2054A60
           - 0x20609E8
```
UsernameFodder, line 688: This NA address list was actually manual. Could we fill these in too, so we're matching the EU list? In the current commit there are 14 for NA but 16 for EU.

UsernameFodder (changes requested):
Thanks a ton for doing this! I've been meaning to get to this but could never bring myself to do it 🤣

symbols/arm9.yml
Metalcape: I did find those missing copies (4 for EU and 2 for NA) but I didn't add them because they seem to be unreachable code. Ghidra doesn't disassemble them unless you do it manually. It's like in the heap allocation table, there's some valid code there as well but it doesn't seem to be referenced anywhere

UsernameFodder: It's interesting that EU only has 2 whereas NA has 4...

Well, I guess it doesn't matter too much at any rate.

UsernameFodder (approved):

PR #40, opened 2022-05-12
Add some fields for file_stream, update FileSeek desc.

theCapypara:
- Clarified what `0x24` - `0x2c` are in the `file_stream` struct
- Updated the FileSeek description to no longer mark seeking from the end as unimplemented, since it seems to be implemented.

### DIFF
--- a/symbols/arm9.yml
+++ b/symbols/arm9.yml
@@ -552,7 +552,7 @@ arm9:
         This function has the a similar API to the fseek(3) library function from C, including using the same codes for the `whence` parameter:
         - SEEK_SET=0
         - SEEK_CUR=1
-        - SEEK_END=2 (maybe not implemented?).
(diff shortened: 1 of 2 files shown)


symbols/arm9.yml
```diff
@@ -552,7 +552,7 @@ arm9:
         This function has the a similar API to the fseek(3) library function from C, including using the same codes for the `whence` parameter:
         - SEEK_SET=0
         - SEEK_CUR=1
-        - SEEK_END=2 (maybe not implemented?).
+        - SEEK_END=2.
```
UsernameFodder, line 555: Remove period :P

UsernameFodder (changes requested):

UsernameFodder (approved):

PR #41, opened 2022-05-12
Add some symbols related to abilities and item effects

Metalcape:
- Add the functions `TrySwitchPlace`, `CheckSlowStart`, `CheckArtificialWeatherAbilities`, `TryAbilityTruant`, `GetNumberOfAttacks`, `IsLevitating`, `TickNoSlipCap`, `AdvanceStatusAndHealthRegen`, `Conversion2IsActive`, and `TryWeatherFormChange`
- Add data labels related to some of the above functions (Slow Start message ID, the max HP cap, Castform and Cherrim form IDs, and exclusive item effects IDs)
- Merge the function `ExclusiveItemEffectIsActive` into `TeamExclusiveItemEffectIsActive`, since they are identical, and add more addresses for it
-  dungeon.h
    - Add a bit of information in the comment about `fractional_turn`
    - Add fields 0x3E34 - 0x3E37 related to the abilities Plus and Minus


symbols/overlay29.yml
```diff
@@ -691,6 +770,22 @@ overlay29:
...
         return: boost
+    - name: TickNoSlipCap
+      address:
+        NA: 0x230FB90
+        EU: 0x2310604
+      description: |-
+        Checks if the entity is a team member and holds the No-Slip Cap, and if so attempts to make one item in the bag sticky.
+        
+        r0: pointer to entity
+    - name: AdvanceStatusAndHealthRegen
```
UsernameFodder, line 781: Can we call this `TickStatusAndHealthRegen` to be consistent with `TickNoSlipCap`?

symbols/overlay29.yml
```diff
@@ -506,6 +586,15 @@ overlay29:
         r0: entity pointer
         r1: ability ID
         return: bool
+    - name: IsLevitating
```
UsernameFodder, line 589: Can we call this `LevitateIsActive` for consistency with other functions?

symbols/overlay29.yml
```diff
@@ -445,6 +472,48 @@ overlay29:
...
         return: bool
+    - name: TryAbilityTruant
+      address:
+        NA: 0x22F97F0
+        EU: 0x22FA1FC
+      description: |-
+        Checks if an entity has the ability Truant, and if so tries to apply the pause status to it.
+        
+        r0: pointer to entity
+    - name: TeamExclusiveItemEffectIsActive
```
UsernameFodder, line 483: Now that I know they're the same, I think I'd prefer the shorter `ExclusiveItemEffectIsActive`.

symbols/overlay29.yml
```diff
@@ -445,6 +472,48 @@ overlay29:
         
         r0: entity pointer
         return: bool
+    - name: TryAbilityTruant
```
UsernameFodder, line 475: Let's call this `TryActivateTruant`, similar to `TryActivateQuickFeet`.

symbols/overlay29.yml
```diff
@@ -384,6 +395,22 @@ overlay29:
         Has a special check to return false for Labyrinth Cave B10F (the Gabite boss fight).
         
         return: bool
+    - name: CheckSlowStart
```
UsernameFodder, line 398: I don't really like calling this "Check*" if it's actually doing something. Let's call it `TryActivateSlowStart`

symbols/overlay29.yml
```diff
@@ -384,6 +395,22 @@ overlay29:
...
+      address:
+        NA: 0x22F923C
+        EU: 0x22F9C48
+      description: |-
+        Runs a check over all monsters on the field for the ability Slow Start, and lowers the speed of those who have it.
+        
+        No params
+    - name: CheckArtificialWeatherAbilities
```
UsernameFodder, line 406: Similar to the Slow Start function, let's call this `TryActivateArtificialWeatherAbilities`

UsernameFodder (changes requested):
I wonder how many other functions in the symbol tables are actually duplicates 😅.

Basically just comments on function names; the content itself looks good.

PR #43, opened 2022-05-16
JP: Overlay load addresses and function symbols

theCapypara:
Overlay load addresses for JP ROMs (DS) and auto-generated symbols for the jp ROM using the tools in this repo and updates to said tools to support the JP ROMs.


tools/offsets.py
```diff
@@ -122,6 +122,45 @@ def __contains__(self, address: int) -> bool:
         "overlay34": Binary(0x22DCB80, 0xDC0),
         "overlay35": Binary(0x22BD3C0, 0x20),
     },
+    "JP": {
+        "arm9": Binary(0x2000000, 0x756920),
```
UsernameFodder, line 126: This length is wrong, should be 0xB8CB8 based on the YAML file updates.

UsernameFodder (changes requested):

tools/offsets.py
theCapypara: Yep, no idea how I got this value.

PR #50, opened 2022-05-28
Symbols related to status effects, enemy spawn and some dungeon.h updates

Metalcape:
I've had these waiting inside Ghidra for a while, I decided to submit all of them so that I can wait for End to finish his info dump without possible conflicts.

- Added some functions related to status effects and enemy spawn
- Added a data structure used during enemy spawn (which is also partially described in irdkwia's docs)
- Added some dungeon.h fields and a few comments to partially unknown fields
- Some more JP symbols found automatically by symbols_vfill
- Found a new version of `GetTile` and renamed the previous one into `GetTileAndCopyIfDefault` to distinguish it
- Fixed a typo in an address of `DefenderAbilityIsActive`

A few things that I'd like some feedback on:

- `FloorLoop` and `TryForcedLoss` have partially unknown parameters because they're pretty complex functions and I couldn't look into them in detail, let me know if that's ok
- That `ReturnInt1` "function" has many copies, some probably unreachable, I only reported the one that's called when spawning enemies
- I decided to split `TryActivateMapSurveyor` into two different symbols for EU and NA because they actually do different things, I can merge them again if you think it's more appropriate
- `MonsterCanSpawn` and the flag it reads seem to be useless, because the function always returns true unless it's trying to spawn Mew, which never spawns randomly on a floor, so maybe it's a check for some unused spawn?


theCapypara:
ReturnInt1 is probably a function that returns a debugging flag, that in the release version just returns 1.
Would probably be interesting to see what happens if you patch it to return 0.

headers/types/dungeon_mode/dungeon_mode.h
```diff
@@ -1387,6 +1385,25 @@ struct prng_state {
 };
 ASSERT_SIZE(struct prng_state, 20);
 
+// Contains the necessary information to spawn a monster.
+// Allocated on the stack and passed via a pointer to SpawnMonster.
+struct spawned_monster_data {
+    monster_id_16 monster_id;  // 0x0: The id of the monster to spawn
```
UsernameFodder, line 1391: struct monster_id_16

headers/types/dungeon_mode/dungeon_mode.h
```diff
@@ -1387,6 +1385,25 @@ struct prng_state {
...
+    monster_behavior behavior; // 0x2: NPC behavior of the monster
```
UsernameFodder, line 1392: struct monster_behavior_8

headers/types/dungeon_mode/dungeon_mode.h
```diff
@@ -1387,6 +1385,25 @@ struct prng_state {
...
+    undefined field_0x3;
+    undefined field_0x4;
+    undefined field_0x5;
+    undefined field_0x6;
+    undefined field_0x7;
+    uint16_t level; // 0x8: The level of the spawned monster
+    position pos;   // 0xA: Position on the floor where the monster will be spawned
```
UsernameFodder, line 1399: struct position

headers/types/dungeon_mode/dungeon.h
```diff
@@ -19065,20 +18944,20 @@ struct dungeon {
     undefined field_0x19901;
     undefined field_0x19902;
     undefined field_0x19903;
-    undefined field_0x19904;
+    undefined field_0x19904; // Unknown entity pointer to a monster.
```
UsernameFodder, line 18947: If this is an entity pointer, does it make sense to change this to `struct entity*`?

headers/functions/overlay29.h
```diff
@@ -128,9 +148,13 @@ bool TryIncreaseHp(struct entity* user, struct entity* target, int hp_restoratio
 bool TryInflictLeechSeedStatus(struct entity* user, struct entity* target, bool log_failure,
                                bool check_only);
 void TryInflictDestinyBond(struct entity* user, struct entity* target);
+bool IsBlinded(struct entity* entity, bool check_held_item);
 void RestoreMovePP(struct entity* user, struct entity* target, int pp, bool suppress_logs);
+void SetReflectDamageCountdownTo4(struct entity* entity);
 bool HasConditionalGroundImmunity(struct entity* entity);
 int Conversion2IsActive(struct entity* entity);
+bool IsTargetInRange(struct entity* user, struct entity* target, int direction_index,
```
UsernameFodder, line 156: `enum direction_id direction`

headers/functions/overlay29.h
```diff
@@ -128,9 +148,13 @@ bool TryIncreaseHp(struct entity* user, struct entity* target, int hp_restoratio
...
+                     int move_range);
```
UsernameFodder, line 157: Since `enum move_range` exists (and this is not the same thing), can we call this something like `n_tiles` instead?

headers/functions/overlay29.h
```diff
@@ -44,6 +59,9 @@ bool MonsterIsType(struct entity* entity, enum type_id type_id);
 bool IqSkillIsEnabled(struct entity* entity, enum iq_skill_id iq_id);
 int GetMovePower(struct entity* entity, struct move* move);
 void AddExpSpecial(struct entity* attacker, struct entity* defender, int base_exp);
+uint8_t GetSleepAnimationId(struct entity* entity);
+void EndFrozenClassStatus(struct entity* user, struct entity* target, bool print_to_log);
```
UsernameFodder, line 63: Can we just call the third param `bool log`? It lines up better with the `log_failure` params in some of the other status-related functions.

headers/functions/overlay29.h
```diff
@@ -24,18 +29,28 @@ void DungeonRngUnsetSecondary(void);
...
 void ResetDamageDesc(undefined4* damage_desc);
 bool FloorNumberIsEven(void);
+enum monster_id GetKecleonIdToSpawnByFloor(void);
 void TryActivateSlowStart(void);
 void TryActivateArtificialWeatherAbilities(void);
 bool DefenderAbilityIsActive(struct entity* attacker, struct entity* defender,
                              enum ability_id ability_id, bool attacker_ability_enabled);
 bool IsMonster(struct entity* entity);
 void TryActivateTruant(struct entity* entity);
+bool MonsterCanSpawn(struct monster_id_16 monster_id, bool fail_if_mew);
```
UsernameFodder, line 44: Use the plain `enum monster_id` in function signatures; there's no distinction since all registers are 32-bit, and the plain enum looks better in the Ghidra decompiler.

symbols/overlay29.yml
```diff
@@ -387,13 +448,45 @@ overlay29:
...
         r1: pointer to target entity
+    - name: FloorLoop
+      address:
+        NA: 0x22EBD08
+        EU: 0x22EC6B8
+        JP: 0x22ED370
+      description: |-
+        The main function which updates the state of the floor. Called in a loop by InitializeDungeon while FloorIsOver returns false.
+        
+        The parameter is always equal to 1 and is used as parameter for another unknown function.
```
UsernameFodder, line 467: This isn't true, the code in `InitializeDungeon` is:

```c
uVar8 = 1;
do {
    FloorLoop(uVar8);
    iVar17 = FloorIsOver();
    uVar8 = 0;
} while (iVar17 == 0);
```

So r0 here must be a flag for "is this the first loop?"

headers/functions/overlay29.h
```diff
@@ -24,18 +29,28 @@ void DungeonRngUnsetSecondary(void);
 void DungeonRngSetSecondary(int i);
 void DungeonRngSetPrimary(void);
 void TrySwitchPlace(struct entity* user, struct entity* target);
+void FloorLoop(int unknown);
```
UsernameFodder, line 32: See my other comment; let's change this to `void FloorLoop(bool first_loop)`

headers/functions/overlay29.h
```diff
@@ -8,9 +8,14 @@ void DungeonFree(void);
 int InitializeDungeon(undefined* dungeon_data, struct dungeon* dungeon);
 bool EntityIsValid(struct entity* entity);
 enum floor_type GetFloorType(void);
+bool TryForcedLoss(bool skip_floor_end_check);
 bool FixedRoomIsSubstituteRoom(void);
+struct tile* GetTileAtEntity(struct entity* entity);
 void SubstitutePlaceholderStringTags(int string_id, struct entity* entity, undefined4 param_3);
 bool ItemIsActive(struct entity* entity, enum item_id item_id);
+struct monster_id_16 GetMonsterIdToSpawn(int spawn_weight);
+uint8_t GetMonsterLevelToSpawn(struct monster_id_16 monster_id);
```
UsernameFodder, line 17: enum monster_id

headers/functions/overlay29.h
```diff
@@ -8,9 +8,14 @@ void DungeonFree(void);
 int InitializeDungeon(undefined* dungeon_data, struct dungeon* dungeon);
 bool EntityIsValid(struct entity* entity);
 enum floor_type GetFloorType(void);
+bool TryForcedLoss(bool skip_floor_end_check);
 bool FixedRoomIsSubstituteRoom(void);
+struct tile* GetTileAtEntity(struct entity* entity);
 void SubstitutePlaceholderStringTags(int string_id, struct entity* entity, undefined4 param_3);
 bool ItemIsActive(struct entity* entity, enum item_id item_id);
+struct monster_id_16 GetMonsterIdToSpawn(int spawn_weight);
```
UsernameFodder, line 16: enum monster_id

symbols/overlay29.yml
```diff
@@ -188,6 +188,18 @@ overlay29:
...
         return: floor type
+    - name: TryForcedLoss
+      address:
+        NA: 0x22E0620
+        EU: 0x22E0F60
+        JP: 0x22E1CAC
+      description: |-
+        Attempts to trigger a forced loss of the type specified in the forced loss reason field of the dungeon struct. Returns true if the loss happens, false otherwise.
+        
+        If the parameter is true, the function will not check for the end of the floor condition and will skip other (unknown) actions in case of forced loss.
```
UsernameFodder, line 199: Put this on the r0 line

headers/functions/arm9.h
```diff
@@ -179,6 +180,7 @@ uint32_t SubFixedPoint(uint32_t val_fp, uint32_t dec_fp);
 uint32_t BinToDecFixedPoint(uint32_t* q16);
 int CeilFixedPoint(uint32_t val_fp);
 bool DungeonGoesUp(enum dungeon_id dungeon_id);
+uint8_t GetMonsterLevelFromSpawnEntry(monster_spawn_entry* entry);
```
UsernameFodder, line 183: struct monster_spawn_entry*

headers/functions/arm9.h
```diff
@@ -135,6 +135,7 @@ int GetSpecialEpisodeType(void);
 void ScenarioFlagBackup(void);
 void InitWorldMapScriptVars(void);
 void InitDungeonListScriptVars(void);
+int ReturnInt1(void);
```
UsernameFodder, line 138: Change return type to `bool`?

UsernameFodder (changes requested):
It seems like you'll need to resolve merge conflicts from End's recent info dump before the tests will run. Hopefully they're not too bad.

Metalcape:
I don't understand why the ASSERT_SIZE is failing, without padding the struct should be 16 bytes right?

symbols/overlay29.yml
```diff
@@ -495,10 +599,24 @@ overlay29:
       address:
         NA: 0x22F97F0
         EU: 0x22FA1FC
+        JP: 0x22FADA8
       description: |-
         Checks if an entity has the ability Truant, and if so tries to apply the pause status to it.
         
         r0: pointer to entity
+    - name: MonsterCanSpawn
```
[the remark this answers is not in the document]
Metalcape: Oh right, I didn't remember that, I was thinking about the mystery part/slab spawns but mew wasn't among those. This would actually make sense, so what about renaming it something like `IsUnusedMewSpawn`? Considering that it would return true for all other monster IDs

symbols/overlay29.yml
```diff
@@ -1539,17 +1809,29 @@ overlay29:
...
+        NA: 0x23360FC
+        EU: 0x2336CCC
+        JP: 0x23374CC
+      description: |-
+        Get the tile at some position. If the coordinates are out of bounds, returns a default tile.
+        
+        r0: x position
+        r1: y position
+        return: tile pointer
+    - name: GetTileAndCopyIfDefault
```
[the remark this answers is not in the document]
Metalcape: That may be the case, I'll change the name because I think it sounds better

symbols/arm9.yml
```diff
@@ -1375,10 +1408,20 @@ arm9:
       address:
         NA: 0x204CE90
         EU: 0x204D1C8
+        JP: 0x204D1F0
       description: |-
         Initializes the DUNGEON_*_LIST script variable values (IDs 0x4f-0x54).
         
         No params.
+    - name: ReturnInt1
```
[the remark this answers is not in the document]
Metalcape: In the specific case of enemy spawn this address is called before the unused Mew check so that if this function actually returns false, the monster spawn always fails. So yes, it could be a debug switch of the type "toggle enemy spawn" used when debugging dungeons or something. It's also called in 2 other fuction which at a glance are also related to enemy spawn, and if it returns false some pieces of code are skipped.

For the other copies, I found 8 of them in total in the EU rom. 3 of them are referenced in some address tables in arm9.bin that probably contain other debug functions, 3 others are referenced by global pointers, while only this one and another one seem to be used in actual gameplay-related functions (the other one seems to disable recruiting if it returns false).

If you think this current information is enough we can rename it `MonsterSpawnsEnabled`, otherwise I'd just call it `ReturnTrue` and then maybe we'll rename it in the future when we better understand what the other copies do

headers/types/dungeon_mode/dungeon.h
```diff
@@ -1678,134 +1682,10 @@ struct dungeon {
...
-    undefined field_0x380e;
-    undefined field_0x380f;
-    undefined field_0x3810;
-    undefined field_0x3811;
-    undefined field_0x3812;
-    undefined field_0x3813;
+    // 0x3794: Table of monsters that can be spawned on the floor.
+    // The number of valid entries is stored in monster_spawn_table_length.
+    // The last 2 entries of the table are always Kecleon (0x17F) and the dummy (0x229).
+    struct monster_spawn_entry monster_spawn_table[16];
```
Metalcape, line 1688: I noticed that there was already an instance of this at 0x2C964. This one is used when spawning monsters, but the other one seems to be used during floor initialization. Should I call this `spawn_entries` and rename the other to something like `spawn_entries_master`?

symbols/overlay29.yml
```diff
@@ -412,13 +473,45 @@ overlay29:
...
+        No params.
+    - name: FloorIsOver
+      address:
+        NA: 0x22EC7E8
+        EU: 0x22ED198
+        JP: 0x22EDE50
+      description: |-
+        Returns true if the player will advance to the next floor on the next turn, and in that case sets the floor_loop_status in the dungeon struct to 2.
+        
+        return: bool
```
End45, line 514:
```suggestion
        Checks if the current floor should end, and updates dungeon::floor_loop_status if required.
        If the player has been defeated, sets dungeon::floor_loop_status to floor_loop_status::FLOOR_LOOP_LEADER_FAINTED.
        If dungeon::end_floor_flag is 1 or 2, sets sets dungeon::floor_loop_status to floor_loop_status::FLOOR_LOOP_NEXT_FLOOR.
        
        return: True if the current floor should end
```
Looking at my notes to write this made me realize that dungeon::end_floor_flag is actually an `uint8_t`, not a `bool`. We'll have to change that at some point.

symbols/overlay29.yml
```diff
@@ -567,10 +715,38 @@ overlay29:
...
+      address:
+        NA: 0x22FD084
+        EU: 0x22FDA80
+        JP: 0x22FE478
+      description: |-
+        Spawns a new monster on a tile, according to the information passed by the first parameter.
+        
+        If the second parameter is true, the monster cannot spawn asleep, otherwise it will randomly be asleep.
+        
+        r0: pointer to spawned monster data (on the stack)
```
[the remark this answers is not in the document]
End45: This is a struct btw, the wording doesn't make that very clear. Maybe we should point that out so it can be added to the headers eventually.

symbols/overlay29.yml
```diff
@@ -1304,6 +1538,17 @@ overlay29:
         
         r0: user entity pointer
         r1: target entity pointer
+    - name: IsBlinded
+      address:
+        NA: 0x23177E4
+        EU: 0x2318244
+        JP: 0x2318CB4
+      description: |-
+        Returns true if the monster has the blinded status, or if it is not the leader and is holding Y-Ray Specs. If the second parameter is false, skips checking for the held item.
```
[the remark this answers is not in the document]
End45: Maybe this should reference `status_id::STATUS_BLINKER`?

End45 (changes requested):

symbols/overlay29.yml
```diff
@@ -412,13 +473,45 @@ overlay29:
...
         EU: 0x22EBB28
+        JP: 0x22EC7E0
       description: |-
         The user entity attempts to switch places with the target entity (i.e. by the effect of the Switcher Orb). 
         
         The function checks for the Suction Cups ability for both the user and the target, and for the Mold Breaker ability on the user.
         
         r0: pointer to user entity
         r1: pointer to target entity
+    - name: FloorLoop
```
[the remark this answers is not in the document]
Metalcape: What about `RunFractionalTurn`? Since it's a term that was already used to describe the sub-actions that happen in a turn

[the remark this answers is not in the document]
End45: That sounds good

Metalcape:
I'm going to mark as resolved the changes that I implemented in the last commit, to better keep track of things

symbols/overlay29.yml
```diff
@@ -241,15 +241,16 @@ overlay29:
...
-    - name: TryActivateMapSurveyorNA
+    - name: UpdateMapSurveyorFlag
       address:
         NA: 0x22E2DD8
+        EU: 0x22E375C
       description: |-
         Sets the Map Surveyor flag in the dungeon struct to true if a team member has Map Surveyor, sets it to false otherwise.
         
-        Exclusive to the NA version; the EU equivalent returns true if the flag was changed.
+        This function has two variants: in the EU ROM, it will return true if the flag was changed. The NA version always returns 0x41 (equivalent to true).
```
End45, line 251: > The NA version always returns 0x41 (equivalent to true)
This is incorrect, this function calls `TeamMemberHasEnabledIqSkill` after the `mov r0,41h`, so it returns the return value of `TeamMemberHasEnabledIqSkill` (so whether a team member has map surveyor enabled in this case).

Metalcape: Oh that's right, then what about:
```
[...] The NA version will return the new value of the flag.
```
(since r0 gets stored as the new flag)

End45: Sounds good

symbols/overlay29.yml
```diff
@@ -412,13 +474,45 @@ overlay29:
...
+        The main function which executes the actions that take place in a fractional turn. Called in a loop by InitializeDungeon while FloorIsOver returns false.
+        
+        r0: first loop flag (true when the function is first called during a floor)
+    - name: TrySpawnMonsterAndActivatePlusMinus
+      address:
+        NA: 0x22EC6DC
+        EU: 0x22ED08C
+        JP: 0x22EDD44
+      description: |-
+        Called at the beginning of FloorLoop. Executed only if FRACTIONAL_TURN_SEQUENCE[fractional_turn * 2] is not 0.
```
End45, line 500:
```suggestion
        Called at the beginning of RunFractionalTurn. Executed only if FRACTIONAL_TURN_SEQUENCE[fractional_turn * 2] is not 0.
```

symbols/overlay29.yml
```diff
@@ -3340,6 +3650,17 @@ overlay29:
...
         Directions start with 0=down and proceed counterclockwise (see enum direction_id). Displacements for x and y are interleaved and encoded as 2-byte signed integers. For example, the first two integers are [0, 1], which correspond to the x and y displacements for the "down" direction (positive y means down).
+    - name: FRACTIONAL_TURN_SEQUENCE
+      address:
+        NA: 0x2352284
+        EU: 0x2352EC2
+      length:
+        NA: 0xFA
+        EU: 0xFA
+      description: |-
+        Read by certain functions that are called by FloorLoop to see if they should be executed.
```
End45, line 3661:
```suggestion
        Read by certain functions that are called by RunFractionalTurn to see if they should be executed.
```

End45 (changes requested):

symbols/arm9.yml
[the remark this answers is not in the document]
UsernameFodder: Just because all the "copies" have the same code doesn't mean we need to label them the same. If `MonsterSpawnsEnabled` is a good name for this particular address, then we should call it that. If the other ones do something different, they probably correspond to different debug flags, and we should give them different names.

symbols/overlay29.yml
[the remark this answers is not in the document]
UsernameFodder: What do you mean by "unused" in `IsUnusedMewSpawn`? Do we know that it's actually unused? Maybe we can call it something like `MewSpawnCheck`.

symbols/overlay29.yml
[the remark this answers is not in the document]
UsernameFodder: Couldn't hurt I guess. I'm less concerned with symbols than I am with struct fields since it should be easy enough to just go look at the symbol address to see the correspondence.

symbols/overlay29.yml
UsernameFodder: Oops this was my suggestion, apparently I don't know how to read assembly 🙃

symbols/overlay29.yml
UsernameFodder: What do the different values mean? If we're updating the name, we should update the field's comment too.

headers/types/dungeon_mode/dungeon.h
```diff
@@ -24,7 +24,9 @@ struct dungeon {
     undefined field_0xb;
     undefined field_0xc;
     undefined field_0xd;
-    undefined field_0xe;
+    // 0xE: If set to 1, artificial weather abilities will be activated on this turn.
```
UsernameFodder, line 27: Use `true` and `false` when talking about `bool`s

headers/types/dungeon_mode/dungeon.h
```diff
@@ -3354,10 +3235,8 @@ struct dungeon {
     undefined field_0x3e1d;
     undefined field_0x3e1e;
     undefined field_0x3e1f;
-    undefined field_0x3e20;
-    undefined field_0x3e21;
-    undefined field_0x3e22;
-    undefined field_0x3e23;
+    // 0x3E20: Number of valid monster spawn entries.
```
UsernameFodder, line 3238: Let's add `See spawn_entries`.

symbols/overlay29.yml
UsernameFodder: I don't think this was fixed

symbols/overlay29.yml
```diff
@@ -412,13 +474,45 @@ overlay29:
...
+        No params.
+    - name: FloorIsOver
+      address:
+        NA: 0x22EC7E8
+        EU: 0x22ED198
+        JP: 0x22EDE50
+      description: |-
+        Checks if the current floor should end, and updates dungeon::floor_loop_status if required.
+        If the player has been defeated, sets dungeon::floor_loop_status to floor_loop_status::FLOOR_LOOP_LEADER_FAINTED.
+        If dungeon::end_floor_flag is 1 or 2, sets sets dungeon::floor_loop_status to floor_loop_status::FLOOR_LOOP_NEXT_FLOOR.
```
UsernameFodder, line 513: You wrote "sets" twice

symbols/overlay29.yml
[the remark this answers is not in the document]
UsernameFodder: It doesn't look like this was changed?

UsernameFodder (changes requested):

symbols/arm9.yml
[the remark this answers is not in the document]
Metalcape: Then I think it's fine to rename it. How should I reference the fact that it's probably a debug switch in the description? Something like:
```
This function seems to be a debug switch that the developers may have used to disable the random enemy spawn. 
If it returned false, the call to SpawnMonster inside TrySpawnMonsterAndTickSpawnCounter would not be executed.
```
would be ok?

symbols/overlay29.yml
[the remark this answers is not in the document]
Metalcape: Yeah that's better, I just assumed that the check is never used but we don't actually know for sure

symbols/overlay29.yml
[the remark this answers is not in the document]
Metalcape: Sorry, I forgot

symbols/overlay29.yml
Metalcape: Forgot about this too, I really need to make smaller commits

symbols/overlay29.yml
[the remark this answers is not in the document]
Metalcape: Would something like this be ok?
```
Returns true if the monster has the blinded status, which is represented by status_id::STATUS_BLINKER (statuses::blinded set to 1). [...]
```

symbols/arm9.yml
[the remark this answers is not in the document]
UsernameFodder: Yeah that sounds good to me. Just make sure to mention that it always returns true somewhere in there.

symbols/overlay29.yml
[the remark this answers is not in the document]
UsernameFodder: That's a bit wordy. `statuses::blinded` already has a comment referencing `STATUS_BLINKER`, so let's just write:

> Returns true if the monster has the blinded status (see statuses::blinded). [...]

headers/functions/overlay29.h
```diff
@@ -295,5 +321,6 @@ void DisplayMessageInternal(int message_id, bool wait_for_input, undefined4 para
                             undefined4 param_4, undefined4 param_5, undefined4 param_6);
 void EuFaintCheck(bool non_team_member_fainted, bool set_unk_byte);
 uint8_t GetMinimapDataE447(void);
+bool UpdateMapSurveyorFlag(void);
```
UsernameFodder, line 324: This is out of order now that we merged NA/EU.

UsernameFodder (changes requested):

symbols/overlay29.yml
Metalcape: I noticed that `quicksave_flag`'s comment is also outdated. would changing it to this be ok?
```C
// 0x6: If equal to 1 or 2, the floor will be advanced at the end of the turn, 
// unless the leader fainted.
uint8_t end_floor_flag;
// 0x7: If set in tandem with either end_floor_flag or end_floor_no_death_check_flag, 
// a quicksave will occur instead of a floor advance.
bool quicksave_flag;
```

UsernameFodder: Yeah that sounds good.

UsernameFodder (approved):
Alright, all looks good to me, thanks! Also thanks to End for helping with reviews!

PR #54, opened 2022-06-07
Info dump 6-7/63: Types and dungeon cycle

End45:

- Types and effectiveness
- Dungeon mode cycle (floors, turns, etc.)

I've also renamed `FloorIsOver` to `IsFloorOver` since I think the question word is supposed to go first in these cases.


End45:
Looks like VSCode automatically gets rid of those spaces as you type. I can't find a setting to disable that behavior.
Could this rule be removed? I don't see why those spaces are required.

UsernameFodder:
> Looks like VSCode automatically gets rid of those spaces as you type. I can't find a setting to disable that behavior. Could this rule be removed? I don't see why those spaces are required.

I'm pretty sure the indentation on blank lines within an already-indented multiline string are required by the [YAML spec](https://yaml.org/spec/1.2.2/#61-indentation-spaces). The deserializer the formatter is using thankfully appears to be able to handle when the indentation is omitted, but it's supposed to be there. I don't think we should change this behavior.

There should be a way to disable it in VSCode though. I don't know exactly what your settings look like, but if this is an issue with auto-formatting on save, you could try adding `"files.trimTrailingWhitespace": false` to your local settings? (the opposite of [these instructions](https://stackoverflow.com/questions/30884131/remove-trailing-spaces-automatically-or-with-a-shortcut))

symbols/arm9.yml
```diff
@@ -901,6 +901,19 @@ arm9:
         r0: the effects bitvector to test
         r1: exclusive item effect ID
         return: bool
+    - name: ApplyGummiBoostsGround
```
UsernameFodder, line 904: Does "Ground" refer to ground mode? If so, can we call this "ApplyGummiBoostsGroundMode"? "Ground" makes me think of the type.

symbols/arm9.yml
```diff
@@ -1883,6 +1896,15 @@ arm9:
...
         return: bool
+    - name: GetMaxRescueAttempts
+      address:
+        EU: 0x20516B8
+        NA: 0x2051380
+      description: |-
+        Returns the maximum rescue attempts allowed in the specified dungeon.
+        
+        r0: dungeon id
+        return: Max rescue attempts, or -1 if rescues are disabled.
```
UsernameFodder, line 1907: This can't be -1 if the return type is `uint8_t`. Do you know how the game checks this value (is it a < 0 check or a != 0xFF check)? I can't tell. If we aren't sure, I'd say to change the return type to `int8_t`.

symbols/overlay29.yml
```diff
@@ -214,6 +218,11 @@ overlay29:
         Returns true if you should get kicked out of the dungeon if an important team member (like the partner or certain story allies) faints.
         
         return: dungeon::nonstory_flag || dungeon::hidden_land_flag
+    - name: FadeToBlack
+      address:
+        EU: 0x22E12A8
+        NA: 0x22E0968
+      description: Fades the screen to black across several frames.
```
UsernameFodder, line 225: Specify "No params." to make it obvious without needing to look at the headers. Saying nothing could mean that we don't know.

symbols/overlay29.yml
```diff
@@ -486,9 +503,18 @@ overlay29:
         NA: 0x22EBD08
         JP: 0x22ED370
       description: |-
-        The main function which executes the actions that take place in a fractional turn. Called in a loop by InitializeDungeon while FloorIsOver returns false.
+        The main function which executes the actions that take place in a fractional turn. Called in a loop by InitializeDungeon while IsFloorOver returns false.
         
         r0: first loop flag (true when the function is first called during a floor)
+    - name: LeaderTurn
```
UsernameFodder, line 509: Can we call this "RunLeaderTurn" to match the other similar functions?

symbols/overlay29.yml
```diff
@@ -511,6 +537,11 @@ overlay29:
         If dungeon::end_floor_flag is 1 or 2, sets dungeon::floor_loop_status to floor_loop_status::FLOOR_LOOP_NEXT_FLOOR.
         
         return: true if the current floor should end
+    - name: DecrementWindCounter
+      address:
+        EU: 0x22ED4F8
+        NA: 0x22ECB48
+      description: "Decrements dungeon::wind_turns and displays a wind warning message if required."
```
UsernameFodder, line 544: No params.

symbols/overlay29.yml
```diff
@@ -817,6 +889,14 @@ overlay29:
         r0: pointer to entity
         r1: speed counter weight
         return: speed stage
+    - name: CalcSpeedStageWrapper
+      address:
+        EU: 0x2300978
+        NA: 0x22FFF4C
+      description: |-
+        Calls CalcSpeedStage with r1 = 1.
```
UsernameFodder, line 897: Since we know what r1 is, let's say "with a speed counter of 1"

symbols/overlay29.yml
```diff
@@ -817,6 +889,14 @@ overlay29:
...
+
+        r0: pointer to entity
```
UsernameFodder, line 899: return: speed stage

symbols/overlay29.yml
```diff
@@ -1758,7 +1847,7 @@ overlay29:
         
         r0: attacker pointer
         r1: defender pointer
-    - name: ApplyGummiBoosts
+    - name: ApplyGummiBoostsDungeon
```
UsernameFodder, line 1850: Even though it's a long name, let's call this ApplyGummiBoostsDungeonMode to be consistent with ApplyGummiBoostsGroundMode

symbols/overlay29.yml
```diff
@@ -500,7 +526,7 @@ overlay29:
         First it calls TrySpawnMonsterAndTickSpawnCounter, then tries to activate the Plus and Minus abilities for both allies and enemies, and finally calls TryForcedLoss.
         
         No params.
-    - name: FloorIsOver
+    - name: IsFloorOver
```
UsernameFodder, line 529: Honestly I'm terrible at staying consistent about this 😅. No objections on this change though.

UsernameFodder (changes requested):

symbols/overlay29.yml
```diff
@@ -332,6 +341,14 @@ overlay29:
         
         r0: pointer to the status turn counter
         return: new counter value
+    - name: NextFrame
```
Metalcape, line 344: I don't know if it's 100% consistent, but all functions seem to have verbs in their names, so I think we should rename this to something like `AdvanceToNextFrame`

symbols/overlay29.yml
```diff
@@ -486,9 +503,18 @@ overlay29:
         NA: 0x22EBD08
         JP: 0x22ED370
       description: |-
-        The main function which executes the actions that take place in a fractional turn. Called in a loop by InitializeDungeon while FloorIsOver returns false.
+        The main function which executes the actions that take place in a fractional turn. Called in a loop by InitializeDungeon while IsFloorOver returns false.
```
Metalcape, line 506: `InitializeDungeon` -> `RunDungeon` to be consistent with the name change

headers/functions/overlay29.h
```diff
@@ -72,6 +83,7 @@ int GetMovePower(struct entity* entity, struct move* move);
 void AddExpSpecial(struct entity* attacker, struct entity* defender, int base_exp);
 void EnemyEvolution(struct entity* enemy);
 void EvolveMonster(struct entity* monster, undefined4* param_2, enum monster_id new_monster_id);
+bool DisplayActions(undefined param_1);
```
Metalcape, line 86: The parameter gets written to `dungeon::current_active_entity` under certain conditions, so I'm pretty sure we can define it as a `struct entity*`

symbols/overlay29.yml
Metalcape: Just wanted to point out that r1 is the speed counter _weight_, to avoid confusion

symbols/arm9.yml
End45: Oh right, the amount of rescues is signed (the game checks if it's < 0). I'll change it here and in the dungeon struct.

symbols/overlay29.yml
End45: Alright

symbols/overlay29.yml
End45: Fixed

symbols/overlay29.yml
End45: Makes sense

headers/functions/arm9.h
```diff
@@ -83,6 +83,8 @@ void ApplyExclusiveItemStatBoosts(enum item_id item_id, uint8_t* atk_boost, uint
                                   uint8_t* def_boost, uint8_t* sp_def_boost);
 void SetExclusiveItemEffect(uint32_t* effect_flags, enum exclusive_item_effect_id effect_id);
 bool ExclusiveItemEffectFlagTest(uint32_t* effect_flags, enum exclusive_item_effect_id effect_id);
+void ApplyGummiBoostsGround(undefined2* param_1, undefined2* param_2, undefined* param_3,
+                            undefined* param_4, undefined2 param_5, void* buffer);
```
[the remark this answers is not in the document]
End45: Right, I missed that one. Thanks!

symbols/arm9.yml
```diff
@@ -901,6 +901,19 @@ arm9:
...
+        NA: 0x201189C
+      description: |-
+        Applies the IQ boosts from eating a Gummi to the target monster.
+        
+        r0: Pointer to something
+        r1: Pointer to something
+        r2: Pointer to something
+        r3: Pointer to something
+        stack[0]: ?
+        stack[1]: Pointer to a buffer to store some result into
```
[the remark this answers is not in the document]
End45: Yup, fixed.

symbols/overlay29.yml
End45: `AdvanceFrame` perhaps?

symbols/overlay29.yml
End45: Fixed, thanks.

headers/functions/overlay29.h
End45: You're right, I'll update it.

symbols/overlay29.yml
UsernameFodder: Is it always so easy? I've certainly had experiences with more complicated functions where it's not always super obvious how many parameters there are. There's not always an explicit `mov` into registers before function calls if those registers are already populated beforehand by previous operations, and Ghidra sometimes infers the parameter list incorrectly. I've also quite often run into the conundrum of "is this function returning something, or does the end of the computation just happen to leave something in r0?" that can only be resolved by looking at the calling context.

I'm not sure what would be better as the "implicit default" when no parameter info is listed. My initial thought was to align ease of typing with ease of research: the less you know/the less work you've done, the less you need to type. That makes it easy to be lazy if you don't want to properly vet the parameter list of every single function while actively doing research, especially if you don't care that much about a particular function. It also means that if you forget to write "No params.", the meaning defaults to missing info, which isn't ideal but isn't _incorrect_. If it were the other way around, the meaning would default to no params, which could actually be wrong.

In practice, I do agree it's somewhat annoying to type "No params." We should optimize for the common, long-term case. And I guess my feeling is that the common case is going to be active research where there are a lot more unknowns and incomplete knowledge. I think your info dumps are somewhat of a special case.

End45: > Is it always so easy? I've certainly had experiences with more complicated functions where it's not always super obvious how many parameters there are.

The easiest way to check is usually to look at the context in which the function is called, since all the parameters have to be initialized there. Checking what the function does with the paramters is a good choice most of the time as well. The main problem comes if a function takes a parameter but doesn't use it or when a function takes paramters in the stack, since those stay there and aren't accessed until the function needs them.

> I've also quite often run into the conundrum of "is this function returning something, or does the end of the computation just happen to leave something in r0?"

Oh yeah, that one is harder to tell. You could have a function that returns something but in the current context the function that calls it doesn't use the return value for anything.

> We should optimize for the common, long-term case. And I guess my feeling is that the common case is going to be active research where there are a lot more unknowns and incomplete knowledge. I think your info dumps are somewhat of a special case.

Eh, also true. I don't have a strong preference for either option, we could leave it this way.

UsernameFodder: > The easiest way to check is usually to look at the context in which the function is called, since all the parameters have to be initialized there. Checking what the function does with the paramters is a good choice most of the time as well. The main problem comes if a function takes a parameter but doesn't use it or when a function takes paramters in the stack, since those stay there and aren't accessed until the function needs them.

I'm glad there isn't an easier way that I'm just missing 😆. There are other problems I've run into related to optimizations that I alluded to. For example, if you have a chain of function calls (either nested of on the same level but right after one another), `r0` might get set at the top of the call stack and "pass through" to subsequent function calls, and if you're just looking at one of the intermediate functions, looking at the immediate calling code isn't going to help. The same thing happens when you have the return value of one function immediately passed into another function in r0.

UsernameFodder (approved):
LGTM. I'll give a bit of time before merging this in case @Metalcape has any additional comments.

Metalcape:
Everything looks good to me 👍

PR #55, opened 2022-06-14
Info dump 8-9/63: AI and move animations

End45:

- AI
- Move animations


End45:
Alright, I assumed this was going to cause some checks to fail.
There's a section at the end of arm9.bin, which starts at [EU]0xB3CC0 / [US]0xB3380, that gets loaded in a different RAM area, starting at [EU/US]0x1FF8000, which also seems to be mirrored at [EU/US]0x0.

How should this be handled?

Metalcape:
> How should this be handled?

I think this is the case where it's possible to define multiple blocks inside the same symbol file, so something like this:
```YAML
arm9_main:
  versions:
    - EU
    - NA
    - JP
  address:
    EU: 0x2000000
    NA: 0x2000000
    JP: 0x2000000
  length:
    EU: 0xB7D38
    NA: 0xB73F8
    JP: 0xB8CB8
  description: [...]
  functions: [...]
  data: [...]
arm9_extra:
  versions:
    - EU
    - NA
  address:
    EU: 0x21FF8000
    NA: 0x21FF8000
  length:
    EU: ?
    NA: ?
  description: [...]
  functions: [...]
  data: [...]
```

symbols/overlay29.yml
```diff
@@ -970,6 +1141,24 @@ overlay29:
...
+        Returns whether a certain monster can see other invisible monsters. To be precise, this function returns true if the monster is holding Goggle Specs or if it has the status status::STATUS_EYEDROPS.
+        
+        r0: Entity pointer
+        return: True if the monster can see invisible monsters.
+    - name: HasDropeyeStatus
+      address:
+        EU: 0x230297C
+        NA: 0x2301F50
+      description: |-
+        Returns whether a certain monster is under the effect of the status::STATUS_DROPEYE status.
```
UsernameFodder, line 1158: "the status::STATUS_DROPEYE status" -> "status::STATUS_DROPEYE" for conciseness

symbols/overlay29.yml
```diff
@@ -1806,6 +2004,46 @@ overlay29:
         
         r0: pointer to entity
         return: int
+    - name: AiConsiderMove
+      address:
+        EU: 0x231A2E0
+        NA: 0x2319880
+      description: |-
+        The AI uses this function to check if a move has any potential targets, to calculate the list of potential targets and to calculate the move's special weight. This weight will be higher if the pokémon has weak-type picker and the target is weak to the move (allies only, enemies always get a result of 1 even if the move is super effective). More things could affect the result.
```
UsernameFodder, line 2012: Elaborate in separate paragraph.

headers/functions/overlay29.h
```diff
@@ -181,6 +204,13 @@ void RestoreMovePP(struct entity* user, struct entity* target, int pp, bool supp
 void SetReflectDamageCountdownTo4(struct entity* entity);
 bool HasConditionalGroundImmunity(struct entity* entity);
 int Conversion2IsActive(struct entity* entity);
+int AiConsiderMove(struct ai_possible_move* ai_possible_move, struct entity* monster,
+                   struct move* move);
+int TryAddTargetToAiTargetList(int current_num_targets, struct move_target_and_range,
```
UsernameFodder, line 209: Second param is missing a name

headers/functions/overlay29.h
```diff
@@ -181,6 +204,13 @@ void RestoreMovePP(struct entity* user, struct entity* target, int pp, bool supp
...
+                               struct entity* user, struct entity* target, struct move* move,
+                               bool check_all_conditions);
+bool IsAiTargetEligible(struct move_target_and_range, struct entity* user, struct entity* target,
```
UsernameFodder, line 212: First param is missing a name

symbols/overlay29.yml
```diff
@@ -1959,6 +2197,29 @@ overlay29:
         
         r0: entity pointer
         r1: base success percentage (100*p). 0 is treated specially and guarantees success.
+    - name: CanAiUseMove
+      address:
+        EU: 0x23254DC
+        NA: 0x2324A74
+      description: |-
+        Checks if an AI-controlled monster can use a move. Will return false if the any of the flags move::f_exists, move::f_subsequent_in_link_chain or move::f_disabled is true. The function does not check if the flag move::f_enabled_for_ai is set. This function also returns true if the call to CanMonsterUseMove is false.
```
UsernameFodder, line 2205: Elaborate in separate paragraph.

symbols/overlay29.yml
```diff
@@ -1959,6 +2197,29 @@ overlay29:
...
+        r0: Entity pointer
+        r1: Move index
+        r2: extra_checks parameter when calling CanMonsterUseMove
+        return: True if the AI can use the move (not accounting for move::f_enabled_for_ai)
+    - name: CanMonsterUseMove
+      address:
+        EU: 0x232558C
+        NA: 0x2324B24
+      description: |-
+        Checks if a monster can use the given move. Will always return true for the regular attack. Will return false if the move if the flag move::f_disabled is true, if the flag move::f_sealed is true. More things will be checked if the extra_checks parameter is true.
```
UsernameFodder, line 2217: Elaborate in separate paragraph.

symbols/overlay29.yml
```diff
@@ -1978,7 +2239,28 @@ overlay29:
         Gets the lower 2 bytes of a 4-byte number and interprets it as a signed short.
         
         r0: 4-byte number x
-        return: (short)x
```
UsernameFodder, line 1981: Doesn't really matter much, but I was envisioning `(short)x` as a cast operation. I don't think this needs changing.

symbols/overlay29.yml
```diff
@@ -1978,7 +2239,28 @@ overlay29:
...
+        return: (short) x
+    - name: GetMoveAnimationId
+      address:
+        EU: 0x2326578
+        NA: 0x2325B10
+      description: |-
+        Returns the move animation ID that should be played for a move. It contains a check for weather ball. After that, if the parameter should_play_alternative_animation is false, the move ID is returned. If it's true, there's a bunch of manual ID checks that result on a certain hardcoded return value.
```
UsernameFodder, line 2248: Elaborate in separate paragraph

UsernameFodder (changes requested):

UsernameFodder:
> Alright, I assumed this was going to cause some checks to fail. There's a section at the end of arm9.bin, which starts at [EU]0xB3CC0 / [US]0xB3380, that gets loaded in a different RAM area, starting at [EU/US]0x1FF8000, which also seems to be mirrored at [EU/US]0x0.
> 
> How should this be handled?

0x0 and 0x1FF8000 aren't part of main memory, they're within the [instruction TCM](https://problemkaputt.de/gbatek.htm#dsmemorymaps) and its upper address mirrors (GBATEK says the TCM doesn't have mirrors, but it seems that DeSmuME and No$GBA both fill in undefined areas with mirrors anyway). To my understanding, the instruction TCM is similar to modern day instruction caches (and similarly, the data TCM is like data caches).

I think we should label the actual section within `arm9.bin` at [EU]0xB3CC0 / [NA]0xB3380. This is what will show up in Ghidra, and what would make sense to link against in CoT patches. If we want to document that this bit is loaded into the instruction TCM, maybe this would be a good place to create a subregion. It seems like the length is 0x4000 in the NA version (the other half looks to be all zeros).

symbols/arm9.yml
```diff
@@ -975,6 +1019,15 @@ arm9:
         
         r0: move pointer
         return: base power
+    - name: IsMoveRangeString19
```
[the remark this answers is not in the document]
End45: I put 19 because there's multiple strings that have "user" as the range

symbols/arm9.yml
```diff
@@ -1951,6 +2004,15 @@ arm9:
         
         r0: monster id
         return: sprite file size
+    - name: GetCanMoveFlag
```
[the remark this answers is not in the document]
End45: All the other functions that get data from monster.md start with `Get`, that's why I used it here too. Should I change it?

symbols/overlay29.yml
```diff
@@ -236,6 +236,26 @@ overlay29:
         
         r0: pointer to entity
         returns: pointer to tile
+    - name: CanTargetEntity
+      address:
+        EU: 0x22E3150
+        NA: 0x22E2810
+      description: |-
+        Checks if a monster can target another entity when controlled by the AI. Checks if the target is invisible, if the user can see invisible monsters, if the user is blinded and if the target position is in sight from the position of the user (this last check is done by calling IsPositionInSight with the user's and the target's position).
```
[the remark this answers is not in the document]
End45: Yeah, it will look better that way. I'll split it.

symbols/overlay29.yml
```diff
@@ -326,6 +346,37 @@ overlay29:
...
+      address:
+        EU: 0x22E99D0
+        NA: 0x22E9020
+      description: |-
+        Gets the direction in which a monster should move to go from the origin position to the target position
+        
+        r0: Origin position
+        r1: Target position
+        return: Direction in which to move to reach the target position from the origin position
+    - name: GetDistance
```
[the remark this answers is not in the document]
End45: Oh, so this distance does have a name. I was wondering if there was a more specific way to refer to it when I wrote this. I'll change it.

symbols/overlay29.yml
```diff
@@ -326,6 +346,37 @@ overlay29:
...
+      address:
+        EU: 0x22E9A3C
+        NA: 0x22E908C
+      description: |-
+        Returns the distance between two positions. Calculated as max(abs(x0-x1), abs(y0,y1)).
```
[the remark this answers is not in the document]
End45: Fixed, thanks

symbols/overlay29.yml
```diff
@@ -326,6 +346,37 @@ overlay29:
...
+        
+        r0: Position A
+        r1: Position B
+        return: Distance between position A and position B
+    - name: IsPositionInSight
+      address:
+        EU: 0x22E9B54
+        NA: 0x22E91A4
+      description: |-
+        Checks if a given target position is in sight from a given origin position. There's multiple factors that affect this check, but generally, it's true if both positions are in the same room or within 2 tiles of each other.
```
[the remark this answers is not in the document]
End45: :+1:

symbols/overlay29.yml
```diff
@@ -500,6 +551,51 @@ overlay29:
         
         r0: pointer to user entity
         r1: pointer to target entity
+    - name: ClearMonsterActionFields
+      address:
+        EU: 0x22EBDA4
+        NA: 0x22EB3F4
+      description: |-
+        Clears the fields related to AI in the monster's data struct, setting them all to 0. Specifically, monster::action_id, monster::action_use_idx and monster::field_0x54 are cleared.
```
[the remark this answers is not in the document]
End45: :+1:

symbols/overlay29.yml
```diff
@@ -500,6 +551,51 @@ overlay29:
...
+        
+        r0: Pointer to the monster's action_id field (this field is probably contained in a struct)
+    - name: SetMonsterActionFields
+      address:
+        EU: 0x22EBDB8
+        NA: 0x22EB408
+      description: |-
+        Sets some the fields related to AI in the monster's data struct. Specifically, monster::action_id, monster::action_use_idx and monster::field_0x54. The last 2 are always set to 0.
```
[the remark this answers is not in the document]
End45: :+1:

headers/functions/overlay29.h
```diff
@@ -35,18 +40,25 @@ void DungeonRngUnsetSecondary(void);
 void DungeonRngSetSecondary(int i);
 void DungeonRngSetPrimary(void);
 void TrySwitchPlace(struct entity* user, struct entity* target);
-void SetForcedLossReason(enum forced_loss_reason forced_loss_reason);
+void ClearMonsterActionFields(void* monster_action_field);
+void SetMonsterActionFields(void* monster_action_field);
```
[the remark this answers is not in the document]
End45: You're right, fixed, thanks.

headers/functions/overlay29.h
```diff
@@ -35,18 +40,25 @@ void DungeonRngUnsetSecondary(void);
...
+void SetActionPassTurnOrWalk(void* monster_action_field, enum monster_id monster_id);
+void SetActionUseMoveAi(void* monster_action_field, uint8_t move_index,
+                        struct direction_id_8 direction);
+void SetActionRegularAttack(void* monster_action_field, struct direction_id_8 direction);
```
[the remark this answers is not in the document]
End45: Fixed, thanks

headers/functions/overlay29.h
```diff
@@ -35,18 +40,25 @@ void DungeonRngUnsetSecondary(void);
 void DungeonRngSetSecondary(int i);
 void DungeonRngSetPrimary(void);
 void TrySwitchPlace(struct entity* user, struct entity* target);
-void SetForcedLossReason(enum forced_loss_reason forced_loss_reason);
+void ClearMonsterActionFields(void* monster_action_field);
+void SetMonsterActionFields(void* monster_action_field);
+void SetActionPassTurnOrWalk(void* monster_action_field, enum monster_id monster_id);
+void SetActionUseMoveAi(void* monster_action_field, uint8_t move_index,
```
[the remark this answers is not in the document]
End45: Oh, so the struct version should only be used when the enum type shows up in a struct definition?

symbols/overlay29.yml
```diff
@@ -928,6 +1050,55 @@ overlay29:
         r1: format
         ...: variadic
         return: number of characters printed, excluding the null-terminator
+    - name: IsMonsterConrnered
```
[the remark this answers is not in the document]
End45: Fixed, thanks

symbols/overlay29.yml
```diff
@@ -928,6 +1050,55 @@ overlay29:
...
+        True if the given monster is cornered (it can't move in any direction)
+        
+        r0: Entity pointer
+        return: True if the monster can't move in any direction, false otherwise.
+    - name: CanAttackInDirection
+      address:
+        EU: 0x2301C60
+        NA: 0x2301234
+      description: |-
+        Returns whether a monster can attack in a given direction. The check fails if the destination tile is impassable, contains a monster that isn't of type entity_type::ENTITY_MONSTER or if the monster can't directly move from the current tile into the destination tile.
```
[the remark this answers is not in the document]
End45: :+1:

headers/functions/overlay29.h
```diff
@@ -70,13 +83,22 @@ void InitTeamMember(enum monster_id, enum type_id type_1, enum type_id type_2,
                     struct team_member* team_member_data, undefined param_5, undefined param_6,
                     undefined param_7, undefined param_8, undefined param_9);
 void ExecuteMonsterAction(struct entity* monster);
+bool HasStatusThatPreventsActing(struct entity* monster);
 int CalcSpeedStage(struct entity* entity, int counter_weight);
 int CalcSpeedStageWrapper(struct entity* entity);
 int GetNumberOfAttacks(struct entity* entity);
+bool IsMonsterConrnered(struct entity* monster);
+bool CanAttackInDirection(struct entity* monster, struct direction_id_8 direction);
```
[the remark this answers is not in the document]
End45: :+1:

headers/functions/overlay29.h
```diff
@@ -70,13 +83,22 @@ void InitTeamMember(enum monster_id, enum type_id type_1, enum type_id type_2,
...
+bool CanAiMonsterMoveInDirection(struct entity* monster, struct direction_id_8 direction,
```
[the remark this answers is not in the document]
End45: :+1:

symbols/overlay29.yml
```diff
@@ -928,6 +1050,55 @@ overlay29:
...
+        
+        r0: Entity pointer
+        r1: Direction
+        return: True if the monster can attack into the tile adjacent to them in the specified direction, false otherwise.
+    - name: CanAiMonsterMoveInDirection
+      address:
+        EU: 0x2301D24
+        NA: 0x23012F8
+      description: |-
+        Checks whether an AI-controlled monster can move in the specified direction. Accounts for walls, other monsters on the target position and IQ skills that might prevent a monster from moving into a specific location, such as House Avoider, Trap Avoider or Lava Evader.
```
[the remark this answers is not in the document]
End45: :+1:

headers/functions/overlay29.h
```diff
@@ -70,13 +83,22 @@ void InitTeamMember(enum monster_id, enum type_id type_1, enum type_id type_2,
...
+                                 struct entity* out_monster_in_target_position);
+bool ShouldMonsterRunAway(struct entity* monster);
+bool ShouldMonsterRunAwayVariation(struct entity* monster);
```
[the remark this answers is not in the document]
End45: Fixed, thanks

symbols/overlay29.yml
```diff
@@ -970,6 +1141,24 @@ overlay29:
         r0: entity pointer
         r1: type ID
         return: bool
+    - name: CanSeeInvisibleMonsters
+      address:
+        EU: 0x2302918
+        NA: 0x2301EEC
+      description: |-
+        Returns whether a certain monster can see other invisible monsters. To be precise, this function returns true if the monster is holding Goggle Specs or if it has the status status::STATUS_EYEDROPS.
```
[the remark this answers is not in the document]
End45: :+1:

symbols/overlay29.yml
End45: :+1:

symbols/overlay29.yml
End45: :+1:

headers/functions/overlay29.h
End45: Fixed! I didn't even know that was valid syntax

headers/functions/overlay29.h
End45: Fixed

symbols/overlay29.yml
End45: :+1:

symbols/overlay29.yml
End45: :+1:

symbols/overlay29.yml
End45: I thought it was usual to put a space between the casting operator and the variable, is that not the case?

symbols/overlay29.yml
End45: :+1:

End45:
> I think we should label the actual section within `arm9.bin` at [EU]0xB3CC0 / [NA]0xB3380.

What do you mean by this exactly? The YAML file needs to contain the RAM offsets, right? Are you suggesting we label them as if they were loaded alongside the rest of the file? If so, that's going to cause problems when debugging, since there's other data in the RAM past [EU]0x20B3CC0 and since No$GBA won't be able to match function calls with their names if the symbol table doesn't have the right RAM offsets.

headers/functions/overlay29.h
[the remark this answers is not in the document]
UsernameFodder: Yeah, that's the only place the data width actually matters. With functions it's always a 32-bit register value anyway.

symbols/arm9.yml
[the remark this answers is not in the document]
UsernameFodder: Oh, alright then, this is fine as is.

symbols/arm9.yml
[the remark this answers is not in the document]
UsernameFodder: I don't feel strongly about it. The current name is okay.

symbols/overlay29.yml
UsernameFodder: I guess it's probably a style convention thing, but I usually see it without a space.

headers/functions/overlay29.h
UsernameFodder: Yeah in C function declarations don't need parameter names. But some styles call for it for neatness, and it also helps Ghidra to include names.

UsernameFodder:
> > I think we should label the actual section within `arm9.bin` at [EU]0xB3CC0 / [NA]0xB3380.
> 
> What do you mean by this exactly? The YAML file needs to contain the RAM offsets, right? Are you suggesting we label them as if they were loaded alongside the rest of the file? If so, that's going to cause problems when debugging, since there's other data in the RAM past [EU]0x20B3CC0 and since No$GBA won't be able to match function calls with their names if the symbol table doesn't have the right RAM offsets.

I was suggesting that we just label the offsets as they are right after arm9.bin gets loaded to 0x2000000 (before any of this weird runtime loading stuff happens), so `ShouldMonsterRunAwayVariationOutlawCheck` would go at 0x20B5710, for example.

I guess that would work for Ghidra but not for No$GBA. I'm not really sure what to do here...it seems like Ghidra/No$GBA are going to be at odds with each other unless we somehow duplicate all the symbol entries in both places somehow? How does the game actually call into the functions at these locations? Is there actually a branch that takes the PC out of main memory? Also, how common are cases like these where things aren't loaded to their standard location within the binary?

End45:
> How does the game actually call into the functions at these locations? Is there actually a branch that takes the PC out of main memory?

Yup, it branches to addresses in the 0x1FF8000 area. So I don't think putting these symbols in their "intended" addresses would would with Ghidra either, since calls to those functions don't jump into the 0x20B3CC0 area and therefore wouldn't be recognized as calls to these fuctions at the end of the file.

> Also, how common are cases like these where things aren't loaded to their standard location within the binary?

I think this is the only case where this happens.

UsernameFodder:
> > How does the game actually call into the functions at these locations? Is there actually a branch that takes the PC out of main memory?
> 
> Yup, it branches to addresses in the 0x1FF8000 area. So I don't think putting these symbols in their "intended" addresses would would with Ghidra either, since calls to those functions don't jump into the 0x20B3CC0 area and therefore wouldn't be recognized as calls to these fuctions at the end of the file.

Could you point me to the addresses with the branch instructions from main memory to the TCM? I was having a hard time finding them. I still find it weird that it would be branching to the undefined 0x1FF8000 area (according to GBATEK) rather than to 0x0, so I'm curious what it looks like.

We would still want to label the functions in the 0x20B3CC0 area since that's where you would look at the disassembly/decompilation in Ghidra. It would at least work for branches between functions in the area.

I thought a bit more and I think I have a solution for how to record these offsets, assuming 0xB3CC0 in arm9.bin is exactly copied to 0x1FF8000 (need to verify this). I believe the region should have a length of 0x4000. We could define special versions "EU-ITCM" and "NA-ITCM" since the instruction TCM has a well defined location and length. I think this would also be a good place to use a subregion. So something like this:

## `arm9.yml`
```yml
arm9:
  versions:
    - EU
    - NA
    - JP
    - EU-ITCM
    - NA-ITCM
  address:
    EU: 0x2000000
    NA: 0x2000000
    JP: 0x2000000
    EU-ITCM: 0x1FF8000
    NA-ITCM: 0x1FF8000
  length:
    ...
    EU-ITCM: 0x4000
    NA-ITCM: 0x4000
  subregions:
    - itcm.yml
  ...
```

## `arm9/itcm.yml`
```yml
itcm:
  versions:
    - EU
    - NA
    - JP
    - EU-ITCM
    - NA-ITCM
  address:
    EU: 0x20B3CC0
    NA: 0x20B3CC0
    JP: 0x20B3CC0
    EU-ITCM: 0x1FF8000
    NA-ITCM: 0x1FF8000
  length:
    EU: 0x4000
    NA: 0x4000
    JP: 0x4000
    EU-ITCM: 0x4000
    NA-ITCM: 0x4000
  description: |-
    The instruction TCM (tightly-coupled memory) and the corresponding region in the ARM9 binary.

    The ITCM is a special area of low-latency memory meant for performance-critical routines. It's similar to an instruction cache, but more predictable. See the ARMv5 Architecture Reference Manual, Chapter B7 (https://developer.arm.com/documentation/ddi0100/i).

    In Explorers of Sky, a fixed region of the ARM9 binary appears to be loaded in the ITCM at all times, and seems to contain functions related to the dungeon AI.
  functions:
    - name: ShouldMonsterRunAwayVariationOutlawCheck
      address:
        EU: 0x20B5710
        NA: 0x20B5710
        EU-ITCM: 0x1FFA390
        NA-ITCM: 0x1FFA390
    ...
  ...
```

End45:
An example of an instruction that jumps into the TCM can be found at [EU]0x2308EEC = [EU]overlay29:0x2C36C

I like the idea of using subregions for this, looks like a good first use case to try them out. I suppose that symbols declared on a version like `x-ITCM` would be included in the symbol table for version `x` so it can be used with the rest of the symbols, right?

UsernameFodder:
> An example of an instruction that jumps into the TCM can be found at [EU]0x2308EEC = [EU]overlay29:0x2C36C

Ah, I found this note from [GBATEK](https://problemkaputt.de/gbatek.htm#dsmemorycontrolcacheandtcm):

> Note: Although ITCM is NOT moveable, the NDS Firmware configures the ITCM size to 32MB, and so, produces ITCM mirrors at 0..1FFFFFFh. Furthermore, the PU can be used to lock/unlock memory in that region. That trick allows to move ITCM anywhere within the lower 32MB of memory.

So I was mistaken; the region at 0x1FF8000 actually is valid. I assume the compiler optimizes its branching to this range rather than 0x0 since it's the closest mirror.

> I like the idea of using subregions for this, looks like a good first use case to try them out. I suppose that symbols declared on a version like `x-ITCM` would be included in the symbol table for version `x` so it can be used with the rest of the symbols, right?

It would produce separate symbol tables for the `x-ITCM` versions, as it does with EU/NA/JP. I think that should be fine? With Ghidra they can just be ignored since the ITCM region doesn't exist, and No$GBA symbol imports requires some manual copy-pasting anyway, so users can just copy from the `x-ITCM` tables if needed.

End45:
> I can't think of anything better than using versions

Yeah, me neither. I guess we'll have to do it that way.
But I still have a couple of questions:

1. What happens with branches from the overlays to the 0x1FF8000 area? Would Ghidra show those properly if we define the symbols past 0x20B3CC0?
2. Normally, to create a symbol table that can be used in No$GBA, you create a single .sym file by combining the .sym files of the different binaries you're going to work with. So with this system, you would combine EU/arm9.sym and EU-ITCM/arm9.sym, right? Wouldn't that cause issues since symbols would end up being defined twice in multiple addresses?

UsernameFodder:
> 1. What happens with branches from the overlays to the 0x1FF8000 area? Would Ghidra show those properly if we define the symbols past 0x20B3CC0?

Ghidra will just fail to resolve the branches to the ITCM. So for example, the branch at [EU]0x2308EEC will show as a branch to the 0x1FF8000 area, but you won't be able to double-click on it to jump to the referenced function, and the auto-analysis won't be able to link that branch to the referent. But if you know that a branch to 0x1FF80000 corresponds to code at 0x20B3CC0, you can just go to 0x20B3CC0 manually and look at the assembly there. And if the things at ﻿0x20B3CC0 are labeled, any branches between functions in the 0x20B3CC0 region will still resolve correctly since the relative offsets are the same as in the ITCM at runtime. So the decompiled code for those functions should still hopefully look okay.

> 2. Normally, to create a symbol table that can be used in No$GBA, you create a single .sym file by combining the .sym files of the different binaries you're going to work with. So with this system, you would combine EU/arm9.sym and EU-ITCM/arm9.sym, right? Wouldn't that cause issues since symbols would end up being defined twice in multiple addresses?

Nah duplicate symbol names are fine. A .sym file is a mapping from address to name; No$GBA doesn't care if the names are duplicated or not, it'll just apply whatever the associated name is when you navigate to that address. So you'd just see the same symbol names in the ITCM region and (incorrectly) at the end of ARM9. If the spurious labels at the end of ARM9 bother you, you can always just delete them from your .sym file, but it won't break No$GBA.

End45:
> > 1. What happens with branches from the overlays to the 0x1FF8000 area? Would Ghidra show those properly if we define the symbols past 0x20B3CC0?
> 
> Ghidra will just fail to resolve the branches to the ITCM. So for example, the branch at [EU]0x2308EEC will show as a branch to the 0x1FF8000 area, but you won't be able to double-click on it to jump to the referenced function, and the auto-analysis won't be able to link that branch to the referent. But if you know that a branch to 0x1FF80000 corresponds to code at 0x20B3CC0, you can just go to 0x20B3CC0 manually and look at the assembly there. And if the things at ﻿0x20B3CC0 are labeled, any branches between functions in the 0x20B3CC0 region will still resolve correctly since the relative offsets are the same as in the ITCM at runtime. So the decompiled code for those functions should still hopefully look okay.
> 
> > 2. Normally, to create a symbol table that can be used in No$GBA, you create a single .sym file by combining the .sym files of the different binaries you're going to work with. So with this system, you would combine EU/arm9.sym and EU-ITCM/arm9.sym, right? Wouldn't that cause issues since symbols would end up being defined twice in multiple addresses?
> 
> Nah duplicate symbol names are fine. A .sym file is a mapping from address to name; No$GBA doesn't care if the names are duplicated or not, it'll just apply whatever the associated name is when you navigate to that address. So you'd just see the same symbol names in the ITCM region and (incorrectly) at the end of ARM9. If the spurious labels at the end of ARM9 bother you, you can always just delete them from your .sym file, but it won't break No$GBA.

Oh, good to know.
I did a quick test and seems like the only issue shows up when using Ctrl+G to go to an address by typing a symbol name. No$GBA will always go to the first address defined for that symbol. So you just put the ITCM symbols first and that solves the problem. That's probably a detail that should be pointed out in the instructions on how to use the .sym files once we make this change.

Alright, so how do we implement this? I see you merged #57, so I guess we can specify the subregions already?

UsernameFodder:
> Oh, good to know. I did a quick test and seems like the only issue shows up when using Ctrl+G to go to an address by typing a symbol name. No$GBA will always go to the first address defined for that symbol. So you just put the ITCM symbols first and that solves the problem. That's probably a detail that should be pointed out in the instructions on how to use the .sym files once we make this change.

I didn't actually know you could use Ctrl+G in No$GBA lol. I suppose we could note that in the instructions; repeated symbol names isn't actually new; all of the symbols with multiple addresses end up as repeated names in the .sym files. You could add to the docs now if you want, or we can do it later in a follow-up change.

> Alright, so how do we implement this? I see you merged #57, so I guess we can specify the subregions already?

## Symbol tables
To implement this, first modify `arm9.yml`:
- Add `EU-ITCM` and `NA-ITCM` to the version list in `arm9.yml` (note that the order matters; I would put these special versions after JP)
- Update the address/length fields to reflect the new versions
- Add a `subregions` list to the `arm9` block that references the subregion file (which I propose we call `itcm.yml`)

Like this (diff notation for clarity):

### `arm9.yml`
```yml
arm9:
  versions:
    - EU
    - NA
    - JP
+   - EU-ITCM
+   - NA-ITCM
  address:
    EU: 0x2000000
    NA: 0x2000000
    JP: 0x2000000
+   EU-ITCM: 0x1FF8000
+   NA-ITCM: 0x1FF8000
  length:
    ...
+   EU-ITCM: 0x8000
+   NA-ITCM: 0x8000
  description: ...
+ subregions:
+   - itcm.yml
  ...
```

Then create the subregion file:

- Make a directory next to `arm9.yml` called `arm9/`
- Create the subregion file `itcm.yml` inside `arm9/`
- Use the same versions as `arm9.yml`, but for the top-level address/length fields, define the range of addresses encompassed by the subregion (we can probably omit the JP address/offset if we aren't sure about it)
- Add a description (draft below)
- Add the relevant symbols normally, listing the "static" addresses (in arm9.bin) under `EU`/`NA`, and the runtime addresses under `EU-ITCM`/`NA-ITCM`. Also move any existing symbols that fall in the subregion from `arm9.yml` to `itcm.yml` (the tests will enforce this).

### `arm9/itcm.yml`
```yml
itcm:
  versions:
    - EU
    - NA
    - JP
    - EU-ITCM
    - NA-ITCM
  address:
    EU: 0x20B3CC0
    NA: 0x20B3CC0
    EU-ITCM: 0x1FF8000
    NA-ITCM: 0x1FF8000
  length:
    EU: 0x4000
    NA: 0x4000
    EU-ITCM: 0x8000
    NA-ITCM: 0x8000
  description: |-
    The instruction TCM (tightly-coupled memory) and the corresponding region in the ARM9 binary.

    The ITCM is a special area of low-latency memory meant for performance-critical routines. It's similar to an instruction cache, but more predictable. See the ARMv5 Architecture Reference Manual, Chapter B7 (https://developer.arm.com/documentation/ddi0100/i).
    
    The Nintendo DS ITCM region is located at 0x0-0x7FFF in memory, but the 32 KiB segment is mirrored throughout the 16 MiB block from 0x0-0x1FFFFFF. The Explorers of Sky code seems to reference only the mirror at 0x1FF8000, the closest one to main memory.
    
    In Explorers of Sky, a fixed region of the ARM9 binary appears to be loaded in the ITCM at all times, and seems to contain functions related to the dungeon AI. While the ITCM has a capacity of 0x8000, only the first 0x4000 seems to be used, while the upper 0x4000 is all zeroes.
  functions:
    - name: ShouldMonsterRunAwayVariationOutlawCheck
      address:
        EU: 0x20B5710
        NA: 0x20B5710
        EU-ITCM: 0x1FFA390
        NA-ITCM: 0x1FFA390
    ...
  ...
```

## C headers
For the C headers, mirror the directory structure used for the symbol tables. So add an `#include` at the top of `arm9.h`:

### `arm9.h`
```c
#ifndef HEADERS_FUNCTIONS_ARM9_H_
#define HEADERS_FUNCTIONS_ARM9_H_

#include "arm9/itcm.h"

...
```

Then create a directory `arm9/` and add `itcm.h` inside:

### `arm9/itcm.h`
```c
#ifndef HEADERS_FUNCTIONS_ARM9_ITCM_H_
#define HEADERS_FUNCTIONS_ARM9_ITCM_H_

bool ShouldMonsterRunAwayVariationOutlawCheck(struct entity* monster, undefined param_2);
...
```

End45:
Alright! I'm a bit busy these days so I'll see when I can make the changes.
If you want to do it yourself, I have the option to allow edits by maintainers enabled, so you should be able to push to my branch as well.

End45:
Alright, that should be it. The only thing I've changed compared to what you posted is the value of the length fields. I don't know where that 0x4000 came from, since this special area in arm9.bin is 0x4080 bytes long. I also don't understand why you said we should specify 0x8000 as the length in some places, since we normally specify the size of the file (or in this case, the special area) and not the size of the memory area in which it's loaded.

As for the tests, it seems like it expects at least one symbol in the data section of itcm.yml, which should probably be changed if that's the case.

symbols/arm9/itcm.yml
```diff
@@ -0,0 +1,70 @@
...
+    - JP
+    - EU-ITCM
+    - NA-ITCM
+  address:
+    EU: 0x20B3CC0
+    NA: 0x20B3380
+    EU-ITCM: 0x1FF8000
+    NA-ITCM: 0x1FF8000
+  length:
+    EU: 0x4084
```
UsernameFodder, line 14: > I don't know where that 0x4000 came from, since this special area in arm9.bin is 0x4080 bytes long.

Are you sure about this? `arm9.bin` does go a bit past 0x4000, but if you actually look in RAM while the game is running, the stuff starting at an offset of 0x4000 is all zeroes, whereas `arm9.bin` itself has some addresses there. RAM and `arm9.bin` only match up to a length of 0x4000, which is why I think we should mark the length as 0x4000.

(Also, even if we are going until the end of `arm9.bin`, the length should be 0x4078, not 0x4084)

symbols/arm9/itcm.yml
```diff
@@ -0,0 +1,70 @@
...
+      address:
+        EU: 0x20B7318
+        NA: 0x20B69D8
+        EU-ITCM: 0x1FFB658
+        NA-ITCM: 0x1FFB658
+      description: |-
+        Determines if an AI-controlled monster will use a move and which one it will use
+        
+        r0: Entity pointer
+  data:
```
UsernameFodder, line 70: > it seems like it expects at least one symbol in the data section of itcm.yml, which should probably be changed if that's the case

To specify an empty data list, you can do `data: []`. We already do this in some of the overlay files.

symbols/arm9/itcm.yml
End45: Okay, you're right, not everything past 0x20B3CC0 gets copied to the ITCM, only the first 0x4000 bytes. I'll change this.

symbols/arm9.yml
```diff
@@ -3,14 +3,22 @@ arm9:
...
     EU: 0x2000000
     NA: 0x2000000
     JP: 0x2000000
+    EU-ITCM: 0x1FF8000
+    NA-ITCM: 0x1FF8000
   length:
     EU: 0xB7D38
     NA: 0xB73F8
     JP: 0xB8CB8
+    EU-ITCM: 0x4084
```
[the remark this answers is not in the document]
End45: Yeah, the ITCM is 0x8000 bytes long, but the arm9 section of it is only 0x4000 bytes long. Since this is what is being represented in this file, I think that's the size we should use. In the case of ram.yml, the file represents the contents of all the documented symbols in the RAM, so it makes sense to specify the RAM's size.

If we specify 0x8000, declaring a symbol at, say, 0x1FF8000+0x5000 would be considered valid since it's within the specified bounds. But I think a declaration like this should throw an error since that address is outside of the special area.

End45:
About https://github.com/UsernameFodder/pmdsky-debug/pull/55/commits/aed4eafc41a628185ba162c7531c3cab8819dcdd: Those symbols were previously defined in arm9.yml since they are placed in the area that used to contain the code moved to the ITCM at runtime. They can't be defined there since that triggers an overlap error with the ITCM region.

symbols/arm9.yml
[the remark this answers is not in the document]
UsernameFodder: You make a good point. I'm convinced. Let's keep it at 0x4000.

symbols/ram.yml
```diff
@@ -18,6 +18,41 @@ ram:
     More specifically, this file is a dumping ground for addresses that are useful to know about, but don't fall in the address ranges of any of the other files. Dynamically loaded constructs that do fall within the address range of a relevant binary should be listed in the corresponding YAML file of that binary, since it still has direct utility when reverse-engineering that particular binary.
   functions: []
   data:
+    - name: MEMORY_ALLOCATION_TABLE
```
UsernameFodder, line 21: > About [aed4eaf](https://github.com/UsernameFodder/pmdsky-debug/commit/aed4eafc41a628185ba162c7531c3cab8819dcdd): Those symbols were previously defined in arm9.yml since they are placed in the area that used to contain the code moved to the ITCM at runtime. They can't be defined there since that triggers an overlap error with the ITCM region.

Gah, that's unfortunate. I think I would prefer if we just move them into `itcm.yml` and not list addresses for the `-ITCM` versions? We can put "[runtime]" at the start of the descriptions (like we do with some other symbols) or something. The normal versions still correspond to `arm9.bin`, so it conceptually makes some sort of sense...

Keeping them under `arm9.yml` is useful when loading stuff into Ghidra so that these functions are properly labeled. If we put them in `ram.yml` then it'll be more annoying to import them into Ghidra (especially with the NTRGhidra stuff which is fully automated).

UsernameFodder (changes requested):

symbols/ram.yml
End45: I think you're getting a bit confused about these symbols. They don't exist in the binaries. Once the game is started, the following steps happen:

1. The code located at [EU]0x20B3CC0 gets moved to the ITCM (0x1FF8000)
2. The area that used to contain that code [EU]0x20B3CC0 gets filled with zeroes
3. That area is now free, so it's used to store the memory table and related symbols

So these symbols only exist in the RAM, so it makes sense to have them declared in ram.yml.

UsernameFodder: I didn't explain myself well. Your steps are correct, but even then, there's still a big advantage to putting them in `arm9.yml` rather than `ram.yml`. Other functions in `arm9.bin` (in particular those related to the memory allocator) refer to these runtime-only symbols. If they aren't listed as part of arm9, they won't be imported into Ghidra, and Ghidra won't be able to resolve those accesses into readable names, and you won't be able to create appropriate global structs at those addresses to help the decompiler.

Note the second paragraph in the description for `ram.yml`:

> More specifically, this file is a dumping ground for addresses that are useful to know about, but don't fall in the address ranges of any of the other files. Dynamically loaded constructs that do fall within the address range of a relevant binary should be listed in the corresponding YAML file of that binary, since it still has direct utility when reverse-engineering that particular binary.

That's why some other runtime-only symbols are not listed in `ram.yml`, but rather in the normal files and annotated with "[Runtime]" in the description.

End45: Oh, I see. That makes sense. I'll undo https://github.com/UsernameFodder/pmdsky-debug/commit/aed4eafc41a628185ba162c7531c3cab8819dcdd then.
But we still need a way to skip the overlap check for these symbols.

End45: That seems too hacky I think. These symbols have nothing to do with the ITCM, since they aren't part of the code that gets moved there nor are ever loaded into the ITCM themselves.

UsernameFodder: It doesn't seem all that bad to me. It has to do with the ITCM because it gets written to the section of ARM9 that used to contain ITCM stuff. That checks out given that blocks/symbols/subregions are all organized in terms of address ranges.

The only downside I can think of is clarity, since programmatically, putting a symbol in a subregion is no different from putting it directly in the main block. So I'm not sure it's worth modifying the format for a minor issue of clarity for a small collection of symbols caused by ITCM weirdness that only really happens in one place. That's why I think it would be simpler to just defer modifying the format until this actually becomes a functional problem.

End45: I never liked that philosophy because this kind of problems are usually easier to deal with when they first show up rather than when you reach a point where you're forced to properly address them, but oh well. I guess we can hope it doesn't become a larger problem or that it's easily fixable in the future if it does.

UsernameFodder: I would agree with you in some situations, but in this particular case I can't envision this becoming a larger problem that isn't just as easy to fix later. It's most likely going to be just moving around some symbols and messing with the checks somehow.

UsernameFodder (changes requested):
It looks like there's some formatting issues (I suppose the formatter doesn't like trailing newlines in the YAML files...oh well). Beyond that, I think this is pretty much ready.

UsernameFodder (approved):
Aha, I figured out how to edit it myself. I'll probably avoid doing my own edits in the future unless it's something trivial like this.

UsernameFodder:
@theCapypara (and @tech_ticks, I can't seem to ping you) pmdsky-debug-py and c-of-time will need to be updated to deal with the changes in this PR:
1. Subregions are now in use. Content-wise, they can be treated like a textual include or import (so subregion symbols can be loaded into the same structure that holds the parent file's symbols), but the subregion files will still need to be parsed separately. See [`resymgen.md`](https://github.com/UsernameFodder/pmdsky-debug/blob/0c6c4cb8488e7691f9743fabd0ebb4d680cdcbfe/docs/resymgen.md#the-resymgen-yaml-specification) for the spec. [`symbols_vfill.py`](https://github.com/UsernameFodder/pmdsky-debug/blob/0c6c4cb8488e7691f9743fabd0ebb4d680cdcbfe/tools/symbols_vfill.py#L87) and [`symdiff.py`](https://github.com/UsernameFodder/pmdsky-debug/blob/0c6c4cb8488e7691f9743fabd0ebb4d680cdcbfe/tools/symdiff.py#L804) show two different ways to do the recursive traversal in Python.
2. There are two special versions, "EU-ITCM" and "NA-ITCM", that should be understood as runtime overrides for relevant RAM addresses. I think this is mostly relevant for c-of-time. If loading addresses for runtime, and a symbol has both "x" and "x-ITCM" addresses, the "x" address should be ignored in favor of the "x-ITCM" one.

PR #60, opened 2022-06-27
Add symbol table diffing tool

### LINKED ISSUE #58: Auto-generated Changelog for Symbol Updates
**Is your feature request related to a problem? Please describe.**
At the moment when updating libraries that use pmdsky-debug such as SkyTemple Files and (to a lesser extend) c-of-time/Rust of Darkness, it's a bit of a "guessing game" / going through the history to see what symbols have changed and what might break. With SkyTemple it's especially problematic due to the lack of options for static analysis (though that's probably also that could be improved over there). SkyTemple does have some tests that check for some symbols, but they aren't really 100% reliable for this...

**Describe the solution you'd like**
It would be great if there was a changelog with each release that outlined the renamed / removed / added / updated symbols. Ideally this would *not* (just) be in the Github releases, since they are very inconvenient to get an overview of changes over multiple releases.

**Describe alternatives you've considered**
- Doing nothing; just relying on external libs to do checks themselves
- Doing this manually (not practical)

**Additional context**
/

UsernameFodder:
Resolves #58.

It can be useful to know what symbols have changed with each
commit/revision, or across a sequence of commits.

It's possible to determine this by viewing the Git history or using
`git diff`, but these don't make for very clean symbol changelogs. It's
also possible to diff different symbol packages generated by resymgen,
but the process is still quite cumbersome, and the resulting changelog
misses some useful information (like where the modified symbols are in
the original symbol tables).

The new `symdiff.py` script runs a specialized diffing algorithm to
automatically generate a high-quality symbol changelog between different
Git revisions.

### DIFF
--- a/tools/README.md
+++ b/tools/README.md
@@ -13,3 +13,6 @@ This directory contains miscellaneous tools for reverse engineering _Explorers o
 
 ## `symbols_vfill.py`
 `symbols_vfill.py` is a command line utility for filling in missing function addresses in the `pmdsky-debug` [symbol tables](../symbols), for addresses that are known in some game versions (e.g., NA, EU) but not in others. It relies on [`resymgen.py`](#resymgenpy) and thus has the same prerequisites. See the help text (`python3 symbols_vfill.py --help`) for usage instructions, and see the description in [`symbols_vfill.py`](symbols_vfill.py) itself for more details.
+
+## `symdiff.py`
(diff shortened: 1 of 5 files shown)


UsernameFodder:
@theCapypara What do you think of this script? Does the changelog it generates match what you were envisioning?

theCapypara:
Oh wow, this is really nice!
You could add this to the releases generated by the Github Actions for bonus points haha.

theCapypara (approved):

UsernameFodder:
> You could add this to the release text generated by the Github Actions for bonus points haha.

I don't think I'm confident enough right now that there are no bugs to make it part of the automation :P. Maybe in the future.

PR #61, opened 2022-07-01
Update instructions to use .sym files on No$GBA

End45:
Each ROM version now has multiple subfolders from which symbols must be copied, so I figured it would be good to include this information in the instructions on how to properly load .sym files into No$GBA.


UsernameFodder (approved):
Oh yeah, I forgot about this. Thanks!

PR #63, opened 2022-07-10
Minor changes

End45:
I've changed a couple of minor things:

- `ShouldGameOverOnImportantTeamMemberFaint` and `ShouldCauseGameOverOnFaint` have been renamed to something more generic as I've found that they are also used to check for other things.
- I've found and documented a function that checks one of the unknown fields in the team_member struct. Seems to be related to guest pokémon, but I don't know exactly how.
- I've added the second return value of the division functions, which return the reminder in r1. (Yes, they return two values, it's not a coincidence. They manually move the result into r1 right before returning, and I've seen the game get the reminder from r1 after calling one of these functions). I don't know if something needs to be changed in the header files to reflect this information.


symbols/arm9.yml
```diff
@@ -2758,7 +2769,8 @@ arm9:
         
         r0: dividend
         r1: divisor
-        return: dividend / divisor
+        return r0: dividend / divisor
+        return r1: reminder
```
UsernameFodder, line 2773: I'm aware that the division impementation behaves this way, but I don't think it's possible to represent this in standard C. Can you give examples on where the game actually uses the remainder? I'll need to look at what the calling convention actually looks like before I can decide what to do.

End45: > Can you give examples on where the game actually uses the remainder?

`GetSpriteSize` contains an instance of this. The game takes the monster ID in r0 and calls DivideInt with 600 as the divisor. Then it takes the reminder, effectively converting the monster ID to a primary gender ID.
(It also casts the result to int16_t in the process, that's why there's some shifts)

Metalcape: Another example is in `GetMonsterIdFromSpawnEntry`, where it divides the monster id in r0 by 0x483, then takes the reminder in r1 and returns it as result after casting it to 16 bit

Metalcape: By the way, I've looked at the [ARM standard for calling conventions](https://github.com/ARM-software/abi-aa/blob/60a8eb8c55e999d74dac5e368fc9d7e36e38dda4/aapcs32/aapcs32.rst#result-return), and it seems that functions can use `r1:r0` to return 64-bit values and `r3:r0` to return 128-bit values, similarly to how x86 can use `rdx:rax` for values larger than 64 bits. I don't know if this applies to this game specifically, but in theory if the developers wrote `DivideInt` as a function that returns, for example, a 64-bit struct containing remainder and quotient, the compiler could have translated it as returning that struct into `r1:r0`. So maybe doing that in the C headers may be an option?

UsernameFodder: Yeah I was actually looking at that document, too. It only appears to apply to _fundamental_ wide types, not structs. I tested it and it's true, when the return type is `unsigned long long` it returns in r0 and r1 (if it's a struct, the values are returned on the stack instead). Both GCC and Clang appear to respect this convention:

- GCC: https://godbolt.org/z/4s68hE3vM
- Clang: https://godbolt.org/z/oo6scfbGz

So, I think what we should do is:

1. Change the return types of `DivideInt`, `DivideUInt`, and `DivideUIntNoZeroCheck` to be `unsigned long long`.
2. Change the description of `DivideInt` to the following (and similarly for the other two). In diff notation:

```
  This appears to be the libgcc implementation of __divsi3 (not sure which gcc version), which implements the division operator for signed ints.

+ The return value is a 64-bit integer, with the quotient (dividend / divisor) in the lower 32 bits and the remainder (dividend % divisor) in the upper 32 bits. In accordance with the Procedure Call Standard for the Arm Architecture (see https://github.com/ARM-software/abi-aa/blob/60a8eb8c55e999d74dac5e368fc9d7e36e38dda4/aapcs32/aapcs32.rst#result-return), this means that the quotient is returned in r0 and the remainder is returned in r1.

  r0: dividend
  r1: divisor
- return: dividend / divisor
+ return: (quotient) | (remainder << 32)
```

Metalcape: > It only appears to apply to fundamental wide types

Ah so that's why it looked super ugly in Ghidra when I tried to change the return type to a struct, if it works with `unsigned long long` then I think that's a good solution

PR #65, opened 2022-07-20
Status icons

Metalcape:
Added some information on status effect icons.

I think the `f_red_exclamation_mark` bit is unused, if you know of a status effect that actually uses it let me know.

### DIFF
--- a/headers/types/dungeon_mode/dungeon.h
+++ b/headers/types/dungeon_mode/dungeon.h
@@ -909,7 +909,10 @@ struct dungeon {
     // to be set to action::ACTION_WALK until it hits an obstacle.
     bool leader_running;
     undefined field_0x78c;
-    undefined field_0x78d;
+    // 0x78D: This flag is set by the unused move 0x395 (Reviver).
+    // If true, the UpdateStatusIconFlag function will set the icon flag
+    // to display a blue exclamation mark on monsters that hold an item.
+    bool unused_reviver_item_flag;
     bool pass_turn; // 0x78E: True if the leader has chosen to pass its turn
     undefined field_0x78f;
     bool thief_alert; // 0x790: If you've stolen from Kecleon (actual dungeon state)
--- a/headers/types/dungeon_mode/dungeon.h
+++ b/headers/types/dungeon_mode/dungeon.h
@@ -909,7 +909,7 @@ struct dungeon {
     // to be set to action::ACTION_WALK until it hits an obstacle.
     bool leader_running;
     undefined field_0x78c;
-    // 0x78D: This flag is set by the unused move 0x395 (Reviver).
+    // 0x78D: This flag is set by the unused move 0x18D (Reviver).
     // If true, the UpdateStatusIconFlag function will set the icon flag
     // to display a blue exclamation mark on monsters that hold an item.
     bool unused_reviver_item_flag;
--- a/headers/types/dungeon_mode/dungeon.h
+++ b/headers/types/dungeon_mode/dungeon.h
@@ -909,7 +909,7 @@ struct dungeon {
     // to be set to action::ACTION_WALK until it hits an obstacle.
     bool leader_running;
     undefined field_0x78c;
-    // 0x78D: This flag is set by the unused move 0x18D (Reviver).
+    // 0x78D: This flag is set by the unused move 0x18B (Reviver).
     // If true, the UpdateStatusIconFlag function will set the icon flag
     // to display a blue exclamation mark on monsters that hold an item.
     bool unused_reviver_item_flag;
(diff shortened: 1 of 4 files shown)


headers/types/dungeon_mode/dungeon.h
```diff
@@ -909,7 +909,10 @@ struct dungeon {
     // to be set to action::ACTION_WALK until it hits an obstacle.
     bool leader_running;
     undefined field_0x78c;
-    undefined field_0x78d;
+    // 0x78D: This flag is set by the unused move 0x395 (Reviver).
```
UsernameFodder, line 912: Hmm, does this mean that `enum move_id` should go higher than it does currently? I can't remember where I pulled the maximum value from, but I definitely didn't go as high as 0x395.

headers/types/dungeon_mode/dungeon_mode.h
```diff
@@ -240,6 +240,59 @@ struct statuses {
...
+    bool f_red_exclamation_mark : 1; // Probably unused
+    bool f_magnet_rise : 1;          // Purple arrow pointing up
+
+    // The following 4 bytes appear to have a different meaning, maybe they are intended to
+    // represent icons that are always displayed and do not cycle (inferred from the difference
+    // between the ice block and the other icons). Except for the first bit, the others do not
+    // seem to have an effect, but the code stores the full 4 bytes as a bitwise OR of some of the
+    // flags (see UpdateStatusIconBitfield).
+    bool f_freeze : 1; // Ice block
+    uint16_t flags_unk2 : 7;
```
UsernameFodder, line 288: Make this `uint8_t`. It's not big enough to need to be `uint16_t`.

headers/types/dungeon_mode/dungeon_mode.h
```diff
@@ -240,6 +240,59 @@ struct statuses {
 ASSERT_SIZE(struct statuses, 118);
 #pragma pack(pop)
 
+#pragma pack(push, 1)
```
UsernameFodder, line 243: I don't think you need packing here? The struct layout doesn't have any implicit padding.

headers/types/dungeon_mode/dungeon_mode.h
```diff
@@ -240,6 +240,59 @@ struct statuses {
...
+    bool f_cross_eyed : 1;           // Blue question mark
+    bool f_eyedrops : 1;             // Blue eye blinking yellow with a circular wave
+    bool f_muzzled : 1;              // Blinking red cross
+    bool f_grudge : 1;               // Purple shield
+    bool f_exposed : 1;              // Blue eye blinking red with a circular wave
+    bool f_sleep : 1;                // Red Z's
+    bool f_lowered_stat : 1;         // Yellow arrow pointing down
+    bool f_heal_block : 1;           // Blinking green cross
+    bool f_miracle_eye : 1;          // Blinking orange cross
+    bool f_red_exclamation_mark : 1; // Probably unused
```
UsernameFodder, line 279: No idea 😆. It doesn't seem to correspond to any normal status condition.

headers/types/dungeon_mode/dungeon_mode.h
```diff
@@ -554,14 +607,7 @@ struct monster {
...
     undefined field_0x217;
-    undefined field_0x218;
-    undefined field_0x219;
-    undefined field_0x21a;
-    undefined field_0x21b;
-    undefined field_0x21c;
-    undefined field_0x21d;
-    undefined field_0x21e;
-    undefined field_0x21f;
+    struct status_icon_flags status_icons;
```
UsernameFodder, line 610: Can we comment this with the offset like all the other fields?

UsernameFodder (changes requested):

headers/types/dungeon_mode/dungeon.h
Metalcape: Sorry, my bad, that's decimal lol, hexadecimal ID is 0x18B

headers/types/dungeon_mode/dungeon_mode.h
Metalcape: I think we can leave it like this then?

headers/types/dungeon_mode/dungeon.h
Metalcape: Actually, I've taken a closer look, and this is not set by move 0x18B but by move 0x189, which is the Identify Orb. The flag is not unused, because the Identify Orb signals to the player any monster that holds an item and is not in the team, and this is done by placing an icon on them. This is such an obscure mechanich that I had no idea it even existed.

headers/types/dungeon_mode/dungeon_mode.h
UsernameFodder: Yeah it's fine the way it is.

UsernameFodder (approved):
Looks good, thanks!

PR #67, opened 2022-08-08
Add a couple of functions

End45:
Renames field 0x3 in the dungeon mode struct and adds its getter and setter functions.
Also adds the function responsible for loading the sprites of monsters from missions.


UsernameFodder (approved):
Looks good, thanks!

PR #69, opened 2022-09-14
Add thrown item entity struct

End45:
Just a minor recent discovery in the dungeon struct


UsernameFodder (approved):
Woohoo, net -182 lines! Love to see it.

PR #74, opened 2022-09-21
Dungeon struct research

End45:
I've researched and documented all the big fields in the dungeon struct, which brings down the length of `dungeon.h` down to ~2100 lines.

You might want to get rid of `#ifndef OMIT_LARGE_FILES` and the paragraph at https://github.com/UsernameFodder/pmdsky-debug/blob/master/docs/using-debug-info.md that starts with

> The Ghidra parser is slow, and parsing can easily take over an hour (mainly because the dungeon.h header file is so large) [...]

There's also a commit that adds a couple of return values that were missing from `overlay29.yml`.


headers/types/dungeon_mode/dungeon_mode.h
```diff
@@ -864,6 +864,21 @@ struct tile {
 };
 ASSERT_SIZE(struct tile, 20);
 
+// Information about the rooms on the current floor
+struct room_data {
+    uint8_t room_id;
+    undefined field_0x1;                 // Initialized to 0
+    struct position bottom_right_corner; // 0x2
+    struct position top_left_corner;     // 0x6
+    undefined field_0xA;                 // Doesn't get initialized, likely padding
```
UsernameFodder: Minor nitpick, but the `field_0x...` is a Ghidra convention, and it uses lowercase 0xa-0xf. I'll go ahead and make the tweak.

UsernameFodder (approved):
Wow, this is unexpected, but amazing! It'll be so much easier to work with `dungeon.h` now! I went ahead and removed `OMIT_LARGE_FILES` as you suggested, along with some other minor fixes.

PR #76, opened 2022-09-30
Info dump 14-17/64: Missions, floors and more

End45:
Adds functions related to the following:

- Missions
- Leader changing
- Dungeon floors
- Personality quiz

I've also documented the struct used to hold mission data and most of its fields.


End45:
Whoops, forgot to fill in the data for one function. Hold on...

headers/types/common/common.h
```diff
@@ -743,6 +743,34 @@ struct rankup_table_entry {
...
+    struct mission_status_8 status; // 0x0
+    struct mission_type_8 type;     // 0x1
+    union mission_subtype subtype;  // 0x2
+    undefined field_0x3;
+    struct dungeon_id_8 dungeon_id; // 0x4
+    uint8_t floor;                  // 0x5
+    undefined field_0x6;            // Likely padding
+    undefined field_0x7;            // Likely padding
+    int field_0x8;                  // 0x8, changing it seems to affect the text of the mission
+    undefined field_0xC;
```
UsernameFodder, line 757: I think we've gotten inconsistent at this point, but the naming convention here originally came from Ghidra's automatically generated names, and used lowercase letters. At some point we'll probably want to make everything consistent again, but for net new code I say we should stick with Ghidra's convention for field names.

symbols/arm9.yml
```diff
@@ -2202,6 +2295,22 @@ arm9:
         r0: monster ID
         r1: ?
         return: bool
+    - name: GetPlayerData
+      address:
+        EU: 0x2055AEC
+        NA: 0x2055770
+      description: |-
+        Returns the data of the player pokémon (first slot in Chimeco Assembly)
```
UsernameFodder, line 2303: Chimecho

symbols/arm9.yml
```diff
@@ -2202,6 +2295,22 @@ arm9:
         r0: monster ID
         r1: ?
         return: bool
+    - name: GetPlayerData
```
UsernameFodder, line 2298: Is this strictly the main character, or whoever happens to be leader? If the former, I think we usually use the term "hero" rather than "player".

symbols/arm9.yml
```diff
@@ -2313,6 +2440,89 @@ arm9:
...
+        
+        r0: Monster ID
+        r1: True to exclude monsters in the MISSION_BANNED_MONSTERS array, false to allow them
+        return: True if the specified monster can be part of a mission
+    - name: IsMonsterMissionAllowedStory
+      address:
+        EU: 0x2062E60
+        NA: 0x2062AE4
+      description: |-
+        Checkss if the specified monster should be allowed to be part of a mission (probably as the client or the target), accounting for the progress on the story.
```
UsernameFodder, line 2517: Checks

UsernameFodder (changes requested):
As usual, will make changes on the unambiguous stuff.

headers/types/common/common.h
End45: Right, I keep fogetting this. I'm not used to typing hex numbers in lowecase letters, it looks wrong to me.

symbols/arm9.yml
End45: I was debating between "hero", "player" and "protagonist".  I settled on "player" because that's the one SkyTemple uses, but I'll change it since it seems like this repo already uses "hero".

headers/types/common/common.h
UsernameFodder: Yeah I don't really like it either and think uppercase is more readable, which is why I used uppercase everywhere else. But I think it's good to adhere to Ghidra's standard where one exists.

UsernameFodder (approved):

Issue #81, opened 2022-11-03
Port symbols from Irdkwia's ASM.ods


UsernameFodder:
The irdkwia_doc archive has a huge ASM.ods spreadsheet with tons of symbols that could be ported to `pmdsky-debug`. It'd probably be too much effort to stringently verify all of them, but it would still be useful to semi-blindly port them over as is, accompanied by notes in the description fields about their unverified status.

## Sheets

- [ ] ARM9
- [x] overlay_0001
- [x] overlay_0010
- [x] overlay_0011
- [x] overlay_0013
- [x] overlay_0014
- [x] overlay_0015
- [x] overlay_0016
- [x] overlay_0017
- [x] overlay_0018
- [x] overlay_0019
- [x] overlay_0020
- [x] overlay_0021
- [x] overlay_0022
- [x] overlay_0023
- [x] overlay_0024
- [x] overlay_0025
- [x] overlay_0026
- [x] overlay_0027
- [x] overlay_0029
- [x] overlay_0030
- [x] overlay_0031
- [x] overlay_0034

UsernameFodder:
overlay_0014 only has one symbol, which is already in `pmdsky-debug` (`FOOTPRINT_DEBUG_MENU`).

PR #93, opened 2022-11-12
Port overlay_0029 from Irdkwia's ASM.ods

UsernameFodder:
The dungeon mode overlay. Ported stuff from the spreadsheet with assorted tweaks:
- Corrected some mistakes
- Did further investigation
- Renamed various symbols to match style conventions
- Consolidated various addresses into existing symbols

This is the last remaining overlay. The only thing left is ARM9.

Parent issue: #81

### DIFF
--- a/symbols/overlay29.yml
+++ b/symbols/overlay29.yml
@@ -173,6 +173,11 @@ overlay29:
           - 0x234987C
           - 0x234DDD0
           - 0x234EC14
+        JP:
+          - 0x2312538
+          - 0x2312DCC
+          - 0x23165F0
+          - 0x23364C4
       description: |-
         Checks if an entity pointer points to a valid entity (not entity type 0, which represents no entity).
         
@@ -202,6 +207,25 @@ overlay29:
         
         r0: if true, the function will not check for the end of the floor condition and will skip other (unknown) actions in case of forced loss.
         return: true if the forced loss happens, false otherwise
+    - name: IsBossFight
+      address:
+        EU: 0x22E11A4
+        NA: 0x22E0864
+        JP: 0x22E1EF4
+      description: |-
+        Note: unverified, ported from Irdkwia's notes
+        
+        r0: fixed_room_id
+        return: bool
+    - name: IsCurrentFixedRoomBossFight
+      address:
+        EU: 0x22E11C0
+        NA: 0x22E0880
+        JP: 0x22E1F10
+      description: |-
+        Note: unverified, ported from Irdkwia's notes
+        
+        return: bool
     - name: FixedRoomIsSubstituteRoom
       address:
         EU: 0x22E120C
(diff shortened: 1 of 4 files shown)


symbols/overlay29.yml
```diff
@@ -771,6 +821,17 @@ overlay29:
         r2: y position
         r3: flags
         stack[0]: visibility flag
+    - name: PerformLeaderAction
+      address:
+        EU: 0x22F1890
+        NA: 0x22F0EDC
+        JP: 0x22F24DC
+      description: |-
+        This appears to be the function that actually performs the leader's action within RunLeaderTurn.
```
End45, line 830: This is either incorrect or incomplete, this function sets the leader's action field depending on the current situation (for example, depending on which buttons are currently pressed). It returns once the leader has an action assigned, so said action is probably executed later on.

UsernameFodder: Hmm are you sure? I admittedly didn't look too much at this function (it was just call "LeaderTurn" in irdkwia's spreadsheet), but it's quite a complex function. There's a big while loop in it that starts with `GetLeader` and then does a bunch of stuff.

On top of that, the way it's called within `RunLeaderTurn` seems pretty suggestive, although I'm sort of confused how it fits in with `ExecuteMonsterAction`:
```c
// excerpt
...
    bVar2 = IsFloorOver();
    if (bVar2 != '\0') {
      return '\0';
    }
    FUN_022e81f8();
    EnemyEvolution(peVar3);
    (*PTR_DUNGEON_PTR_022ec600)->no_action_in_progress = '\x01';
    LeaderTurn();
    (*PTR_DUNGEON_PTR_022ec600)->no_action_in_progress = '\0';
    bVar2 = IsFloorOver();
    if (bVar2 != '\0') {
      return '\x01';
    }
    ExecuteMonsterAction(peVar3);
    FUN_022ef9c8();
    TryForcedLoss('\0');
    bVar2 = IsFloorOver();
    if (bVar2 != '\0') {
      return '\x01';
    }
    puVar5 = (ushort *)peVar3->info;
    if ((*puVar5 & 0x8000) != 0) {
      *puVar5 = *puVar5 & 0x7fff;
      *puVar5 = *puVar5 | 0x4000;
    }
    bVar2 = IsFloorOver();
    ppdVar1 = PTR_DUNGEON_PTR_022ec600;
    if (bVar2 != '\0') {
      return '\x01';
    }
...
```

If what you say is true, I suppose that means `dungeon::no_action_in_progress` is probably wrong as well?

UsernameFodder: Oh lol I just realized I was mixing up what **no**_action_in_progress means. I guess your explanation does seem to be in line with what the code above is doing.

I was just guessing while migrating. Could you put up a PR to fix the description? I don't actually have a good idea what this function does.

End45: Sure, I'll see if I can come up with a better name and description. The function does multiple things (and I haven't even fully researched it).

PR #98, opened 2022-11-14
Update PerformLeaderAction and document Anger Point flag

End45:
Hopefully this explanation for `SetLeaderAction` is clearer.


UsernameFodder (approved):
Thanks!

PR #109, opened 2022-12-19
Use GITHUB_OUTPUT env file for release actions

UsernameFodder:
The set-output command is being deprecated in favor of the newer $GITHUB_OUTPUT environment file, so switch over to using that in release.yml.

https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/


UsernameFodder:
There is no good way to test this before merging... 🤞

PR #110, opened 2023-01-05
Info dump 24-26/66: PP, IQ and recruiting

End45:

- Using and recovering PP
- IQ skills
- Enemy recruiting in dungeons

Other relevant changes include:

- The `faint_reason` enum has been renamed to `damage_source`, since I discovered that values of this enum are passed to `CalcDamageFixed`. This function checks if the damage source is equal to `DAMAGE_SOURCE_THROWN_ITEM` and increases the damage if the user has the Power Pitcher IQ skill, so it's more than just a faint reason (although many of the values are used as faint reasons)
- The old `damage_source` enum has been renamed to `damage_message`, since it looks like it's only used to print damage-related messages to the message log.


headers/functions/arm9.h
```diff
@@ -92,7 +92,7 @@ void AllocAndLoadFileInPack(enum pack_file_id pack_id, uint32_t file_index,
                             struct pack_alloc_and_load_result* output, uint32_t malloc_flags);
 void OpenPackFile(struct pack_file_opened* pack_file, const char* file_name);
 uint32_t GetFileLengthInPack(struct pack_file_opened* pack_file, uint32_t file_index);
-union faint_reason GetFaintReason(enum move_id, enum item_id);
+union damage_source GetFaintReason(enum move_id, enum item_id);
```
UsernameFodder, line 95: Should the function be renamed as well?

headers/functions/overlay29.h
```diff
@@ -293,7 +308,7 @@ bool DungeonRandOutcomeUserAction(struct entity* user, int percentage);
 bool CanAiUseMove(struct entity* monster, int move_index, bool extra_checks);
 bool CanMonsterUseMove(struct entity* monster, struct move* move, bool extra_checks);
 void UpdateMovePp(struct entity* entity, bool can_consume_pp);
-union faint_reason GetFaintReasonWrapper(struct move* move, enum item_id item_id);
+union damage_source GetFaintReasonWrapper(struct move* move, enum item_id item_id);
```
UsernameFodder, line 311: This would also be renamed if the other comment applies.

symbols/arm9.yml
```diff
@@ -6231,6 +6254,59 @@ arm9:
...
+        Enables an IQ skill and disables any other skills that are incompatible with it.
+        
+        r0: Pointer to the bitarray containing the list of enabled IQ skills
+        r1: ID of the skill to enable
+    - name: GetSpeciesIqSkill
+      address:
+        EU: 0x20591E4
+        NA: 0x2058E68
+      description: |-
+        Gets the <index>th skill on the list of IQ skills that a given monster species can learn.
```
UsernameFodder, line 6305: Is this zero-indexed? I assume so, but probably best to specify.

symbols/overlay10.yml
```diff
@@ -443,7 +443,15 @@ overlay10:
       length:
         EU: 0x2
         NA: 0x2
-      description: The percentage increase in experience from exp-boosting exclusive items
+      description: The percentage increase in experience from exp-boosting exclusive items.
+    - name: AFTERMATH_CHANCE
```
UsernameFodder, line 447: Needs an addition to `data/overlay10.h`

symbols/overlay29.yml
```diff
@@ -883,6 +926,16 @@ overlay29:
...
+    - name: DebugRecruitingEnabled
+      address:
+        EU: 0x22F0420
+        NA: 0x22EFA6C
+      description: |-
+        Always returns true. Called by SpecificRecruitCheck.
+        
+        Seems to be a function used during development to disable recruiting. If it returns false, SpecificRecruitCheck will also return false.
+        
+        No params.
```
UsernameFodder, line 938: Returns a boolean, not no params?

symbols/overlay29.yml
```diff
@@ -1634,6 +1687,16 @@ overlay29:
         r0: entity pointer
         r1: IQ skill ID
         return: bool
+    - name: UpdateIqSkills
```
UsernameFodder, line 1690: Was the function signature omitted on purpose for this one? If the params aren't known that's fine, just want to make sure.

UsernameFodder (changes requested):

headers/functions/arm9.h
End45: Fixed!
I also updated a few more things that I forgot to rename

symbols/arm9.yml
End45: :+1:

symbols/overlay29.yml
End45: Oh, looks like I forgot to fill that in. Shouldn't `symbol_check` catch that I didn't add this function to the headers? Because when I run it, it prints nothing.

UsernameFodder: `symbol_check` only enforces things in one direction (everything in the headers must be in the symbols), since it's fine to have symbols that you don't know the exact type for.

I added `symbol_check --extra` a while back to warn you about things in the other direction, which you can run manually, but it's a little noisy. It doesn't make sense to put it in the automation though.

UsernameFodder (approved):
Fixed a minor capitalization issue. Looks good otherwise, thanks!

PR #115, opened 2023-01-30
Info dump 28/66: Moves and effects

End45:
Adds some symbols related to moves, move effects and item effects.
I also added the EU offset of some `DoMove___` functions I stumbled upon, but there's still a lot of EU offsets missing for those.

### DIFF
--- a/symbols/overlay29.yml
+++ b/symbols/overlay29.yml
@@ -105,6 +105,7 @@ overlay29:
           - 0x2320664
           - 0x2320BE4
           - 0x23211CC
+          - 0x2321EA0
           - 0x2326088
           - 0x232F280
           - 0x23349EC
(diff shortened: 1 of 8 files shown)


symbols/overlay29.yml
```diff
@@ -301,6 +302,27 @@ overlay29:
         
         r0: position
         return: entity pointer for the newly added item, or null on failure
+    - name: ShouldDisplayEntityMessages
+      address:
+        EU: 0x22E2EB4
+        NA: 0x22E2574
+      description: |-
+        Checks if messages that involve a certain entity should be displayed or supressed.
```
UsernameFodder, line 310: \*suppressed

symbols/overlay29.yml
```diff
@@ -1488,6 +1522,19 @@ overlay29:
...
         return: True if the target has at least one super effective move against the user, false otherwise.
+    - name: TryEatItem
+      address:
+        EU: 0x22FBC20
+        NA: 0x22FB214
+      description: |-
+        The user attempts to eat an item from the target.
+        
+        The function tries to eat the target's held item first. If that's not possible and the target is part of the team, it attempts to eat a random edible item from the bag instead.
+        Fun fact: The code used to select the random bag item that will be eaten is poorly coded. As a result, there's a small chance of the first edible item in the bag being picked instead of a random one. The exact chance of this happening is (N/B)^B, where N is the amount of non-edible items in the bag and B is the total amount of items in the bag.
```
UsernameFodder, line 1533: So without looking at the code I'm guessing it just picks a random item, checks if it's edible, and rerolls if not up to _B_ times?

symbols/overlay29.yml
UsernameFodder: Fixed

UsernameFodder (approved):
Looks good, thanks!

symbols/overlay29.yml
End45: Yeah, pretty much. It actually builds an array with `B` random entries from the bag (which can be repeated) and then appends the numbers 0 to `B - 1` to it. Then the game iterates said array until one of the entries in it corresponds to an edible item.
No idea what they were thinking there lol.

PR #117, opened 2023-02-12
Document random movement chance

End45:
Adds details for one of the previously unknown dungeon restrictions. Thanks to @AnonymousRandomPerson for pointing out the fact that this field is the same as a flag used in RRT and providing an explanation on how it works.

### DIFF
--- a/headers/types/dungeon_mode/dungeon_mode.h
+++ b/headers/types/dungeon_mode/dungeon_mode.h
@@ -1332,7 +1332,7 @@ struct dungeon_restriction {
     uint8_t max_party_size;    // 0x6: Maximum number of team members that can enter the dungeon
     undefined field_0x7;
     uint16_t turn_limit_per_floor; // 0x8: Number of turns per floor before the wind blows you out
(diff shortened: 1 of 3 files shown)


headers/types/dungeon_mode/dungeon_mode.h
```diff
@@ -1333,7 +1332,8 @@ struct dungeon_restriction {
     uint8_t max_party_size;    // 0x6: Maximum number of team members that can enter the dungeon
     undefined field_0x7;
     uint16_t turn_limit_per_floor; // 0x8: Number of turns per floor before the wind blows you out
-    undefined field_0xa;
+    // 0x9: Chance of setting the monster::random_movement field to 1 when spawning an enemy
```
AnonymousRandomPerson, line 1335: Should this be 0xA rather than 0x9?

End45: Right, thanks

UsernameFodder (approved):
LGTM

PR #118, opened 2023-02-12
Fix size of the random movement chance restriction

End45:
Just checked the ASM again, it's actually a signed 2-byte value.


UsernameFodder (approved):
Lol I should've waited longer on the last PR

PR #127, opened 2023-04-02
Added information for dungeon fainting, statuses, post-move attacker/defender interactions, traps, etc.

Jawshoeuh:
Added dungeon fainting, statuses, and item/ability post-move attacker-defender interactions, trap information, two functions for item effects, and one function for move animations.


headers/types/dungeon_mode/dungeon.h
```diff
@@ -1176,7 +1180,7 @@ struct dungeon {
     undefined field_0x3e19;
     undefined field_0x3e1a;
     undefined field_0x3e1b;
-    undefined4 field_0x3e1c;
+    undefined field_0x3e1c;
```
End45, line 1183: Why was this changed? It's going to break the size of the whole struct.

headers/types/dungeon_mode/dungeon.h
```diff
@@ -1753,6 +1759,8 @@ struct dungeon {
     undefined field_0x2ca27;
     undefined field_0x2ca28;
     undefined field_0x2ca29;
+    // 0x2CA0C: Probably a char[] that holds the name for the monster that fainted.
+    // Uncertain of exact size.
```
End45, line 1763: If it holds the name of a monster, then it's probably 10 characters long, but if we don't know for sure, I guess it's better to leave it like this.

headers/types/dungeon_mode/dungeon_mode.h
```diff
@@ -101,7 +101,11 @@ struct statuses {
     uint8_t burn_turns;  // 0x17: Turns left for the status in statuses::burn
     // 0x18: Turns left until residual damage for the status in statuses::burn, if applicable
     uint8_t burn_damage_countdown;
-    undefined field_0x19;
+    // 0x19: The number of times the statuses::burn_damage_countdown has reached 0. Only used
+    // when badly poisoned. Determines how much damage the badly poisoned status condition
+    // will deal. There is no noticable difference because the table this value is looked up
+    // on is filled with 0x6
+    uint8_t burn_damaged_count;
```
End45, line 108: If it's used for badly poisoned pokémon, then why is it called `burn_damaged_count`?

headers/types/dungeon_mode/dungeon_mode.h
```diff
@@ -458,14 +465,15 @@ struct monster {
     undefined field_0x189;
     undefined field_0x18a;
     undefined field_0x18b;
-    undefined field_0x18c;
-    undefined field_0x18d;
-    undefined field_0x18e;
-    undefined field_0x18f;
+    // 0x18C: Bitflags to that cause non-damaging items to trigger on the attacker
+    // after they have completed their move. (Only uses first 21 bits).
```
End45, line 469: This description isn't very clear. Maybe you could add an example?
Also, is "to that" a typo?

headers/types/dungeon_mode/dungeon_mode.h
```diff
@@ -458,14 +465,15 @@ struct monster {
...
+    uint32_t item_trigger_bitflags;
     undefined field_0x190;
     undefined field_0x191;
-    undefined field_0x192;
-    undefined field_0x193;
+    // 0x192: Bitflags to that cause non-damaging abilities to trigger on the attacker
+    // after they have completed their move. (Only uses first 11 bits). One exception
+    // is the move Rapid Spin which sets one of the flags for the user.
```
End45, line 475: Same as above.

symbols/overlay29.yml
```diff
@@ -1085,6 +1085,40 @@ overlay29:
         r2: y position
         r3: flags
         stack[0]: visibility flag
+    - name: PrepareTrapperTrap
+      address:
+        NA: 0x22EDBD4
+      description: |-
+        Saves the relevant information in the dungeon struct to later place a trap at the
+        location of the entity. (Only called with trap ID 0x19, but could be used with others).
```
End45, line 1093: > Only called with trap ID 0x19

Even when a trapper orb is used?

Jawshoeuh: This function is only called by DoMoveTrapperOrb and it only uses 0x19. The function that creates traps seems to randomly select a trap ID that's not the wonder tile trap if it's above 0x18.

End45: Oh, so it uses 0x19 as a way to make the function pick a random trap. I see.

headers/types/dungeon_mode/dungeon.h
Jawshoeuh: I thought I made a mistake typing because it's directly above one of my comments. Will correct.

headers/types/dungeon_mode/dungeon.h
Jawshoeuh: It calls GetMonsterName with this location as the buffer, but it seemed longer than 10 characters and that function still says unverified and imported from Irdkwia's notes. So, I was wary to say definitively. If you would like to see for yourself it's done in 0x022ECDE4

symbols/overlay29.yml
UsernameFodder: Trap ID 0x19 is `TRAP_NONE`. Can we use that terminology here instead? And update the comment over `TRAP_NONE` with this additional use?

symbols/overlay29.yml
```diff
@@ -1085,6 +1085,40 @@ overlay29:
...
+        
+        r0: entity pointer
+        r1: trap ID
+        r2: team (see struct trap::team)
+    - name: TrySpawnTrap
+      address:
+        NA: 0x22EDCBC
+      description: |-
+        Checks if the a trap can be placed on the tile. If the trap ID is above 0x18 (the
```
UsernameFodder, line 1102: Can we say "If the trap ID is >= TRAP_NONE"?

symbols/overlay29.yml
```diff
@@ -1085,6 +1085,40 @@ overlay29:
...
+        last value for a trap), randomly select another trap (except for wonder tile). After
+        30 failed attempts to select a non-wonder tile trap ID, default to chestnut trap.
+        If the checks pass, spawn the trap.
+        
+        r0: position
```
UsernameFodder, line 1107: This parameter doesn't line up with the function signature. Can we fix whichever one is wrong?

headers/functions/overlay29.h
```diff
@@ -97,6 +97,9 @@ enum forced_loss_reason GetForcedLossReason(void);
 void BindTrapToTile(struct tile* tile, struct entity* trap, bool is_visible);
 void SpawnEnemyTrapAtPos(enum trap_id trap_id, int16_t x, int16_t y, uint8_t flags,
                          bool is_visible);
+void PrepareTrapperTrap(struct entity* entity, enum trap_id trap_id, uint8_t team);
+void TrySpawnTrap(struct entity* entity, enum trap_id trap_id, uint8_t team, bool visible);
```
UsernameFodder, line 101: The description says this returns a bool?

headers/functions/overlay29.h
```diff
@@ -97,6 +97,9 @@ enum forced_loss_reason GetForcedLossReason(void);
...
+void TrySpawnTrapperTrap(struct entity* entity);
```
UsernameFodder, line 102: The description says this returns a bool?

symbols/overlay29.yml
```diff
@@ -1326,6 +1360,16 @@ overlay29:
         r0: target entity pointer
         r1: current_id
         return: ?
+    - name: TryActivateTraceAndColorChange
```
UsernameFodder, line 1363: The function signature says this returns a bool?

symbols/overlay29.yml
```diff
@@ -1380,6 +1424,16 @@ overlay29:
         
         r0: entity pointer
         return: bool
+    - name: TryActivateConversion2
```
UsernameFodder, line 1427: The function signature says this returns a bool?

headers/functions/overlay29.h
```diff
@@ -360,6 +367,8 @@ void ViolentSeedBoost(struct entity* attacker, struct entity* defender);
 void ApplyGummiBoostsDungeonMode(struct entity* user, struct entity* target,
                                  enum type_id gummi_type, int random_stat_boost);
 bool CanMonsterUseItem(struct entity* entity, struct item* item);
+void ApplyGrimyFoodEffect(struct entity* attacker, struct entity* defender);
```
UsernameFodder, line 370: The parameter names should be `user` and `target`?

headers/functions/overlay29.h
```diff
@@ -360,6 +367,8 @@ void ViolentSeedBoost(struct entity* attacker, struct entity* defender);
...
+void ApplyMixElixirEffect(struct entity* attacker, struct entity* defender);
```
UsernameFodder, line 371: The parameter names should be `user` and `target`?

symbols/overlay29.yml
```diff
@@ -3927,6 +4040,17 @@ overlay29:
         Activates the Mud Sport or Water Sport condition on the dungeon floor for some number of turns.
         
         r0: water sport flag (false for Mud Sport, true for Water Sport)
+    - name: TryActivateWeather
```
UsernameFodder, line 4043: This is missing a function signature. Do we know enough to specify one in the headers?

symbols/overlay29.yml
```diff
@@ -6561,6 +6685,24 @@ overlay29:
         Note that this array is only used for the cardinal directions. The elements at odd indexes are unused and unconditionally set to 0.
         
         This array is used by the dungeon generation algorithm when generating room imperfections. See GenerateRoomImperfections.
+    - name: GUMMI_LIKE_MESSAGE_IDS
```
UsernameFodder, line 6688: Assuming this means what I think it means, can we use the term `STRING_IDS` for all three of these symbols rather than `MESSAGE_IDS`?

headers/types/dungeon_mode/dungeon.h
```diff
@@ -1096,6 +1097,9 @@ struct dungeon {
     undefined field_0x3b73;
     // 0x3B74: Unknown array, likely one entry per monster species
     uint8_t unknown_array_0x3B74[600];
+    // 0x3DCC: The following entries are 4 bytes long and probably an array. I'm not certain
```
UsernameFodder, line 1100: Rather than saying what this probably is in a comment, can we just change the struct, and specify in the comment that we aren't totally sure?

```c
undefined4 field_0x3dcc[20];
```
```

File: headers/types/dungeon_mode/dungeon.h [hunk 18]
```diff
@@ -1723,6 +1727,8 @@ struct dungeon {
     uint16_t spawn_table_entries_chosen[16];
     undefined field_0x2ca0a;
     undefined field_0x2ca0b;
+    // 0x2CA0C: Probably a char[] that holds the name for the monster that caused
```
UsernameFodder, line 1730: Can we make this a `char[10]` and say that the exact size is a guess?

headers/types/dungeon_mode/dungeon.h
UsernameFodder: I would rather just make this a `char[10]` and specify that the exact size is a guess. It's more descriptive that way.

headers/types/dungeon_mode/dungeon.h
```diff
@@ -1783,6 +1791,9 @@ struct dungeon {
     undefined field_0x2ca45;
     undefined field_0x2ca46;
     undefined field_0x2ca47;
+    // 0x2CA0C: Probably a char[] because strcpy is called on it; however, unable to determine
```
UsernameFodder, line 1794: Since we know that it's a char array but don't know the length, can we make this a `char[1]`?

headers/types/dungeon_mode/dungeon.h
```diff
@@ -1815,25 +1826,19 @@ struct dungeon {
...
-    undefined field_0x2ca79;
-    undefined field_0x2ca7a;
+    struct dungeon_id_8 fainted_id; // 0x2CA68: Copied from dungeon::id, upon fainting.
+    uint8_t fainted_floor;          // 0x2CA69: Copied from dungeon::floor, upon fainting.
+    // 0x2CA6A: Copy of the fainted monster's held item.
+    struct item fainted_monster_held_item;
+    int fainted_monster_exp; // 0x2CA70: Copy of fainted monster's experience.
+    // 0x2CA74: Copy of the fainted monster's ACTUAL max hp (monster::max_hp_stat +
+    // monster::max_hp_boost)
+    int16_t fainted_monster_actual_max_hp;
```
UsernameFodder, line 1836: Can we just call this `fainted_monster_max_hp`? No qualifier should imply "actual".

headers/types/dungeon_mode/dungeon_mode.h
```diff
@@ -138,11 +142,12 @@ struct statuses {
...
     undefined field_0x39;
     undefined field_0x3a;
-    undefined field_0x3b;
-    undefined field_0x3c;
-    undefined field_0x3d;
-    undefined field_0x3e;
-    undefined field_0x3f;
+    // 0x3B: Leech seed related tracker. Set to the value in statuses:0x7 after finding
+    // statuses::leech_seed_source_monster_index.
+    uint32_t field_0x3b;
```
UsernameFodder, line 147: Can we name this something with `leech_seed`? Like `unk_leech_seed_tracker` or something. If we have a general sense of what this is for, I would rather have a descriptive name because you don't really see comments in Ghidra's decompiler.

headers/types/dungeon_mode/dungeon_mode.h
```diff
@@ -436,11 +441,13 @@ struct monster {
     undefined field_0x16d;
     undefined field_0x16e;
     undefined field_0x16f;
-    undefined field_0x170;
+    bool field_0x170;
     undefined field_0x171;
-    undefined field_0x172;
+    // 0x172: Set when the leader and falling through a pitfall trap.
+    bool field_0x172;
```
UsernameFodder, line 447: Can we call this `pitfall_trap_flag_0x172`?

headers/types/dungeon_mode/dungeon_mode.h
```diff
@@ -436,11 +441,13 @@ struct monster {
...
     undefined field_0x173;
-    undefined field_0x174;
+    // 0x174: Set when the leader and falling through a pitfall trap.
+    bool field_0x174;
```
UsernameFodder, line 450: Can we call this `pitfall_trap_flag_0x174`?

UsernameFodder (changes requested):

symbols/overlay29.yml
Jawshoeuh: I am not entirely confident on what the booleans passed in are used for. So I didn't know what to call them parameters in the header. Should I still add it based upon my guesses?

UsernameFodder: I usually add a signature as long as I've confirmed _how many_ parameters there are, and I know what the return type is. If I don't know the rest, I just copy it from Ghidra. So the minimal signature would be something like

```c
<return type> MyFunction(undefined param_1, ..., undefined param_n);
```

The rest can be filled in as and when they're figured out.

UsernameFodder (approved):
Looks good to me, thanks! BTW, feel free to resolve comments yourself after implementing suggestions. It's a good way to keep track of what's been done.

PR #130, opened 2023-04-02
Added some tiny info about picking up items

Adex-8x:
Added two functions that handle what should happen when an entity steps on an item.


UsernameFodder (approved):
Thanks! Glad you decided to contribute.

PR #133, opened 2023-04-17
Added info about the wan table

marius851000:
This a structure that store the loaded Wan sprites

### DIFF
--- a/headers/types/common/common.h
+++ b/headers/types/common/common.h
@@ -892,40 +892,42 @@ ASSERT_SIZE(struct version_exclusive_monster, 4);
 
 // An entry correspond to sprite loaded in memory and ready to be displayed
 struct wan_table_entry {
-    char path[0x20];                // Need to be null-terminated. Only used for direct file.
-    bool file_externally_allocated; // True if the iov_base shouldn’t be freed by this struct.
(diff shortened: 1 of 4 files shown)


headers/types/common/common.h
```diff
@@ -6,6 +6,7 @@
 #include "enums.h"
 #include "../dungeon_mode/dungeon_mode_common.h"
 #include "file_io.h"
+#include "../files/wan.h"
```
UsernameFodder, line 9: Did you mean to include `../files/files.h`? Otherwise `files.h` is unused and I think there's no point in creating it.

UsernameFodder (commented):
Please note the style changes I made in https://github.com/UsernameFodder/pmdsky-debug/pull/133/commits/b85e7927c4e506b1dd06b70e1e4ccadac7e5782a for future reference:

- Use `int16_t` rather than `short`
- Use decimal numbers when defining array and struct lengths

headers/types/common/common.h
marius851000: Hmmm... I put it cause I already had created for other uncommited file, but it indeed is useless as of now. I’ll remove it.

marius851000: Done

UsernameFodder (approved):
Looks good, thanks!

PR #134, opened 2023-04-21
Defined unknown functions in ApplyItemEffect

Jawshoeuh:
Mostly changes related to figuring out what the functions in ApplyItemEffect did.

In case you wanted to know why I shrunk the dungeon::secret_item_room_weights and gave some its space the the trap_weights is because the function 0x022E7A30 (NA) specifically starts looking at DUNGEON_PTR + 0x02C932. The bounds on the check is 24 and that happens to line up with the number of traps that can be placed during dungeon generation (GRUDGE_TRAP).


headers/types/dungeon_mode/dungeon.h
```diff
@@ -597,17 +600,23 @@ struct dungeon {
...
-    bool pass_turn; // 0x78E: True if the leader has chosen to pass its turn
-    undefined field_0x78f;
-    bool thief_alert; // 0x790: If you've stolen from Kecleon (actual dungeon state)
+    bool pass_turn;        // 0x78E: True if the leader has chosen to pass its turn
+    bool drought_orb_flag; // 0x78F: True if a drought orb has been used.
+    bool thief_alert;      // 0x790: If you've stolen from Kecleon (actual dungeon state)
     // 0x791: If you've stolen from Kecleon (triggers music and other events?)
     bool thief_alert_event;
-    undefined field_0x792;
+    // 0x792: Controls when to play certain special music? Can be 0x7C (Outlaw), 0x7D
```
UsernameFodder, line 615: The phrasing here makes it seem like the flag itself can be 0x7C, etc. I think what you mean is that 0, 1, 2, 3, and 4 correspond to these music IDs? Can we rephrase this to be clearer?

headers/types/dungeon_mode/dungeon.h
```diff
@@ -1613,23 +1627,9 @@ struct dungeon {
...
-    undefined field_0x2c959;
-    undefined field_0x2c95a;
-    undefined field_0x2c95b;
-    undefined field_0x2c95c;
-    undefined field_0x2c95d;
-    undefined field_0x2c95e;
-    undefined field_0x2c95f;
-    undefined field_0x2c960;
-    undefined field_0x2c961;
+    uint16_t secret_room_item_weights[1400];
```
UsernameFodder, line 1630: Hmm, @End45 what do you think of this? The trap weights thing seems reasonable, but it also seems weird that `secret_room_item_weights` would be shorter than all the other lists before it.

headers/types/dungeon_mode/dungeon.h
```diff
@@ -1868,18 +1868,19 @@ struct dungeon {
...
-    undefined field_0x2cb04;
-    undefined field_0x2cb05;
-    undefined field_0x2cb06;
-    undefined field_0x2cb07;
+    // Related to playing dungeon music?
+    undefined2 field_0x2cb02;
+    // 0x2CB04: Related to playing dungeon music?
+    undefined2 field_0x2cb04;
+    // 0x2CB06: Current Dunegon Music ID
+    uint16_t dungeon_music_playing_id;
```
UsernameFodder, line 1883: Use `struct music_id_16` here

headers/types/dungeon_mode/dungeon_mode.h
```diff
@@ -138,8 +136,10 @@ struct statuses {
     // 0x2E: Turns left until residual healing for the status in statuses::reflect, if applicable
     uint8_t reflect_damage_countdown;
     uint8_t curse; // 0x2F: STATUS_CURSED if 1
-    undefined field_0x30;
-    undefined field_0x31;
+    // 0x30: Set to monster::0x6 of the monster the attacker (the one causing the decoy status).
```
UsernameFodder, line 139: Use the actual field names rather than offsets when possible, `monster::is_not_team_member`

headers/types/dungeon_mode/dungeon_mode.h
```diff
@@ -177,14 +177,24 @@ struct statuses {
...
+    // through the function?
+    undefined unk_speed_boost_tracker;
+    // 0x59: Maybe related to being a team member and new recruit? Set to 1 in TryRecruit
+    // and 0 in SpawnTeam. Also checked in EnemyEvolution to be 0 before evolving. Maybe to
+    // prevent a recently recruited ally from evolving after and or to add a monster to the
+    // assembly after the completion of a dungeon?
     undefined field_0x59;
-    bool in_action;            // 0x5A: Possibly a flag while in action
+    // 0x5A: Possibly a flag while in action. Could also be a flag that to cause the  burn from
+    // lava, heal a burn from water, and decreasce hunger in the walls.
```
UsernameFodder, line 196: \*decrease

headers/types/dungeon_mode/dungeon_mode.h
```diff
@@ -177,14 +177,24 @@ struct statuses {
...
+    // by the RunLeaderTurn function to know if the leader has changed their speed stage partway
+    // through the function?
+    undefined unk_speed_boost_tracker;
+    // 0x59: Maybe related to being a team member and new recruit? Set to 1 in TryRecruit
+    // and 0 in SpawnTeam. Also checked in EnemyEvolution to be 0 before evolving. Maybe to
+    // prevent a recently recruited ally from evolving after and or to add a monster to the
+    // assembly after the completion of a dungeon?
     undefined field_0x59;
-    bool in_action;            // 0x5A: Possibly a flag while in action
+    // 0x5A: Possibly a flag while in action. Could also be a flag that to cause the  burn from
```
UsernameFodder, line 195: There's a double-space before "burn"

headers/types/dungeon_mode/dungeon_mode.h
```diff
@@ -1023,13 +1039,18 @@ struct dungeon_generation_info {
     // 0x0: Set if the floor layout is guaranteed to be a Monster House, or the dungeon generation
     // algorithm fails
     bool force_create_monster_house;
-    undefined field_0x1;
-    undefined field_0x2;
+    // 0x1: Set if the locked door on the dungeon floor has already been open.
```
UsernameFodder, line 1042: has already been open**ed**? And is this referring to a prior visit to the dungeon? Otherwise I'm not sure why this would be set on the dungeon generation info at the start of a floor.

headers/types/dungeon_mode/dungeon_mode.h
```diff
@@ -1023,13 +1039,18 @@ struct dungeon_generation_info {
...
+    bool locked_door_opened;
+    // 0x2: Set if a kecleon shop was properly spawned.
+    bool kecleon_shop_spawned;
+    // 0x3: When a non-zero value, the one-room orb will fail.
     undefined field_0x3;
```
UsernameFodder, line 1047: Can we name this one?

headers/functions/overlay29.h
```diff
@@ -227,12 +228,18 @@ int GetMovePower(struct entity* entity, struct move* move);
 bool UpdateStateFlags(struct monster* monster, uint16_t mask, bool set_flags);
 void AddExpSpecial(struct entity* attacker, struct entity* defender, int base_exp);
 void EnemyEvolution(struct entity* enemy);
+bool LevelUpItemEffect(struct entity* attacker, struct entity* defender, int levels, bool message,
```
UsernameFodder, line 231: The first two params should be `user` and `target`

symbols/overlay29.yml
```diff
@@ -1673,6 +1673,15 @@ overlay29:
...
         r0: pointer to entity whose moves will be restored
+    - name: BoostIQ
+      address:
+        NA: 0x22FA144
+      description: |-
+        Tries to boost the target's IQ.
+        
+        r0: monster entity pointer
+        r1: iq boost
+        r2: bool suppress logs
```
UsernameFodder, line 1684: The signature says there's a return value?

headers/functions/overlay29.h
```diff
@@ -227,12 +228,18 @@ int GetMovePower(struct entity* entity, struct move* move);
...
+                       bool dialog);
 bool TryDecreaseLevel(struct entity* user, struct entity* target, int n_levels);
-bool LevelUp(struct entity* user, struct entity* target, bool message, undefined4 param_4);
+bool LevelUp(struct entity* user, struct entity* target, bool message, bool dialog);
 void EvolveMonster(struct entity* monster, undefined4* param_2, enum monster_id new_monster_id);
```
UsernameFodder, line 235: Since you swapped params 1 and 2 in the symbol tables, you should swap them here too

headers/functions/overlay29.h
```diff
@@ -227,12 +228,18 @@ int GetMovePower(struct entity* entity, struct move* move);
...
 uint8_t GetSleepAnimationId(struct entity* entity);
 bool DisplayActions(struct entity* param_1);
 void CheckNonLeaderTile(struct entity* entity);
+bool EndNegativeStatusCondition(struct entity* user, struct entity* target, bool animation,
+                                bool fail_message, bool remove_wrapping);
+bool EndNegativeStatusConditionWrapper(struct entity* user, struct entity* target, bool animation,
```
UsernameFodder, line 241: There shouldn't be a `remove_wrapping` param here according to your description

symbols/overlay29.yml
```diff
@@ -2492,6 +2514,38 @@ overlay29:
...
+        NA: 0x2305C28
+      description: |-
+        Calls EndNegativeStatusCondition with remove wrapping status false.
+        
+        r0: pointer to user
+        r1: pointer to target
+        r2: bool play animation
+        r3: bool log failure message
+        return: bool succesfully removed negative status
+    - name: TransferNegativeStatusCondition
```
UsernameFodder, line 2541: This one needs a signature in the headers

symbols/overlay29.yml
```diff
@@ -3577,6 +3661,33 @@ overlay29:
...
+        r2: flag to only perform the check for inflicting without actually inflicting
+        return: Whether or not the status could be inflicted
+    - name: TryInflictEyedropStatus
+      address:
+        NA: 0x231795C
+      description: |-
+        Inflicts the Eyedrop status condition on a target monster if possible.
+        
+        r0: user entity pointer
+        r1: target entity pointer
```
UsernameFodder, line 3681: Hmm it feels like this should return a success flag like the others. Does it leave a success flag in `r0` when it returns? Sometimes Ghidra doesn't catch it if callers never use the return value, but I like to include it in the signature if it's there.

symbols/overlay29.yml
```diff
@@ -3785,6 +3932,48 @@ overlay29:
         stack[0]: defender pointer
         stack[1]: thrown item pointer
         others: ?
+    - name: ApplyCheriBerryEffect
```
UsernameFodder, line 3935: This one needs a signature in the headers

symbols/overlay29.yml
```diff
@@ -3835,6 +4046,49 @@ overlay29:
...
+        r1: target entity pointer
+    - name: ApplyViaSeedEffect
+      address:
+        NA: 0x231D74C
+      description: |-
+        Tries to randomly teleport the target with a message for eating the seed.
+        
+        r0: user entity pointer
+        r1: target entity pointer
+    - name: ApplyGravelyrockEffect
```
UsernameFodder, line 4068: Needs a signature in the headers

symbols/overlay29.yml
```diff
@@ -3835,6 +4046,49 @@ overlay29:
...
+      address:
+        NA: 0x231D7C0
+      description: |-
+        Restores 10 hunger to the target and will raise the target's IQ if they are a bonsly
+        or sudowoodo.
+        
+        r0: user entity pointer
+        r1: target entity pointer
+    - name: ApplyGonePebbleEffect
```
UsernameFodder, line 4077: Needs a signature in the headers

symbols/overlay29.yml
```diff
@@ -3835,6 +4046,49 @@ overlay29:
...
+      address:
+        NA: 0x231D838
+      description: |-
+        Causes a few visual effects, temporarily changes the dungeon music to the Goodnight
+        track, and gives the target the enduring status.
+        
+        r0: user entity pointer
+        r1: target entity pointer
+    - name: ApplyGracideaEffect
```
UsernameFodder, line 4086: Needs a signature in the headers

headers/functions/overlay29.h
```diff
@@ -560,6 +589,7 @@ enum hidden_stairs_type GetHiddenStairsType(struct dungeon_generation_info* gen_
                                             struct floor_properties* floor_props);
 int GetFinalKecleonShopSpawnChance(int base_kecleon_shop_chance);
 void ResetHiddenStairsSpawn(void);
+void ApplyKeyEffect(struct entity* attacker, struct entity* defender);
```
UsernameFodder, line 592: Params should be `user` and `target`

UsernameFodder (changes requested):

headers/types/dungeon_mode/dungeon.h
End45: A strange situation indeed. The code pointed out by Hecka does access an address that is included in the last positions of the array and starts writing trap data there. I guess this change works, but it would be a good idea to point out the inconsistency in a comment.

headers/types/dungeon_mode/dungeon_mode.h
```diff
@@ -177,14 +177,24 @@ struct statuses {
...
+    bool pickup_flag;
     bool grudge;       // 0x54: STATUS_GRUDGE
     bool exposed;      // 0x55: STATUS_EXPOSED (Foresight/Odor Sleuth)
     bool type_changed; // 0x56: Flag for if the monster's type has been changed
     bool boss_flag;    // 0x57: Seems to be true for boss monsters
-    undefined field_0x58;
+    // 0x58: Appears to be a flag for when a monster increasces their speed. Maybe only used
+    // by the RunLeaderTurn function to know if the leader has changed their speed stage partway
+    // through the function?
+    undefined unk_speed_boost_tracker;
```
End45, line 189: This name makes it look like this field is related to the Speed Boost ability, could it be renamed?

headers/types/dungeon_mode/dungeon_mode.h
End45: The description isn't very clear either. "A flag that to cause ..."? Or it that a typo?

headers/types/dungeon_mode/dungeon_mode.h
Jawshoeuh: The function (0x023439D8 NA) that actually opens the locked door that's called by the ApplyKeyEffect specifically checks if the value is 0, if it is it will try to open the door and will set the value to 1. When it's 1, it just logs a can't be used message without checking. It's also initialized to 0 in GenerateFloor.

Since it's initialized to 0 in GenerateFloor I'm pretty sure it's a tracker on a by floor basis.

headers/types/dungeon_mode/dungeon_mode.h
Jawshoeuh: I wasn't exactly sure what to name it because I couldn't find where the value gets set, but I know it gets checked in the function DoMoveOneRoom. Did you have any suggestions?

symbols/overlay29.yml
Jawshoeuh: After double checking, I am pretty confident it does not return anything. It immediately returns after calling UpdateStatusIconFlags. It's probably this way because the function is exclusively called in ApplyItemEffect. The other functions are used for thinks like exclusive items or have their error message handled by ApplyItemEffect. This function has the error message inside.

headers/functions/overlay29.h
Jawshoeuh: I think you meant for EvolveMonster. After double checking my original work, I will be correcting this and making a change to EnemyEvolution as well. (I was double checking I meant to swap param 1 and 2).

headers/types/dungeon_mode/dungeon_mode.h
UsernameFodder: Okay, let's just fix the typo then.

headers/types/dungeon_mode/dungeon_mode.h
UsernameFodder: `unk_one_room_flag` sounds good to me

symbols/overlay29.yml
UsernameFodder: Okay, just wanted to make sure. Leave as is then.

headers/types/dungeon_mode/dungeon.h
Jawshoeuh: Upon further review, I'm pretty sure the trap weight array is 25 long and I've updated it to match (because I realized the loop starts at 0 and thus is 1 longer than I initially thought). Will leave a comment about the inconsistency around the secret room weights.

headers/functions/overlay29.h
```diff
@@ -392,12 +412,24 @@ bool IsInSpawnList(undefined* spawn_list, enum monster_id monster_id);
...
+void ApplyVileSeedEffect(struct entity* attacker, struct entity* defender);
 void ViolentSeedBoost(struct entity* attacker, struct entity* defender);
+void ApplyGinsengEffect(struct entity* attacker, struct entity* defender);
+void ApplyBlastSeedEffect(struct entity* attacker, struct entity* defender, bool thrown);
 void ApplyGummiBoostsDungeonMode(struct entity* user, struct entity* target,
                                  enum type_id gummi_type, int random_stat_boost);
 bool CanMonsterUseItem(struct entity* entity, struct item* item);
-void ApplyGrimyFoodEffect(struct entity* user, struct entity* target);
-void ApplyMixElixirEffect(struct entity* user, struct entity* target);
+void ApplyGrimyFoodEffect(struct entity* attacker, struct entity* defender);
```
UsernameFodder, line 426: All the functions in this block should be `user` and `target`, not `attacker` and `defender` to match what the symbol descriptions say.

headers/functions/overlay29.h
```diff
@@ -366,9 +375,20 @@ bool TryInflictLeechSeedStatus(struct entity* user, struct entity* target, bool
...
+                      int max_belly_boost, bool log_failure);
 void TryTransform(struct entity* user, struct entity* target);
+bool TryInflictBlinkerStatus(struct entity* user, struct entity* target, bool check_only,
+                             bool log_failure);
 bool IsBlinded(struct entity* entity, bool check_held_item);
+bool TryInflictCrossEyedStatus(struct entity* user, struct entity* target, bool check_only);
+void TryInflictEyedropStatus(struct entity* user, struct entity* target);
+bool TryInflictSlipStatus(struct entity* user, struct entity* target);
 void RestoreMovePP(struct entity* user, struct entity* target, int pp, bool suppress_logs);
+void ApplyProteinEffect(struct entity* attacker, struct entity* defender, int stat_boost);
```
UsernameFodder, line 388: Does it actually make sense to call these `attacker` and `defender`? I would think all these item effects should be `user` and `target` instead

headers/functions/overlay29.h
```diff
@@ -392,12 +412,24 @@ bool IsInSpawnList(undefined* spawn_list, enum monster_id monster_id);
 int ChangeShayminForme(struct entity* entity, int forme);
 void ApplyItemEffect(undefined4 param_1, undefined4 param_2, undefined4 param_3,
                      struct entity* attacker, struct entity* defender, struct item* thrown_item);
+void ApplyCheriBerryEffect(struct entity* attacker, struct entity* defender);
+void ApplyPechaBerryEffect(struct entity* attacker, struct entity* defender);
+void ApplyRawstBerryEffect(struct entity* attacker, struct entity* defender);
+void ApplyHungerSeedEffect(struct entity* attacker, struct entity* defender);
+void ApplyVileSeedEffect(struct entity* attacker, struct entity* defender);
 void ViolentSeedBoost(struct entity* attacker, struct entity* defender);
```
UsernameFodder, line 420: Let's change this to `user` and `target` with the rest of them.

UsernameFodder (changes requested):

headers/types/dungeon_mode/dungeon_mode.h
UsernameFodder: This still needs to be fixed

UsernameFodder: I fixed this one.

headers/types/dungeon_mode/dungeon.h
```diff
@@ -1626,12 +1625,13 @@ struct dungeon {
     // 0x2B332: Spawn weights for bazaar grab bag items. Same format as regular_item_weights.
     uint16_t grab_bag_item_weights[1416];
     // 0x2BE42: Spawn weights for secret room items in treasure boxes.
-    // Same format as regular_item_weights.
+    // Same format as regular_item_weights? For some reason the weights for the secret rooms
+    // are of a different length than the other item weight lists before it? The trap weights
+    // appear to overlap what would be the 16 categories entries? Maybe categories not used for
+    // this list?
```
End45, line 1631: > The trap weights appear to overlap what would be the 16 categories entries?

Not true, category weights are stored at the start of the list, the trap data overlaps with the end of the list.
Also, the code does write category weights here, so that can't be the reason.

```suggestion
    // Same format as regular_item_weights? For some reason the weights for the secret rooms
    // are of a different length than the other item weight lists before it. The trap weights
    // appear to overlap what would be entries 1400-1415.
```

End45 (changes requested):

headers/functions/overlay29.h
Jawshoeuh: I believe I made most of them attacker and defender because that's what they're called in ApplyItemEffect. Will change them to `user` and `target`. Would you like me to also change the function for ViolentSeedBoost to user target and/or change its name to ApplyViolentSeedBoost to match the other functions?

UsernameFodder: Yeah let's change all the item effect stuff to use "user" and "target" (in both the symbol descriptions and the function headers; what we say in one should match the other). The attacker/defender terminology comes from the damage calcs and move effects, but I don't think it really makes sense for items since there are so many items that don't fit that mold.

> Would you like me to also change the function for ViolentSeedBoost to user target and/or change its name to ApplyViolentSeedBoost to match the other functions?

Yeah that sounds good to me. The item effects should be consistent, like how the move effects are.

UsernameFodder (approved):
Thanks! It's great to be getting some of these item effects tabulated.

PR #137, opened 2023-05-07
Document interrupt flag functions

End45:
Adds a few functions that modify the CPU's interrupt flags. Used by the game to enable/disable/restore interrupts.

Also adds another WaitForever function.


UsernameFodder (approved):
Thanks!

PR #143, opened 2023-05-28
Port Adex's script_variable_meanings.txt

UsernameFodder:
Mostly a verbatim port, with minor tweaks and additions.

There's some obvious follow-up research that can be done here:
- Investigating the raw function/data offsets mentioned in the comments and adding actual symbols for them
- Locating data symbols that back some of the special script variables whose values don't come from the script var value table, and/or investigating the relationship between these script variables and existing  data symbols.
- Tracing special process code to glean more information about related script variables
- Tracing known ARM9 functions that definitely or likely coincide with script variables, such as `Load/SaveScriptVariableValue`, `GetPerformanceFlagWithChecks`, etc.

### DIFF
--- a/headers/types/ground_mode/enums.h
+++ b/headers/types/ground_mode/enums.h
@@ -4,11 +4,31 @@
 #define HEADERS_TYPES_GROUND_MODE_ENUMS_H_
 
 // Predefined variables available to the script engine.
+// Also see struct script_var_value_table, where the corresponding values for these variables are
+// stored.
+// TODO: replace raw address references with actual symbol names
 enum script_var_id {
+    // Its only purpose is to be checked at [NA]0x204C448, which compares the saved value to the
+    // default value in the script variable table (which is 1 in base game). If the values are not
+    // equal, then the save file is treated as corrupted, which thus gets deleted.
     VAR_VERSION = 0,
+    // Seemingly unused?
     VAR_CONDITION = 1,
+    // Two indexed values determined via the Top Menu. These are referenced by Unionall for general
+    // control flow.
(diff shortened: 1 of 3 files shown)


headers/types/ground_mode/enums.h
```diff
@@ -18,104 +38,394 @@ enum script_var_id {
...
+    // True to its name, this variable is randomly generated in the function [NA]0x205EB5C, which is
+    // called within GenerateDailyMissions. There also seems to be an exception where the game
+    // doesn't generate the species randomly (for Magnemite, maybe).
     VAR_RANDOM_REQUEST_NPC03_KIND = 68,
+    // The value of the word at [NA]0x20AFF28. Saved at [NA]0x2048990, which is done by the special
+    // process INIT_MAIN_TEAM_AFTER_QUIZ. Changing the variable doesn't seem to do much, so its true
+    // purpose is unknown.
     VAR_CONFIG_COLOR_KIND = 69,
+    // Its purpose is unknown, but it's initialized to 0 at [NA]0x204B0E4.
     VAR_ROM_VARIATION = 70,
```
End45, line 259: Wasn't this the one used to signal if the game is the demo version or not? (1 = Demo, 0 = Release)

UsernameFodder: I dunno, but @Adex-8x might.

Adex-8x: At the very least, setting this variable to 1 (whether by a script command or editing it in RAM directly) doesn't affect the `variation` check used by the script engine, so I'm not sure on this.

PR #147, opened 2023-06-21
Info dump 32-34/66: Invisibility and player input

End45:

- Determining when or how to display monsters
- Player input and automated player actions

I also updated `OpenMenu`. I don't know why it had a fourth parameter listed, that doesn't seem to be the case.


UsernameFodder:
Fixed a couple things in 5924cc3d944e85bc28d8b33e58db2b0ffe53a2c0.

PR #149, opened 2023-06-23
Info dump 36-43/66: Many things

End45:

- Exclusive items
- Quicksaving
- The grid that shows up in dungeon mode when you hold Y
- Moving pokémon to a different position
- Dungeon tiles
- Dungeon results screen ("The Last Outing")
- Fixed rooms
	- Includes some functions about spawning and initializing entities, since I documented those when I started researching fixed rooms

`InitMonster` has been renamed to `SubInitMonster`, since there's another function above it that is the main function responsible for initializing a new monster.

### DIFF
--- a/headers/functions/arm9.h
+++ b/headers/functions/arm9.h
@@ -108,6 +108,8 @@ void OpenPackFile(struct pack_file_opened* pack_file, const char* file_name);
 uint32_t GetFileLengthInPack(struct pack_file_opened* pack_file, uint32_t file_index);
 uint32_t LoadFileInPack(struct pack_file_opened* pack_file, void* output_buffer,
                         uint32_t file_index);
+void GetDungeonResultMsg(union damage_source damage_source_or_result, void* buffer, int buffer_size,
+                         undefined* param_4);
 union damage_source GetDamageSource(enum move_id, enum item_id);
 enum item_category GetItemCategoryVeneer(enum item_id item_id);
 enum move_id GetItemMoveId16(enum item_id item_id);
@@ -117,6 +119,7 @@ bool IsEdible(enum item_id item_id);
 bool IsHM(enum item_id item_id);
 bool IsGummi(enum item_id item_id);
 bool IsAuraBow(enum item_id item_id);
+bool IsChest(enum item_id item_id);
 void InitItem(struct item* item, enum item_id item_id, uint16_t quantity, bool sticky);
 void InitStandardItem(struct item* item, enum item_id item_id, bool sticky);
 int GetDisplayedBuyPrice(struct item* item);
@@ -188,6 +191,7 @@ bool SpecialProcAddItemToBag(struct bulk_item* item);
 bool AddItemToBagNoHeld(struct item* item);
 bool AddItemToBag(struct item* item, int held_by);
 bool ScriptSpecialProcess0x39(void);
+int CountNbItemsOfTypeInStorage(enum item_id item_id);
 int CountItemTypeInStorage(struct bulk_item* item);
 bool RemoveItemsTypeInStorage(struct bulk_item* item);
 bool AddItemToStorage(struct bulk_item* item);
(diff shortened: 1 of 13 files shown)


symbols/overlay11.yml
```diff
@@ -336,6 +336,15 @@ overlay11:
...
         return: number of characters printed, excluding the null-terminator
+    - name: GetExclusiveItemRequirements
+      address:
+        EU: 0x230B8D4
+        NA: 0x230AF38
+      description: |-
+        Used to calculate the items required to get a certain exclusive item in the swap shop.
+        
+        r0: ?
+        r1: ?
```
UsernameFodder, line 347: The signature suggests there's an unknown return type. Is that true?

headers/types/dungeon_mode/dungeon.h
```diff
@@ -506,7 +506,9 @@ struct dungeon {
     undefined field_0x745;
     undefined field_0x746;
     undefined field_0x747;
-    struct dungeon_id_8 id;             // 0x748
+    // 0x748: Current dungeon ID. Is actually a dungeon_floor_pair struct that also contains
```
UsernameFodder, line 509: Is there a reason not to just change this to be `struct dungeon_floor_pair`? Using the actual struct seems preferable to a comment like this.

headers/functions/arm9.h
```diff
@@ -108,6 +108,8 @@ void OpenPackFile(struct pack_file_opened* pack_file, const char* file_name);
 uint32_t GetFileLengthInPack(struct pack_file_opened* pack_file, uint32_t file_index);
 uint32_t LoadFileInPack(struct pack_file_opened* pack_file, void* output_buffer,
                         uint32_t file_index);
+void GetDungeonResultMsg(union damage_source damage_source_or_result, void* buffer, int buffer_size,
```
UsernameFodder, line 111: Is `buffer` always a string output? If so I think using `char*` would probably make more sense.

symbols/arm9.yml
```diff
@@ -1235,6 +1246,17 @@ arm9:
         
         r0: item ID
         return: bool
+    - name: IsChest
```
UsernameFodder, line 1249: Can we use the term "treasure box" to match the category name, rather than "chest"? It's less ambiguous than "chest" which could conceivably also refer to things like the Miracle/Wonder chest.

symbols/overlay29.yml
```diff
@@ -306,6 +306,26 @@ overlay29:
...
+    - name: UpdateEnityPixelPos
+      address:
+        EU: 0x22E2380
+        NA: 0x22E1A40
+      description: |-
+        Updates an entity's pixel_pos field using the specified pixel_position struct, or its own pos field if it's null.
+        
+        r0: Entity pointer
+        r1: Pixel position to use, or null to use the entity's own position
+    - name: InitEnemyEntity
```
UsernameFodder, line 318: The convention with `Init*` functions is that they take a pointer to initialize in-place, and usually have no return value. If this actually spawns the enemy on the floor somewhere, I think `SpawnEnemyEntity` would make more sense. If not, maybe we could rename this to something like `CreateEnemyEntity`?

UsernameFodder (changes requested):
Made some minor fixes for obvious things on top of the more discussion-worthy comments.

symbols/overlay11.yml
End45: I don't know if it returns anything. I took another look and it doesn't seem like it does, so replacing the return type with `void` should be alright.

headers/types/dungeon_mode/dungeon.h
End45: Technically, the ASM accesses the values directly most of the time. I thought leaving it like that wouldn't be a bad idea, since they are relevant fields and I didn't want to hide them under a struct.
But I didn't give it much thought, it could be changed too.

headers/functions/arm9.h
End45: Right

symbols/arm9.yml
End45: Sure

symbols/overlay29.yml
End45: I didn't go with `SpawnEnemyEntity` because it doesn't spawn the enemy, that's handled by `SpawnMonster`, which calls this function. `CreateEnemyEntity` sounds good, I'll change it.

headers/types/dungeon_mode/dungeon.h
UsernameFodder: Hmm okay. I think it's fine as is then, let's just leave it.

symbols/overlay11.yml
UsernameFodder: Changed to `void`

UsernameFodder (approved):

PR #150, opened 2023-06-25
Add script to add docstrings to functions/globals

UsernameFodder:
When viewing the C headers directly or including them in other source code, it can be annoying to have to cross-reference the symbol table descriptions for functions and globals. This PR adds:
- A script that cross-references the C headers and the symbol tables to annotate the C headers with Javadoc-style docstrings.
- A separate `headers-with-docstrings` artifact, parallel to the existing `headers` artifact. In some cases, it may be preferable to download the headers with docstrings. But since `add_header_docstrings.py` is somewhat hacky and experimental, and because the docstrings add a bunch of potentially unneeded cruft to the headers, make a separate artifact rather than modifying the existing release artifact.


tech-ticks (approved):
Thanks!

PR #156, opened 2023-07-08
Info regarding portrait box, live actor direction and music in dungeon

marius851000:
I’m still a bit unsure in which order I want to add stuff in pmdsky-debug. I think I’ll for now focus in the palette in overlay11 stuff. But I did that in the meantime.

Also, I think portrait_box is the "kaomodo" struct in this PR: https://github.com/UsernameFodder/pmdsky-debug/pull/21/files

### DIFF
--- a/headers/types/common/common.h
+++ b/headers/types/common/common.h
@@ -125,20 +125,13 @@ struct dialog_box {
 };
 ASSERT_SIZE(struct dialog_box, 224);
 
-// Structure for dialog boxes with portraits?
+// Represent a portrait that appear inside a dialogue box
 struct portrait_box {
-    undefined field_0x0;
-    undefined field_0x1;
-    undefined field_0x2;
+    struct monster_id_16 monster_id;
+    struct portrait_emotion_8 portrait_emotion;
     undefined field_0x3;
-    undefined field_0x4;
-    undefined field_0x5;
-    undefined field_0x6;
-    undefined field_0x7;
-    undefined field_0x8;
-    undefined field_0x9;
-    undefined field_0xa;
-    undefined field_0xb;
+    undefined4 field_0x4;
+    undefined4 field_0x8;
     undefined field_0xc;
     undefined field_0xd;
     undefined field_0xe;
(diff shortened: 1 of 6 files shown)


headers/types/common/common.h
```diff
@@ -144,20 +144,13 @@ struct dialog_box {
 };
 ASSERT_SIZE(struct dialog_box, 224);
 
-// Structure for dialog boxes with portraits?
+// Represent a portrait that appear inside a dialogue box
 struct portrait_box {
```
UsernameFodder, line 148: > Also, I think portrait_box is the "kaomodo" struct in this PR: https://github.com/UsernameFodder/pmdsky-debug/pull/21/files

Hmm I think you're very likely correct. Could we maybe copy the fields over from https://github.com/UsernameFodder/pmdsky-debug/pull/21 (with comments saying that they need verification)?

UsernameFodder (changes requested):

UsernameFodder:
(The formatting error is should go away once you rebase this branch against `master` and rerun the formatter)

headers/types/common/common.h
UsernameFodder: Actually, on second thought, copying over just that stuff from #21 probably doesn't make sense out of context. We can always port that stuff another day, or try to get the original PR merged.

headers/types/common/enums.h
```diff
@@ -708,6 +708,35 @@ enum monster_id {
 ENUM_16_BIT(monster_id);
 #pragma pack(pop)
 
+// Emotion for portraits
+enum face_emotion {
+    FACE_NORMAL = 0,
```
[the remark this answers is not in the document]
marius851000: The reason I used FACE is cause it’s what SkyTemple use, but I guess EMOTION_ will do just fine too.

headers/types/common/common.h
marius851000: I agree with that (thought kaomodo may not be an appropriate name for this struct, given that’s also how the file format is called)

UsernameFodder (approved):
Thanks! This portrait stuff is pretty useful info to have.

PR #157, opened 2023-07-08
Add a shell scripts that run the various checks

marius851000:
I often make mistakes that are caught by these tools. This script will run the various checks (that I’m aware of). It work on my machines, and should likely work on other unix-like OS with the required programs installed.

### DIFF
--- a/check_and_format.sh
+++ b/check_and_format.sh
@@ -1,0 +1,12 @@
+#!/bin/sh
+
+set -e
+
+pushd headers
+make
+make format
+make symbol-check
+popd
+
+cargo run -- check -r -Vvbomu -d screaming_snake_case -f pascalcase -f snake_case symbols/**.yml
+cargo run -- fmt -r symbols/**.yml
--- a/check_and_format.sh
+++ b/check_and_format.sh
@@ -1,12 +1,14 @@
 #!/bin/sh
 
+# Convenience script that runs the checks and formatters for both the C headers and symbol tables.
+# Expects the working directory to be the repo root.
+
 set -e
 
-pushd headers
-make
-make format
-make symbol-check
-popd
+make -C headers
+make -C headers format
+make -C headers symbol-check
 
-cargo run -- check -r -Vvbomu -d screaming_snake_case -f pascalcase -f snake_case symbols/**.yml
-cargo run -- fmt -r symbols/**.yml
+cargo run --release -- check -r -Vvbomu -d screaming_snake_case -f pascalcase -f snake_case symbols/*.yml
+cargo run --release -- fmt -r symbols/*.yml
+


check_and_format.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh
+
+set -e
+
+pushd headers
```
UsernameFodder, line 5: If one of the `make` commands fails, I think the script will exit before the `popd`, and the working directory won't be reset. Can we use `make -C headers` instead of `make`? This tells `make` to change directories before running its rules, without needing to change directories manually.

check_and_format.sh
```diff
@@ -0,0 +1,13 @@
...
+make
+make format-check
+make symbol-check
+#make symbol-check-extra
```
UsernameFodder, line 9: Let's delete this unused code

check_and_format.sh
```diff
@@ -0,0 +1,13 @@
...
+popd
+
+cargo run -- check -r -Vvbomu -d screaming_snake_case -f pascalcase symbols/**.yml
```
UsernameFodder, line 12: I just realized I forgot to update the docs, but this should be:

```suggestion
cargo run --release -- check -r -Vvbomu -d screaming_snake_case -f pascalcase -f snake_case symbols/*.yml
```

check_and_format.sh
```diff
@@ -0,0 +1,13 @@
...
+cargo run -- fmt -r symbols/**.yml
```
UsernameFodder, line 13: We're doing a real format here, but with the headers we're only doing a format check. I think we should be consistent and make this a check too:

```suggestion
cargo run --release -- fmt --check -r symbols/*.yml
```

UsernameFodder (changes requested):

check_and_format.sh
marius851000: That’s not a problem if the directory is invalid, given the script is called as a child process, and so change of the current working dir by the child won’t impact the parent process (the user’s shell).

check_and_format.sh
marius851000: Oh! I thought the format-check was both formatting and checking. I’ll instead update it so it format both.

check_and_format.sh
UsernameFodder: Oh yeah you're right. Never mind then.

UsernameFodder: Actually, now that I think about this more, `pushd` and `popd` are bash-only (`/bin/bash`), not POSIX shell (`/bin/sh`). So I would still prefer `make -C headers` since that's more compatible if someone's default shell isn't bash.

UsernameFodder (approved):
I made a couple minor tweaks per comments I've made previously. Can you double-check that this still works properly in your environment? If it works them this'll be good to merge.

check_and_format.sh
marius851000: Thanks for the info. Will do.
(actually, you already did that)

marius851000:
I confirm those changes work well in my environment (NixOS/Linux)

UsernameFodder:
Thanks! Hopefully this makes contributing a little easier.

PR #159, opened 2023-07-09
Small findings regarding VRAM banks management

marius851000:
Just two function and a static variable.

Hope this one doesn’t have too much issues.

*while trying to figure out a certain portion of the palette management stuff, with a call tree that eventually call those*


UsernameFodder (approved):
I added the NA offsets, plus some other minor tweaks. Looks good otherwise.

PR #160, opened 2023-07-09
Starting analysis of the graphic part of the engine

marius851000:
In the end, I think not making too big commits might be a good idea.
(I say starting, but that’s actually research I did during the jam. Plus new research too. Plus older research also.)


symbols/arm9.yml
```diff
@@ -3054,6 +3054,28 @@ arm9:
...
+        EU: 0x201C14C
+      description: |-
+        Initialise the animation_control structure, and set a certain value in a bitflag to 1
+        
+        r0: animation_control
+    - name: InitAnimationControlWithSetClone
+      address:
+        EU: 0x201C168
+      description: |-
+        Identical to InitAnimationControlWithSet
```
UsernameFodder, line 3075: By "identical", do you mean all the instructions are exactly the same? If so, we should just give `InitAnimationControlWithSet` a second address, i.e.

```yml
    - name: InitAnimationControlWithSet
      address:
        EU:
          - 0x201C14C
          - 0x201C168
```

headers/types/common/graphics.h
```diff
@@ -0,0 +1,73 @@
...
+    undefined field32_0x44;
+    undefined field33_0x45;
+    undefined field34_0x46;
+    undefined field35_0x47;
+    struct WanAnimationFrame** field36_0x48;
+    struct WanAnimationFrame** field37_0x4c;
+    struct WanOffset* wan_offsets;
+    undefined* wan_frames;
+    void** wan_fragments_byte_store;
+    struct WanPalettes* wan_palettes;
```
UsernameFodder, line 51: I don't see any definitions for `struct WanAnimationFrame`, `struct WanOffset`, and `struct WanPalettes`. Can we add them? The struct names should also be `snake_case`

headers/functions/arm9.h
```diff
@@ -307,6 +307,9 @@ void PlaySeFullSpec(int param_1, int param_2, int param_3, int param_4);
 void SeChangeVolume(int param_1, int param_2, int param_3);
 void SeChangePan(int param_1, int param_2, int param_3);
 void StopSe(int param_1, int param_2);
+void InitAnimationControl(struct animation_control*);
+void InitAnimationControlWithSet(struct animation_control*);
+void InitAnimationControlWithSetClone(struct animation_control*);
```
UsernameFodder, line 312: Let's give these `struct animation_control*` parameters names

UsernameFodder (changes requested):

headers/types/common/graphics.h
marius851000: Oops. I used a different in Ghidra, and forgot to fix the export (however I’m a bit surprised this hasn’t been detected by a check)

symbols/arm9.yml
marius851000: They are. Will fix that.
(curious part is that they are adjacent. I’ll presume one of them was a variant of the other at some point and the devs then made them the same function. And the compiler didn’t optimized that).

headers/functions/arm9.h
marius851000: I’m lacking a bit of inspiration for a name, but I’ll use `animation_control` (didn’t even knew C allowed unnamed parameters)

UsernameFodder: Yeah `animation_control` is fine with me. C/C++ allows function declarations without parameter names (since the parameter names aren't part of the function's "type"). But some projects (like `pmdsky-debug`) follow the convention of naming them in declarations for clarity.

headers/types/common/graphics.h
UsernameFodder: > however I’m a bit surprised this hasn’t been detected by a check

It's a bit of a quirk with the C language itself. You're allowed to use _pointers_ to undeclared types (called "incomplete types") in struct fields, because from the compiler's perspective, all struct pointers are just pointers, which are just integers. The idea is that you could define a struct to have a pointer to another struct, and define the other struct in another file somewhere else.

symbols/arm9.yml
UsernameFodder: > (curious part is that they are adjacent. I’ll presume one of them was a variant of the other at some point and the devs then made them the same function. And the compiler didn’t optimized that).

I've seen this in a couple other places too. This seems like a plausible explanation of how it could happen, though I don't know for sure.

UsernameFodder (approved):
Looks good, thanks!

PR #162, opened 2023-07-11
Added information about setting world map camera and mode

marius851000:
From a patch I have yet to debug on NA and release.


UsernameFodder (approved):
LGTM

PR #166, opened 2023-07-29
Info dump 54/66: Mobility

End45:
Adds functions mostly related to monster mobility.
A couple of existing functions have been renamed.

I started researching this area and then realized halfway through that I had already documented most of it, so I ended up accidentally dumping a new section of my notes. Oops!

### DIFF
--- a/symbols/overlay29.yml
+++ b/symbols/overlay29.yml
@@ -2333,7 +2333,7 @@ overlay29:
         EU: 0x23002F4
         NA: 0x22FF8C8
       description: |-
-        Modifies the given movility type to account for All-Terrain Hiker and Absolute Mover, if the user has them.
+        Modifies the given mobility type to account for All-Terrain Hiker and Absolute Mover, if the user has them.
         
         r0: Entity pointer
         r1: Mobility type
--- a/symbols/overlay29.yml
+++ b/symbols/overlay29.yml
@@ -2290,6 +2290,28 @@ overlay29:
         
         r0: Entity pointer
         return: True if the specified monster can't act because of a status problem, false otherwise.
+    - name: GetMobilityTypeCheckSlip
+      address:
+        EU: 0x2300058
+        NA: 0x22FF62C
+      description: |-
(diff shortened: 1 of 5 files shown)


symbols/overlay29.yml
```diff
@@ -4561,15 +4610,15 @@ overlay29:
         
         r0: user entity pointer
         r1: target entity pointer
-    - name: HasConditionalGroundImmunity
+    - name: IsLevitating
```
UsernameFodder, line 4613: I don't really like this name. "Levitating" implies the levitate ability, which is notably not what this function does. That's why I called it `HasConditionalGroundImmunity`. It applies to things that can give ground immunity that aren't permanent.

I guess since this applies to more than just type matchups, we should still rename it though. Can we do something like `IsHovering` or `IsFloating`?

symbols/overlay29.yml
```diff
@@ -2290,6 +2290,28 @@ overlay29:
...
+        NA: 0x22FF62C
+      description: |-
+        Returns the mobility type of a monster species, accounting for STATUS_SLIP.
+        
+        The function also converts MOBILITY_LAVA and MOBILITY_WATER to other values if required.
+        
+        r0: Monster species
+        r1: True if the monster can walk on water
+        return: Mobility type
+    - name: GetMobilityTypeCheckSlipAndLevitate
```
UsernameFodder, line 2305: It seems misleading to use the word "Levitate" when it doesn't check for the levitate ability.

UsernameFodder (changes requested):

symbols/overlay29.yml
End45: Good point. I like `IsFloating`, I'll change it.

symbols/overlay29.yml
End45: Updated to match `IsFloating`

UsernameFodder (approved):
If you don't mind, I added the mention of "typing" (alongside ability) back to `IsFloating`, since IMO it's still relevant, especially for the type-advantage code context.

PR #167, opened 2023-07-29
Properly document MainGame

End45:
Renames `CheckEndDungeon` to `MainGame` and documents it in more detail.

### DIFF
--- a/symbols/overlay10.yml
+++ b/symbols/overlay10.yml
@@ -137,7 +137,7 @@ overlay10:
       description: |-
         Contains several functions that handle switching between ground and dungeon mode. It also handles other situations, like what happens right after exiting a dungeon.
         
-        The name comes from a debug print with the string "MainGame enter dungeon mode %d %d". Despite its name, it's not a main function. It doesn't get called until the player selects the option to resume a saved game, and it might return under certain circumstances.
+        The function doesn't get called until the player selects the option to resume a saved game and doesn't return until the player returns to the main menu.
         
         r0: End condition code? Seems to control what tasks get run and what transition happens when the dungeon ends
         return: return code?
--- a/symbols/overlay10.yml
+++ b/symbols/overlay10.yml
@@ -129,15 +129,15 @@ overlay10:
         
         r0: Tileset ID
         return: True if the tileset ID corresponds to a background, false if it corresponds to a regular tileset
-    - name: CheckEndDungeon
+    - name: MainGame
       address:
         EU: 0x22C31E4
         NA: 0x22C288C
         JP: 0x22C3F74
       description: |-
-        Do the stuff when you lose in a dungeon.
+        Contains several functions that handle switching between ground and dungeon mode. It also handles other situations, like what happens right after exiting a dungeon.
         
-        Note: unverified, ported from Irdkwia's notes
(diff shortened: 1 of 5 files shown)


symbols/overlay10.yml
```diff
@@ -118,15 +118,15 @@ overlay10:
...
       address:
         EU: 0x22C31E4
         NA: 0x22C288C
         JP: 0x22C3F74
       description: |-
-        Do the stuff when you lose in a dungeon.
+        Contains several functions that handle switching between ground and dungeon mode. It also handles other situations, like what happens right after exiting a dungeon.
         
-        Note: unverified, ported from Irdkwia's notes
+        The name comes from a debug print with the string "MainGame enter dungeon mode %d %d". Despite its name, it's not a main function. It doesn't get called until the player selects the option to resume a saved game, and it might return under certain circumstances.
```
UsernameFodder, line 129: This doesn't seem like a solid reason to call the entire function "MainGame" when it's not a good descriptor of what the function actually does. It seems like the debug message is only printed for a certain case where `r5 == 5`, so this could be referring to an enum name, a parent function related to the circumstances where the condition will be true, or something else. The function also has two other debug messages: "dungeon fade out mode %d\n" and `dungeon getout result %d\n`. These don't mention "MainGame" anywhere.

All three debug messages mention "dungeon". What about something like `DungeonModeTransition`?

End45: I did see the other messages, but I didn't realize the `r5 == 5` condition. Hm.

> All three debug messages mention "dungeon". What about something like DungeonModeTransition?

The function is also responsible for loading the ground mode overlay, calling its main function and then unloading it whenever ground mode is entered. It also does this for other overlays, so `DungeonModeTransition` doesn't seem to cover everything the function does.

If it helps, there's two other functions above this one. The topmost one calls the middle one and then `WaitForever`. The middle one has an infinite loop and never returns. It's quite big, and calls this function in multiple places.
I don't know the details of how those other two functions work, so I didn't include them.

UsernameFodder: > The function is also responsible for loading the ground mode overlay, calling its main function and then unloading it whenever ground mode is entered. It also does this for other overlays, so DungeonModeTransition doesn't seem to cover everything the function does.

So is it kind of an orchestrator for switching the overlay that's currently running the "main game" in any given mode (overlay11 for ground mode, overlay29 for dungeon mode, other overlays for other things)? What about something like `MainOverlayTransition` or `SwitchMainOverlay`?

End45: More info: This function runs when the save file is loaded and returns when the player goes back to the main menu. I tried going into a dungeon, doing sentry duty and advancing a day, and none of that caused a return.
Maybe we could call it `MainGameplayLoop` and also name the one above it `MainLoop`.

theCapypara: But if you say "run" do you mean it's called? If it's called then I wouldn't call it a loop function. A loop function would only ever be returned to by returning.

End45: It does get called. The loop it contains is not infinite, but it does run most of the time while the game is playing, so I think that's a good description of what the function does.

UsernameFodder: In that case it seems like the original proposal of `MainGame` actually _is_ an accurate name. I say we just leave it as is then.

UsernameFodder: But can you update the description with those new findings?

UsernameFodder (approved):

PR #168, opened 2023-07-30
Some 3d engine documentation

marius851000:
I had this lying around since about 2 weeks, but finally decided to do the needed finalisation.

The gist of that is the game use the 3D rendering system to render a few things. From what I’ve seen, it mostly include weather and text box border/background, but could be easily extended (and is what Irdkwia used for the moving spirit effect in the "angry" part of Tangent’s hack (that I can’t remember the name of).

### DIFF
--- a/headers/types/common/graphics.h
+++ b/headers/types/common/graphics.h
@@ -108,7 +108,7 @@ ASSERT_SIZE(struct render_3d_element, 52);
 
 // Function that will use the relevant 3d render API to render a render_3d_element from the
 // RENDER_3D queue.
-typedef void (*render_3d_element_concrete)(struct render_3d_element*);
+typedef void (*render_3d_element_concrete_fn_t)(struct render_3d_element* element);
 
 // A global, unique structure that stores element relating to the 3d engine, in particular the list
 // of elements to render later in the frame.
--- a/headers/types/common/graphics.h
+++ b/headers/types/common/graphics.h
@@ -73,12 +73,12 @@ struct animation_control {
 ASSERT_SIZE(struct animation_control, 124);
 
 // Represent a single element to render using the 3D engine
-// This structure is used in two different way. The first one is for planning rendering, that will
+// This structure is used in two different ways. The first one is for planning rendering, that will
 // then call function that will call a function that will then add themm to RENDER_3D for rendering
 // later in the frame.
 struct render_3d_element {
     undefined2 render_function_id; // range from 0 to 3 (included)
-    undefined2 field1_0x2;         /* appears to be a render priority level. Impact sorting. */
+    undefined2 field1_0x2;         // appears to be a render priority level. Impact sorting.
     undefined4 field2_0x4;
     undefined4 field3_0x8;
     uint16_t x_tileset_start;
@@ -103,8 +103,8 @@ struct render_3d_element {
 };
 ASSERT_SIZE(struct render_3d_element, 52);
 
-// A global, unique structure that store element relating to the 3d engine, in particular the list
-// of element to render later in the frame.
+// A global, unique structure that stores element relating to the 3d engine, in particular the list
+// of elements to render later in the frame.
 struct render_3d_global {
     uint16_t current_index;
     uint16_t max_index; // Seems to consistently be 128, size of render_stack
--- a/headers/types/common/graphics.h
+++ b/headers/types/common/graphics.h
@@ -24,12 +24,13 @@ struct animation_control {
     struct vec2_16 anim_frame_offset;
     struct vec2_16 anim_frame_shadow_offset;
     uint32_t anim_frame_flag;
-    uint32_t anim_frame_flag_sum; // ORed from the current WAN animation frame flag
-    undefined2 field18_0x30;
+    uint32_t anim_frame_flag_sum;     // ORed from the current WAN animation frame flag
+    uint16_t animations_or_group_len; // number of animation group Chara sprites, other number of
+                                      // animation in the first animation group
     undefined2 field19_0x32;
     undefined field20_0x34;
     undefined field21_0x35;
-    undefined2 field22_0x36;
+    undefined2 field22_0x36; // seems to be an index into the wan fragments table
     undefined2 field23_0x38;
     uint16_t anim_frame_frame_id;
     undefined2 field25_0x3c;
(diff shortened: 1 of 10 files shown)


headers/types/files/wan.h
```diff
@@ -38,7 +38,8 @@ struct wan_palettes {
     uint16_t unk1;
     uint16_t nb_color;
     uint16_t unk2;
-    uint16_t unk3;
+    uint8_t unk3;
+    uint8_t unk5;
```
UsernameFodder, line 42: Is it really necessary to keep `unk4` as is? I would rather rename this one to `unk4` and `unk4` to `unk5` so that everything stays in order.

headers/types/common/graphics.h
```diff
@@ -71,4 +72,97 @@ struct animation_control {
...
+    undefined field19_0x30;
+    char field20_0x31;
+    bool field21_0x32;
+    undefined field22_0x33;
+};
+ASSERT_SIZE(struct render_3d_element, 52);
+
+// A global, unique structure that store element relating to the 3d engine, in particular the list
+// of element to render later in the frame.
+struct render_3d_global {
```
UsernameFodder, line 108: I don't see a data symbol that uses this type. Did you forget to add `RENDER_3D` somewhere? You mention this struct a bunch in the descriptions.

headers/types/common/graphics.h
```diff
@@ -71,4 +72,97 @@ struct animation_control {
 };
 ASSERT_SIZE(struct animation_control, 124);
 
+// Represent a single element to render using the 3D engine
+// This structure is used in two different way. The first one is for planning rendering, that will
+// then call function that will call a function that will then add themm to RENDER_3D for rendering
```
UsernameFodder: > that will then call function that will call a function

Is this intentional because there are two layers of function calls, or is this a typo?

symbols/arm9/itcm.yml
```diff
@@ -215,3 +231,11 @@ itcm:
         Note: This symbol isn't actually part of the ITCM, it gets created at runtime on the spot in RAM that used to contain the code that was moved to the ITCM.
         
         type: struct mem_block[256]
+    - name: RENDER_3D_FUNCTIONS
```
UsernameFodder, line 234: Needs a corresponding declaration in `headers/data/arm9/itcm.h`

symbols/arm9/itcm.yml
```diff
@@ -215,3 +231,11 @@ itcm:
...
+      address:
+        EU: 0x20B3DE0
+        EU-ITCM: 0x1FF8120
+      description: |-
+        Pointers to the 4 function avalaible from render_3d_element (in ITCM)
+        
+        type: void* [4] // array of 4 pointer to functions. Take 1 parameter, a pointer to the render_3d_element
```
UsernameFodder, line 241: Function pointers are the one case where this project uses `typedef`. For example, see `move_effect_fn_t` in `dungeon_mode.h`. In this case, instead of trying to describe it in the symbol description, you could define something like

```c
typedef void (*render_3d_fn_t)(struct render_3d_element* element);
```

and then define the global variable like this:

```c
extern render_3d_fn_t RENDER_3D_FUNCTIONS[4];
```

symbols/arm9/itcm.yml
```diff
@@ -29,6 +29,22 @@ itcm:
     
     In Explorers of Sky, a fixed region of the ARM9 binary appears to be loaded in the ITCM at all times, and seems to contain functions related to the dungeon AI, among other things. The ITCM has a max capacity of 0x8000, although not all of it is used.
   functions:
+    - name: AllocateNewRender3dElement
```
UsernameFodder, line 32: Per convention, just `AllocRender3dElement` would be better.

headers/types/common/graphics.h
```diff
@@ -71,4 +72,97 @@ struct animation_control {
...
+    undefined field48_0x38;
+    undefined field49_0x39;
+    undefined field50_0x3a;
+    undefined field51_0x3b;
+    undefined field52_0x3c;
+    undefined field53_0x3d;
+    undefined field54_0x3e;
+    undefined field55_0x3f;
+    struct render_3d_element* render_stack[128];
+};
```
UsernameFodder, line 166: Needs an `ASSERT_SIZE` after it.

symbols/arm9.yml
```diff
@@ -3144,6 +3144,37 @@ arm9:
...
+        
+        (args same as SetAndPlayAnimationForAnimationControl)
+        r0: animation_control
+        r1: animation key (either an animation or animation group depending on the type of sprite and if it does have animation group with this animation key as index)
+        r2: direction_id (unsure) (the key to the wan_animation in itself, only used when animation key represent a wan_animation_group)
+        r3: ?
+        stack[0]: low_palette_pos
+        stack[1] (0x4): ?
+        stack[2] (0x8): ?
+        stack[3] (0xC): ?
```
UsernameFodder, line 3177: This last parameter isn't given in the function header you defined. It should either be removed here, or added to the function header.

symbols/arm9.yml
```diff
@@ -3153,6 +3184,20 @@ arm9:
...
+        Set the animation to play with the animation control, and start it.
+        
+        r0: animation_control
+        r1: animation key (either an animation or animation group depending on the type of sprite and if it does have animation group with this animation key as index)
+        r2: direction_id (unsure) (the key to the wan_animation in itself, only used when animation key represent a wan_animation_group)
+        r3: ?
+        stack[0]: low_palette_pos
+        stack[1] (0x4): ?
+        stack[2] (0x8): ?
+        stack[3] (0xC): ?
```
UsernameFodder, line 3200: This last parameter isn't given in the function header you defined. It should either be removed here, or added to the function header.

UsernameFodder (changes requested):

headers/types/files/wan.h
marius851000: Totally possible. It’s just that I thought it would avoid breaking potential (certainly non-existant) code that depend on unk4. I’ll do that if you think it’s better (I’m a bit undecided myself).

marius851000:
So, I’ll need help for one things, you who seems to be more familiar with C than me:
- Line 172 of graphics.h (render_stack definition), how do I define it to be a pointer to an array instead of an array of pointer

(originally was two, but fixed one myself)

That aside, the other issues should be fixed.

Also, I think I’ll focus more on the sky decomp now that AnonymousRandomPerson started one.

UsernameFodder:
> So, I’ll need help for one things, you who seems to be more familiar with C than me:
> 
>     * Line 172 of graphics.h (render_stack definition), how do I define it to be a pointer to an array instead of an array of pointer

It would be `struct render_3d_element (*render_stack)[128]`. Though since the syntax is confusing, for clarity I usually prefer to define an equivalent auxiliary struct for the array, e.g.,

```c
struct render_stack {
    struct render_3d_element stack[128];
};

struct render_3d_global {
    ...
    struct render_stack* render_stack;
};
```

I'll go ahead and make this change.

> Also, I think I’ll focus more on the sky decomp now that AnonymousRandomPerson started one.

That's cool, my understanding is that the we're going to try to keep the symbols somewhat in sync between these two projects, so contribute to whichever one you prefer.

UsernameFodder:
> > So, I’ll need help for one things, you who seems to be more familiar with C than me:
> > ```
> > * Line 172 of graphics.h (render_stack definition), how do I define it to be a pointer to an array instead of an array of pointer
> > ```
> 
> It would be `struct render_3d_element (*render_stack)[128]`. Though since the syntax is confusing, for clarity I usually prefer to define an equivalent auxiliary struct for the array, e.g.,
> 
> ```c
> struct render_stack {
>     struct render_3d_element stack[128];
> };
> 
> struct render_3d_global {
>     ...
>     struct render_stack* render_stack;
> };
> ```
> 
> I'll go ahead and make this change.

Actually, since this array appears to be heap-allocated and the code technically supports different lengths via `max_index`, I think a bare pointer is actually better here. I'll keep it as just a pointer.

UsernameFodder (approved):
Thanks!

PR #177, opened 2023-09-03
Various JP symbols (ov18 - ov34)

Adex-8x:
Adds various JP symbols in some overlays (ranging from 18 to 34). Related to https://github.com/SkyTemple/skytemple-files/issues/235.


UsernameFodder (approved):
Looks good, thanks for finding these!

PR #179, opened 2023-09-04
Various JP symbols (arm9, ram, ov00 - ov17)

Adex-8x:
Adds various JP symbols in arm9, ram, and some overlays (ranging from 00 to 17; commit name is wrong, oops). Related to https://github.com/SkyTemple/skytemple-files/issues/235.


UsernameFodder (approved):
Looks good, thanks.

PR #175, opened 2023-08-30
Addition of Symbol in JP version

Laioxy:
Each symbol confirmed in JP version is added.

Also, the description of UNOWN_STONE_DROP_CHANCE was different from the actual value. (21% -> 33%)
The binary "0x21" was probably mistaken for "21%".

---
Nice to meet you. I am a Japanese who is analyzing PMD2 Sky.
I sent a pull request for the first time to contribute to the JP ROM support of SkyTemple.
For now, I have added only the parts that I can confirm, filling in the holes. Please check it out.

(My English is terrible, so I use DeepL translation. Sorry if the text is incomplete)

PR #181, opened 2023-09-04
Addition of JP symbols for overlay31

Laioxy:
All symbols in overlay31 that are written now are covered.


Laioxy:
I am not familiar with Git and GitHub, is this correct?
It appears to contain previous commits

UsernameFodder:
> I am not familiar with Git and GitHub, is this correct?
> It appears to contain previous commits

It will be okay in this case. All the new changes in the source branch (Laioxy:master) are compatible with the ones in the target branch (UsernameFodder:master), so Git can merge them correctly.

GitHub decides what commits to show in a pull request by comparing the source branch and target branch, finding the most recent shared commit, and showing all subsequent commits. Here, the most recent shared commit is the one just before your previous pull request (https://github.com/UsernameFodder/pmdsky-debug/pull/175). There were multiple commits on your source branch, but it was merged into the target branch as one combined commit (a0aa596). Now GitHub thinks the old commits on your branch (starting with 8fdca91) are new, so it shows them all in this pull request.

A common way to avoid this problem is to make a new Git branch for every pull request, and make changes there. That way, the master branch be kept as an exact copy of the base repository. [Here's a tutorial on how to do this, in Japanese](https://github.com/firstcontributions/first-contributions/blob/main/translations/README.ja.md).

UsernameFodder (approved):
Thanks!

Laioxy:
Oh, that's how it is. I'm glad to hear it's not a problem. Thanks for explaining!

PR #186, opened 2023-09-07
Add version-dependent structs/enums in the C headers

### LINKED ISSUE #182: Support per-version C headers
**Is your feature request related to a problem? Please describe.**
@Adex-8x found that `struct monster` in the JP version doesn't have all the fields that exist in NA/EU. JP is missing:

- `in_action`
- `field_0x64`
- `field_0x65`
- `field_0x66`

With version-specific enums like `enum script_level_id*` and `enum script_object_id*`, the headers just define version-specific enum tags, e.g., `enum script_level_id_eu`. This is already somewhat unideal, and with `struct monster`, this approach will be cumbersome since `struct monster` is itself embedded in other structs like `struct dungeon`, so version differences would be "contagious".

We should figure out a good way to represent version differences that doesn't make the headers harder to use (e.g., importing in Ghidra).

**Describe the solution you'd like**
It would probably work to use `#define VERSION` in tandem with `#if` or something, and deal with conditional struct fields at the preprocessor level. This would require `#define`'ing the version before importing the headers. We could ship wrappers around the main `pmdsky.h` that define the variable for different game versions, and have people pick the one they want when importing. That way, you could still just import a single header into Ghidra.

By default, `pmdsky.h` should set the version to NA or EU if not defined, for compatibility.

Right now the headers shipped aren't version-specific, so in the release package, we would probably make separate copies of `pmdsky.h` and place them next to the already version-specific symbols.

UsernameFodder:
Closes #182.

- Add a mechanism for defining/using version-dependent C headers
- Make `script_level_id` and `script_object_id` version-dependent
- Make `struct monster` version-dependent

The default version of `pmdsky.h` is NA if unspecified.


headers/types/dungeon_mode/dungeon_mode.h
```diff
@@ -193,9 +193,12 @@ struct statuses {
     // prevent a recently recruited ally from evolving after and or to add a monster to the
     // assembly after the completion of a dungeon?
     undefined field_0x59;
+#if PMDSKY_VERSION == PMDSKY_JP
```
UsernameFodder, line 196: @Adex-8x Is the new struct definition here in line with your findings?

Adex-8x: Yeah, that lines up with what I found.

headers/pmdsky.h
```diff
@@ -1,6 +1,13 @@
 #ifndef HEADERS_PMDSKY_H_
 #define HEADERS_PMDSKY_H_
 
+#include "versions.h"
+
+// Default to NA if there's no predefined version
+#ifndef PMDSKY_VERSION
+#define PMDSKY_VERSION PMDSKY_NA
```
End45, line 8: What are the implications of choosing one default version or another? Does it matter for anything?

End45 (commented):
I can't think of a better way of implementing this. Using the preprocessor seems like the best choice.

headers/pmdsky.h
UsernameFodder: I think the only implications should be small, and the choice is more-or-less arbitrary (NA has been my "default" for this repo for various things, and NA/EU are mostly the same right now besides the two script enums):
- someone wasn't reading the setup instructions properly and imports `pmdsky.h` instead of `pmdsky_*.h`; at least this way the imported headers are always valid by some measure
- if you don't care about the version-specific stuff and are lazy or the code is older; one of my other projects imports `pmdsky.h` and I don't feel super inclined to change it

Ideally anyone using the headers for something sensitive to version (Ghidra, c-of-time, etc.) would always use the appropriate version-aware headers.

End45 (approved):

PR #187, opened 2023-09-07
Synced function symbols from decomp

AnonymousRandomPerson:
- Core functions that are standard across most DS/NitroSDK games (e.g., crt0 init functions).
- Veneers auto-detected by Ghidra while setting up the decomp.

### DIFF
--- a/headers/functions/arm9.h
+++ b/headers/functions/arm9.h
@@ -3,7 +3,15 @@
 
 #include "arm9/itcm.h"
 
-void EntryArm9(void);
+void SvcWaitByLoop(void);
+void SvcSoftReset(void);
+void SvcCpuSet(void);
+void _start(void);
+void MIiUncompressBackward(void);
+void do_autoload(void);
+void StartAutoloadDoneCallback(void);
+void OSiReferSymbol(void);
+void NitroMain(void);
 void InitMemAllocTable(void);
 void SetMemAllocatorParams(get_alloc_arena_fn_t get_alloc_arena,
                            get_free_arena_fn_t get_free_arena);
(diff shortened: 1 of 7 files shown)


headers/functions/arm9.h
```diff
@@ -3,7 +3,15 @@
...
-void EntryArm9(void);
+void SvcWaitByLoop(void);
+void SvcSoftReset(void);
+void SvcCpuSet(void);
+void _start(void);
+void MIiUncompressBackward(void);
+void do_autoload(void);
+void StartAutoloadDoneCallback(void);
+void OSiReferSymbol(void);
+void NitroMain(void);
```
UsernameFodder, line 14: Do you know if these are (mostly) the only core functions with weird naming? It seems like after these you get into the allocator functions. If these are the only exceptions maybe it might make more sense to implement an exception list for the case checker so we can use the real names.

symbols/arm9.yml
```diff
@@ -38,6 +50,26 @@ arm9:
         The entrypoint for the ARM9 CPU. This is like the "main" function for the ARM9 subsystem.
         
         No params.
+    - name: MIiUncompressBackward
+      address:
+        NA: 0x2000970
+      description: Startup routine in crt0.
```
UsernameFodder, line 56: What is crt0? Can we add more details in the description?

UsernameFodder (changes requested):

symbols/arm9.yml
AnonymousRandomPerson: It's a standard set of startup routines for C programs. https://en.wikipedia.org/wiki/Crt0

I can link this in the descriptions.

headers/functions/arm9.h
AnonymousRandomPerson: These are not the only ones. I've seen this naming convention in some of the library code, and the number of functions would make an exception list grow very large.

I don't personally like this naming convention of mixing case with <domain>_<action>, but it is a real naming convention used by some C programming circles (e.g., https://github.com/pret/pokeheartgold/issues/213). It may be worth supporting this style outright, rather than making exceptions.

UsernameFodder (approved):
LGTM

PR #169, opened 2023-08-03
Info dump 55/66: Dungeon music

End45:
Adds some functions related to dungeon music and playing audio. The bulk of my notes about the audio system is yet to be dumped though (section 61: Sky Jukebox & Music).

I also updated the functions that refer to debug flags to be more accurate.

PR #189, opened 2023-09-15
Added debug functions from decomp

AnonymousRandomPerson:
-Decomped by @marius851000 .


symbols/arm9.yml
```diff
@@ -994,7 +994,19 @@ arm9:
         r0: [output] pointer to an IO struct {ptr, len}
         r1: file path string pointer
         r2: flags
-    - name: GetDebugFlag
+    - name: InitDebug
+      address:
+        NA: 0x200C0D4
+      description: |-
+        Would have initialized debugging-related things, if they were not removed.
+        As for the release version, does nothing but set DEBUG_IS_INITIALIZED to true.
```
UsernameFodder, line 1002: Is there supposed to be a global with this name?

headers/functions/arm9.h
```diff
@@ -98,14 +98,23 @@ void FileSeek(struct file_stream* file, int offset, int whence);
...
+void SetDebugFlag1(enum debug_flag flag, bool val);
+void InitDebugStripped6(void);
 int AppendProgPos(char* str, struct prog_pos_info* prog_pos, const char* msg);
+void InitDebugStripped5(void);
 void DebugPrintTrace(const char* msg, struct prog_pos_info* prog_pos);
 void DebugPrint0(const char* fmt, ...);
-bool GetDebugLogFlag(enum debug_log_flag flag);
-void SetDebugLogFlag(enum debug_log_flag flag, bool val);
+void InitDebugFlag2(void);
+bool GetDebugFlag2(enum debug_log_flag flag);
```
UsernameFodder, line 111: `enum debug_log_flag` ought to be changed to match the new function name, e.g., `debug_flag_2`

UsernameFodder (changes requested):

symbols/arm9.yml
AnonymousRandomPerson: The sync script doesn't support globals yet. Adding it manually.

headers/functions/arm9.h
AnonymousRandomPerson: Renamed. Also renamed `debug_flag` to `debug_flag_1`.

symbols/arm9.yml
```diff
@@ -1060,7 +1080,13 @@ arm9:
...
         r0: format
         ...: variadic
-    - name: GetDebugLogFlag
+    - name: InitDebugFlag2
+      address:
+        NA: 0x200C230
+      description: |-
+        Would have initialised the second set of debug flags.
+        Does nothing in release binary.
+    - name: GetDebugFlag2
```
UsernameFodder, line 1089: Hmm @End45 any thoughts on this name change? I'm just now remembering that `GetDebugFlag2` was the old name, before we renamed it in https://github.com/UsernameFodder/pmdsky-debug/pull/169. So this would actually be a reversion. @marius851000 do you think `DebugFlag2` is a more appropriate name than `DebugLogFlag`?

symbols/arm9.yml
```diff
@@ -10155,11 +10197,18 @@ arm9:
         - 0x3 (overlay 35)
         
         type: enum overlay_group_id
+    - name: DEBUG_IS_INITIALIZED
+      address:
+        NA: 0x20AF698
+      length:
+        NA: 0x4
```
UsernameFodder, line 10204: `bool` should be 1 byte long?

symbols/arm9.yml
End45: I think this would be a regression. The flags were renamed because now we know what each set of flags is used for, I think it's better to keep it that way.
Maybe Marius was using the old names when decompiling the functions?

AnonymousRandomPerson: I'll rename the flag back to `debug_log_flag` unless there are any objections.

UsernameFodder (approved):
Looks good now

symbols/arm9.yml
```diff
@@ -1091,6 +1133,22 @@ arm9:
...
+        NA: 0x200C250
+      description: "Does nothing, only called in the debug initialisation function."
+    - name: InitDebugStripped2
+      address:
+        NA: 0x200C254
+      description: "Does nothing, only called in the debug initialisation function."
+    - name: InitDebugStripped1
+      address:
+        NA: 0x200C258
+      description: "Does nothing, only called in the debug initialisation function."
```
End45, line 1151: Maybe these descriptions should be changed to "initiali**z**ation" for consistency?

AnonymousRandomPerson: I wasn't sure if we were enforcing a style of British/American spelling. If there is a preferred style then I can make a follow-up PR to fix this.

End45: I don't think there's a spelling style, but I did a quick search and most of the repo uses initialize. Might want to wait and see what @UsernameFodder has to say, but I'm guessing it's a good idea to keep it consistent, specially since `DEBUG_IS_INITIALIZED` was just introduced in this same PR.

AnonymousRandomPerson: I've made a PR for this: https://github.com/UsernameFodder/pmdsky-debug/pull/190

UsernameFodder: I don't really enforce it strictly as a matter of respect for the European researchers, but if given the choice I do generally prefer American spelling for consistency

symbols/arm9.yml
marius851000: Yes. I used the old name. Sorry for that, my fault.

PR #191, opened 2023-09-15
Added TEAM_MEMBER_TABLE for JP version

Laioxy:
Add the address of the symbols you discovered while researching about team members in the JP version.


UsernameFodder (approved):
Thanks!

PR #194, opened 2023-09-17
Synced strings symbols from decomp

AnonymousRandomPerson:


headers/data/arm9.h
```diff
@@ -17,6 +17,21 @@ extern uint32_t MONSTER_ID_LIMIT;
...
+extern char STRING_DEBUG_NEWLINE[1];
+extern char STRING_DEBUG_LOG_NULL[7];
+extern char STRING_DEBUG_STRING_NEWLINE[3];
+extern char STRING_EFFECT_EFFECT_BIN[17];
+extern char STRING_MONSTER_MONSTER_BIN[19];
+extern char STRING_BALANCE_M_LEVEL_BIN[19];
+extern char STRING_DUNGEON_DUNGEON_BIN[19];
+extern char STRING_MONSTER_M_ATTACK_BIN[20];
+extern char STRING_MONSTER_M_GROUND_BIN[20];
+extern char STRING_FILE_DIRECTORY_INIT[36];
```
UsernameFodder, line 34: These buffers should include the space needed for the null terminator

symbols/arm9.yml
```diff
@@ -8651,7 +8651,7 @@ arm9:
         r1: divisor
         return: (quotient) | (remainder << 32)
   data:
-    - name: ARM9_HEADER
+    - name: SECURE
```
UsernameFodder, line 8654: Can we name this more specifically? Something like `SECURE_AREA_HEADER`?

UsernameFodder (changes requested):

symbols/arm9.yml
UsernameFodder: Oh I guess this comes from `_secure`, which is some standardized name? If so we can leave it I guess.

symbols/arm9.yml
```diff
@@ -10457,7 +10532,7 @@ arm9:
         NA: 0x20AF698
       length:
         NA: 0x1
-    - name: PACK_FILE_OPENED
+    - name: PACK_FILES_OPENED
```
UsernameFodder, line 10535: Also need to update the description of `OpenAllPackFiles`

symbols/arm9.yml
```diff
@@ -8670,7 +8670,7 @@ arm9:
         EU: 0xCC
         NA: 0xCC
         JP: 0xCC
-      description: "Note: unverified, ported from Irdkwia's notes"
+      description: Parameters used by the NitroSDK to read the ROM.
```
UsernameFodder, line 8673: It seems like we know the meanings of all the fields here, can we make a struct for it?

symbols/arm9.yml
AnonymousRandomPerson: Yes, I changed the case convention but left it as is otherwise.

symbols/arm9.yml
AnonymousRandomPerson: I don't know enough about the struct to provide descriptions, but I can at least put the var names there.

headers/data/arm9.h
UsernameFodder: I wasn't thinking carefully, sorry. I think these ought to match the underlying data symbol length, which I assume includes not just the null terminator but also any padding in the binary. I updated the lengths, lmk if you object.

UsernameFodder (approved):
LGTM, assuming there's no objections to the buffer length changes

headers/data/arm9.h
AnonymousRandomPerson: Fine by me.

PR #195, opened 2023-09-19
Correct CART_REMOVED_IMG_DATA NA size

Adex-8x:
Double-checked `CART_REMOVED_IMG_DATA`'s size in the NA ROM; it should be the same size as its EU and JP counterparts.


UsernameFodder (approved):
Thanks!

PR #198, opened 2023-10-03
Bag + Storage Documentation

Jawshoeuh:
Added a bunch of functions related to storage and the bag with a few outliers (GetExclusiveItemForMonsterFromBag and GetHpBoostFromExclusiveItems).

If you were wondering why I didn't add a new struct or use a bulk item for storage... for some reason, the items in storage are stored in two arrays like so. I don't really know why it doesn't use bulk_items because those take up the same space. So item 0 has it's id held at item_id[0] and the number at item_num[0]
item_id[1000];
item_num[1000];

Also the game converts the storage items into struct items, calls the item list sorting function, and then turns the sorted list of items back into storage items. It discards the stuct item array and doesn't return it or store it using an input. Noting it just in case anyone also had questions about it.


symbols/arm9.yml
```diff
@@ -2243,30 +2354,102 @@ arm9:
...
+        r0: item index
+        r1: [output] pointer to an item
+        return: bool whether or not the item id is not 0
+    - name: RemoveItemAtIdxInStorage
+      address:
+        NA: 0x20101A0
+      description: |-
+        Remove an item at the specified index from storage.
+        
+        r0: storage item idx
```
UsernameFodder, line 2397: The prototype says this returns a bool?

symbols/arm9.yml
```diff
@@ -2243,30 +2354,102 @@ arm9:
...
+        return: bool whether an item was removed
+    - name: RemoveItemInStorage
+      address:
+        NA: 0x2010260
+      description: |-
+        Removes a storage item equivalent to the item passed from storage.
+        
+        r0: pointer to an item
+        return: bool whether an item was removed
+    - name: StorageZInit
```
UsernameFodder, line 2417: Can you add a prototype?

symbols/arm9.yml
```diff
@@ -2243,30 +2354,102 @@ arm9:
...
         NA: 0x201031C
         JP: 0x20102EC
       description: |-
+        Attempts to add the bulk_item to storage.
         Implements SPECIAL_PROC_ADD_ITEM_TO_STORAGE (see ScriptSpecialProcessCall).
         
-        r0: pointer to an owned_item
-        return: bool
+        r0: pointer to a bulk_item
+        return: bool whether an item was removed
```
UsernameFodder, line 2434: Should this be:

```suggestion
        return: bool whether an item was added
```
?

symbols/arm9.yml
```diff
@@ -2243,30 +2354,102 @@ arm9:
...
+    - name: AddItemToStorage
+      address:
+        NA: 0x20103AC
+      description: |-
+        Attempts to add the item to storage.
+        
+        r0: pointer to an item
+        return: bool whether an item was removed
```
UsernameFodder, line 2442: Should this be:

```suggestion
        return: bool whether an item was added
```
?

UsernameFodder (changes requested):

UsernameFodder (approved):
Thanks!

Issue #164, opened 2023-07-14
Undocumented struct monster_summary


Jawshoeuh:
**Existing information**
The function 0x22F8CC (NA) appears to use an undocumented struct that is essentially a summary or snapshot of the monster's current condition. From what I can tell it is used when you open up the summary page and exist in the dungeon struct at +0x2CA7C when getting out of a dungeon. Because the struct is undocumented, I wanted to double check with others that my observations were correct before pushing to the repo. Also want to ensure that the struct and its contents are reasonably named and I didn't miss anything obvious during my analysis.

The function begins by copying basic information (hp, exp, species, etc) into the struct and then making a list of enum_status_id's (up to 0x1D, the 0x1Eth status is set to 0 as a kind of terminator). 

```
// Please ignore minor mistakes, they will be corrected before pushing to the repo.
struct monster_summary{
    struct monster_id_16 id; // 0x0
    char  monster_name[10]; // 0x2: Exact size is a guess
    undefined field_0xD;
    undefined field_0xE;
    undefined field_0xF;
    undefined field_0x10;
    undefined field_0x11;
    undefined field_0x12;
    undefined field_0x13;
    undefined field_0x14;
    undefined field_0x15;
    struct type_id_8 types[2]; // 0x16
    struct ability_id_8 abilities[2]; // 0x18
    struct dungeon_id_8 joined_at; // 0x1A
    uint8_t joined_at_floor; // 0x1B
    struct item held_item; // 0x1C
    int32_t hp; // 0x24:
    int32_t max_hp; // 0x28: Actual max HP
    uint32_t level; // 0x2C
    int exp; // 0x30
    uint8_t offensive_stats[2]; // 0x34: {atk, sp_atk}
    uint8_t defensive_stats[2]; // 0x36: {def, sp_def}
    bool is_team_leader; // 0x38
    uint8_t attack_boost; // 0x39: from things like Power Band, Munch Belt
    uint8_t special_attack_boost; // 0x3A
    uint8_t defense_boost; // 0x3B
    uint8_t special_defense_boost; // 0x3C
    undefined field_0x3D;
    int16_t iq; // 0x3E
    undefined field_0x40;
    undefined field_0x41;
    undefined field_0x42; // Set to 0.
    undefined field_0x43; // Set to 0.
    // 0x44: Appears to be some kind of evolution status related number?
    // If DUNGEON_PTR + 0x75A (unknown) is 0, it gets set to 3.
[...]

PR #199, opened 2023-10-04
New struct monster_summary documentation

Jawshoeuh:
Added the struct mentioned in https://github.com/UsernameFodder/pmdsky-debug/issues/164

Taking better naming conventions for the functions because I'm uncertain if they're good.


headers/types/dungeon_mode/dungeon_mode_common.h
```diff
@@ -62,4 +62,61 @@ struct move {
...
+    struct tactic_id_8 tactic;  // 0x54
+    undefined field_0x55;
+    undefined field_0x56;
+    undefined field_0x57;
+    // 0x58: Appears to be a list of all the currently inflicted statues in their enum form. The
+    // last entry (30th) appears to always be STATUS_NONE to serve as a terminator for the list.
+    // While in ground mode, it's always filled with STATUS_NONE.
+    struct status_id_8 active_statuses[30];
+};
+ASSERT_SIZE(struct monster_summary, 118);
```
UsernameFodder, line 120: As is there's two bytes of implicit padding at the end of the struct. Do you have confirmation (e.g., from a `MemAlloc` call or from the way they're laid out contiguously in arrays) that the struct size really is 118, as opposed to 120?

UsernameFodder (commented):
Note: I fixed some formatting issues, and defined `struct status_id_8`, and moved `struct monster_summary` to `dungeon_mode_common.h` where I think it fits better (and has access to `status_id_8`). The only remaining issue is with the sizing of `struct monster_summary`.

headers/types/dungeon_mode/dungeon_mode_common.h
Jawshoeuh: You are correct. The function that call CreateMonsterSummaryFromMonster uses 0x78 (120) on the stack.

That size also perfectly lines up with the empty space on the dungeon struct where the struct exists. (Seems like I forgot to update the dungeon struct. I'll add it on there as well.)

UsernameFodder (approved):
Thanks!

PR #201, opened 2023-10-27
Some arm7 interrupt documentation

adakite1:
I was working on some arm7 reversing and found some mirrored code between arm9 and arm7 where `EnableIrqFlag` and `SetIrqFlag` were being used, and from there I was able to find a few more arm9 interrupt functions in the arm7 as well. Interestingly although all of these ones are identical, the last interrupt function `GetIrqFlag` is not, and it does an `and` with a different mask, so I didn't include it here.

Also a note for all future arm7 documentation: I did a checksum on the `arm7.bin` of all three editions of EoS and they are all identical (`SHA256 CB2A9B26D77A6E7D0ADD85C27E3642105918695D040D02F70D5EBC1D52C9495B`), and since they are also all loaded at `0x2380000`, addresses for all editions will be identical if the function is in the binary. There do appear to be some functions that seem to be written in during runtime though, so this might not apply for those ones.

### DIFF
--- a/CREDITS.txt
+++ b/CREDITS.txt
@@ -27,6 +27,7 @@ other parts of the project, feel free to add yourself to this file!
 - Jawshoeuh
 - Adex
 - Laioxy
+- Adakite
 - The rest of the contributors to Project Pokémon's technical documentation
   for the NDS Mystery Dungeon games, which can be found here:
   https://projectpokemon.org/home/docs/mystery-dungeon-nds/
--- a/headers/functions/arm7.h
+++ b/headers/functions/arm7.h
@@ -2,5 +2,10 @@
 #define HEADERS_FUNCTIONS_ARM7_H_
 
 void EntryArm7(void);
+int ClearIrqFlag(void);
+int EnableIrqFlag(void);
+int SetIrqFlag(int new_value);
+int EnableIrqFiqFlags(void);
+int SetIrqFiqFlags(int new_value);
 
 #endif
--- a/symbols/arm7.yml
+++ b/symbols/arm7.yml
@@ -27,4 +27,51 @@ arm7:
         The entrypoint for the ARM7 CPU. This is like the "main" function for the ARM7 subsystem.
         
         No params.
+    - name: ClearIrqFlag
+      address:
+        EU: 0x2385ED4
+        NA: 0x2385ED4
+        JP: 0x2385ED4
+      description: |-
+        Enables processor interrupts by clearing the i flag in the program status register (cpsr).
+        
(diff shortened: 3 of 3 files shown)


UsernameFodder:
> Interestingly although all of these ones are identical, the last interrupt function GetIrqFlag is not, and it does an and with a different mask, so I didn't include it here.

If we know what the function is, can we just add it with a different name (doesn't have to be this PR though)? I think it's better to have it somewhere than not.

UsernameFodder (approved):
Looks good to me, thanks! I'm not 100% sure if I like duplicating the function names (since `pmdsky.h` has one global namespace), but seeing as these are actually mirrored functions, I suppose it's okay. We can revisit this later if needed.

adakite1:
> > Interestingly although all of these ones are identical, the last interrupt function GetIrqFlag is not, and it does an and with a different mask, so I didn't include it here.
> 
> If we know what the function is, can we just add it with a different name (doesn't have to be this PR though)? I think it's better to have it somewhere than not.

We can do that but only once we know a bit about its purpose, otherwise I'm not even sure what its name should be lol xP
It does an and with a really weird bitmask.

End45:
Where is that function located?

adakite1:
Right after 02385f28h `SetIrqFiqFlags`, at 02385f40h. It looks similar to `GetIrqFlag`, but it gets all 5 lower bits of cpsr instead of the bit 7

adakite1:
> Bits 0 trhrough 4 contain the execution mode of the processor, so that's what the function seems to be returning. I'd call it `GetProcessorMode`. From the No$GBA docs:
> 
> ```
> The Mode Bits M4-M0 contain the current operating mode.
>   Binary Hex Dec  Expl.
>   0xx00b 00h 0  - Old User       ;\26bit Backward Compatibility modes
>   0xx01b 01h 1  - Old FIQ        ; (supported only on ARMv3, except ARMv3G,
>   0xx10b 02h 2  - Old IRQ        ; and on some non-T variants of ARMv4)
>   0xx11b 03h 3  - Old Supervisor ;/
>   10000b 10h 16 - User (non-privileged)
>   10001b 11h 17 - FIQ
>   10010b 12h 18 - IRQ
>   10011b 13h 19 - Supervisor (SWI)
>   10111b 17h 23 - Abort
>   11011b 1Bh 27 - Undefined
>   11111b 1Fh 31 - System (privileged 'User' mode) (ARMv4 and up)
> Writing any other values into the Mode bits is not allowed.
> ```

I'll do a pr. Should I include this table in the description for reference?

End45:
Seems a bit excessive, it's general ARM knowledge. Maybe you could define an enum with each state and use it as the return type, but it might not be worth it.

adakite1:
Sure, I'll leave it out.

PR #206, opened 2023-11-12
Further Arm7 documentation, as well as a note on FileRead limitations

adakite1:
Note: I'd appreciate advice on how I should structure this! I've looked at how the ITCM documentation is done which I guess would be the closest parallel to this but how the addresses are documented I don't quite understand.

Changed the arm7 addresses to reflect addresses after the memory map inside of the entry function as these addresses are the ones used by function calls and accessing constants.

I'm not quite sure what to name these areas, but for now I have them named as ENTRY, WRAM, and SLOW.
ENTRY is the very small chunk of code used solely for mapping the memory.
WRAM is where a lot of important functions like the irq functions, the stdlib math functions, and the arm7 main event loop reside. (Interestingly the copy seems to copy partially into the Shared WRAM instead of just into the ARM7 only WRAM)
SLOW is named as such since its the rest of arm7 that couldn't possibly fit into the WRAM and so is placed in the main memory instead.

I haven't had a chance to take a closer look at the 0x27E0000 area since most of the code I've had to work with for arm7 patching has been within the 0x37F8000 area, but given that 0xF608 is just under 64kB which is the amount of WRAM reserved specifically for arm7 (see https://problemkaputt.de/gbatek.htm#dsmemorymaps), and the 0x27E0000 area has a much larger size than the WRAM area, I think this is a good hypothesis for now.

There were also some small changes I've made:
- **[arm7]** The DSE audio engine is entirely on the arm9 and does IPC with arm7 only as a way to get the audio chip to kick in, so in this way, the arm7 just handles audio I/O with the mitsumi MM3205B sound chip, setting a few registers when called. Apparently Nintendo didn't allow anyone to touch arm7 and add stuff to it? I've changed the note on the binary description to reflect this.
- **[arm7]** While working on arm7 patches, at one point, I found a need for a __divsi3 function, and so after some fuzzing I've found __divsi3, __udivsi3, and __udivsi3_no_zero_check on arm7. They worked as intended in the patch, and appears to be identical to the arm9 version as well bar a few constant accesses. It might be possible to extrapolate some more standard library functions from here.
- **[arm9]** In early versions of adpcm streaming, I had the decoder on arm7 ipc arm9 for new ADPCM chunks since only arm9 can read nitrofs files, but when arm7 called arm9, the call would go through but after all the read parameters had been calculated and once it was time to actually read the next block from the rom, it would freeze on FileRead (this is with or without calling DataTransferInit and DataTransferStop). I was confused for a while, but noticed that the irq handler runs in the irq processor mode, not the usual SYSTEM mode where FileReads are generally done. Lo-and-behold I could get FileRead to work once again by forcing cpsr into SYSTEM mode. This corrupts states of course and crashes the game after a few seconds, but the file read itself worked. So I'm not entirely sure why, but FileRead appears to only work in some processor mode, so I've added a note on that.

### DIFF
--- a/headers/functions/arm7.h
+++ b/headers/functions/arm7.h
@@ -2,7 +2,7 @@
 #define HEADERS_FUNCTIONS_ARM7_H_
 
 void EntryArm7(void);
-void MainArm7(void);
+void NitroSpMain(void);
 int GetProcessorMode(void);
 
 #endif
--- a/headers/functions/arm7.h
+++ b/headers/functions/arm7.h
@@ -3,14 +3,6 @@
 
 void EntryArm7(void);
 void MainArm7(void);
-int ClearIrqFlag(void);
-int EnableIrqFlag(void);
-int SetIrqFlag(int new_value);
-int EnableIrqFiqFlags(void);
-int SetIrqFiqFlags(int new_value);
 int GetProcessorMode(void);
-unsigned long long __divsi3(int dividend, int divisor);
-unsigned long long __udivsi3(uint32_t dividend, uint32_t divisor);
-unsigned long long __udivsi3_no_zero_check(uint32_t dividend, uint32_t divisor);
 
 #endif
--- a/symbols/arm7.yml
+++ b/symbols/arm7.yml
@@ -1,34 +1,34 @@
 arm7:
   versions:
-    - EU-ENTRY
-    - NA-ENTRY
-    - JP-ENTRY
+    - EU
+    - NA
+    - JP
     - EU-WRAM
     - NA-WRAM
     - JP-WRAM
-    - EU-SLOW
-    - NA-SLOW
-    - JP-SLOW
+    - EU-RAM
+    - NA-RAM
+    - JP-RAM
   address:
-    EU-ENTRY: 0x2380000
-    NA-ENTRY: 0x2380000
-    JP-ENTRY: 0x2380000
+    EU: 0x2380000
+    NA: 0x2380000
+    JP: 0x2380000
     EU-WRAM: 0x37F8000
     NA-WRAM: 0x37F8000
     JP-WRAM: 0x37F8000
-    EU-SLOW: 0x27E0000
-    NA-SLOW: 0x27E0000
-    JP-SLOW: 0x27E0000
+    EU-RAM: 0x27E0000
+    NA-RAM: 0x27E0000
+    JP-RAM: 0x27E0000
   length:
-    EU-ENTRY: 0x1E8
-    NA-ENTRY: 0x1E8
-    JP-ENTRY: 0x1E8
+    EU: 0x27080    #0x1E8 long ENTRY block
+    NA: 0x27080
+    JP: 0x27080
     EU-WRAM: 0xF608
     NA-WRAM: 0xF608
     JP-WRAM: 0xF608
-    EU-SLOW: 0x17878
-    NA-SLOW: 0x17878
-    JP-SLOW: 0x17878
+    EU-RAM: 0x17878
+    NA-RAM: 0x17878
+    JP-RAM: 0x17878
   description: |-
     The ARM7 binary.
     
@@ -54,36 +54,48 @@ arm7:
         No params.
     - name: MainArm7
       address:
-        EU: 0x37F8000
-        NA: 0x37F8000
-        JP: 0x37F8000
+        EU: 0x23801E8
+        NA: 0x23801E8
+        JP: 0x23801E8
+        EU-WRAM: 0x37F8000
+        NA-WRAM: 0x37F8000
+        JP-WRAM: 0x37F8000
       description: |-
         This main function for the ARM7 subsystem. Contains the main event loop.
 
         No params.
     - name: ClearIrqFlag
       address:
-        EU: 0x37FDCEC
-        NA: 0x37FDCEC
-        JP: 0x37FDCEC
+        EU: 0x2385ED4
+        NA: 0x2385ED4
+        JP: 0x2385ED4
+        EU-WRAM: 0x37FDCEC
+        NA-WRAM: 0x37FDCEC
+        JP-WRAM: 0x37FDCEC
       description: |-
         Copy of the ARM9 function. See arm9.yml for more information.
         
         return: Old value of cpsr & 0x80 (0x80 if interrupts were disabled, 0x0 if they were already enabled)
     - name: EnableIrqFlag
       address:
-        EU: 0x37FDD00
-        NA: 0x37FDD00
-        JP: 0x37FDD00
+        EU: 0x2385EE8
+        NA: 0x2385EE8
+        JP: 0x2385EE8
+        EU-WRAM: 0x37FDD00
+        NA-WRAM: 0x37FDD00
+        JP-WRAM: 0x37FDD00
       description: |-
         Copy of the ARM9 function. See arm9.yml for more information.
         
         return: Old value of cpsr & 0x80 (0x80 if interrupts were already disabled, 0x0 if they were enabled)
     - name: SetIrqFlag
       address:
-        EU: 0x37FDD14
-        NA: 0x37FDD14
-        JP: 0x37FDD14
+        EU: 0x2385EFC
+        NA: 0x2385EFC
+        JP: 0x2385EFC
+        EU-WRAM: 0x37FDD14
+        NA-WRAM: 0x37FDD14
+        JP-WRAM: 0x37FDD14
       description: |-
         Copy of the ARM9 function. See arm9.yml for more information.
         
@@ -91,18 +103,24 @@ arm7:
         return: Old value of cpsr & 0x80 (0x80 if interrupts were disabled, 0x0 if they were enabled)
     - name: EnableIrqFiqFlags
       address:
-        EU: 0x37FDD2C
-        NA: 0x37FDD2C
-        JP: 0x37FDD2C
+        EU: 0x2385F14
+        NA: 0x2385F14
+        JP: 0x2385F14
+        EU-WRAM: 0x37FDD2C
+        NA-WRAM: 0x37FDD2C
+        JP-WRAM: 0x37FDD2C
       description: |-
         Copy of the ARM9 function. See arm9.yml for more information.
         
         return: Old value of cpsr & 0xC0 (contains the previous values of the i and f flags)
     - name: SetIrqFiqFlags
       address:
-        EU: 0x37FDD40
-        NA: 0x37FDD40
-        JP: 0x37FDD40
+        EU: 0x2385F28
+        NA: 0x2385F28
+        JP: 0x2385F28
+        EU-WRAM: 0x37FDD40
+        NA-WRAM: 0x37FDD40
+        JP-WRAM: 0x37FDD40
       description: |-
         Copy of the ARM9 function. See arm9.yml for more information.
         
@@ -110,9 +128,12 @@ arm7:
         return: Old value of cpsr & 0xC0 (contains the previous values of the i and f flags)
     - name: GetProcessorMode
       address:
-        EU: 0x37FDD58
-        NA: 0x37FDD58
-        JP: 0x37FDD58
+        EU: 0x2385F40
+        NA: 0x2385F40
+        JP: 0x2385F40
+        EU-WRAM: 0x37FDD58
+        NA-WRAM: 0x37FDD58
+        JP-WRAM: 0x37FDD58
       description: |-
         Gets the processor's current operating mode.
 
@@ -121,9 +142,9 @@ arm7:
         return: cpsr & 0x1f (the cpsr mode bits M4-M0)
     - name: __divsi3
       address:
-        EU: 0x3806BC8
-        NA: 0x3806BC8
-        JP: 0x3806BC8
+        EU-WRAM: 0x3806BC8
+        NA-WRAM: 0x3806BC8
+        JP-WRAM: 0x3806BC8
       description: |-
         Copy of the ARM9 function. See arm9.yml for more information.
         
@@ -132,9 +153,9 @@ arm7:
         return: (quotient) | (remainder << 32)
     - name: __udivsi3
       address:
-        EU: 0x3806DD4
-        NA: 0x3806DD4
-        JP: 0x3806DD4
+        EU-WRAM: 0x3806DD4
+        NA-WRAM: 0x3806DD4
+        JP-WRAM: 0x3806DD4
       description: |-
         Copy of the ARM9 function. See arm9.yml for more information.
         
@@ -143,9 +164,9 @@ arm7:
         return: (quotient) | (remainder << 32)
     - name: __udivsi3_no_zero_check
       address:
-        EU: 0x3806DDC
-        NA: 0x3806DDC
-        JP: 0x3806DDC
+        EU-WRAM: 0x3806DDC
+        NA-WRAM: 0x3806DDC
+        JP-WRAM: 0x3806DDC
       description: |-
         Copy of the ARM9 function. See arm9.yml for more information.
         
--- a/symbols/arm7.yml
+++ b/symbols/arm7.yml
@@ -1,8 +1,14 @@
 arm7:
   versions:
-    - EU
-    - NA
-    - JP
+    - EU-ENTRY
+    - NA-ENTRY
+    - JP-ENTRY
+    - EU-WRAM
+    - NA-WRAM
+    - JP-WRAM
+    - EU-SLOW
+    - NA-SLOW
+    - JP-SLOW
   address:
     EU-ENTRY: 0x2380000
     NA-ENTRY: 0x2380000
--- a/symbols/arm7.yml
+++ b/symbols/arm7.yml
@@ -35,12 +35,12 @@ arm7:
     This is the secondary binary that gets loaded when the game is launched.
     
     Speaking generally, this is the program run by the Nintendo DS's secondary ARM7TDMI CPU, which handles the audio I/O, the touch screen, Wi-Fi functions, cryptography, and more.
-
+    
     Memory map: (binary is initially loaded at 0x2380000)
     0x2380000-0x23801E8 => Contains EntryArm7 and two more methods, all related to memory mapping.
     0x23801E8-0x238F7F0 => Mapped to 0x37F8000, contains NitroSpMain and functions crucial to execution.
     0x238F7F0-0x23A7068 => Mapped to 0x27E0000, contains everything else that won't fit in the fast WRAM.
-
+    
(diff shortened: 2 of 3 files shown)


symbols/arm7.yml
```diff
@@ -1,88 +1,161 @@
...
-    - JP
+    - EU-ENTRY
+    - NA-ENTRY
+    - JP-ENTRY
+    - EU-WRAM
+    - NA-WRAM
+    - JP-WRAM
+    - EU-SLOW
+    - NA-SLOW
+    - JP-SLOW
```
UsernameFodder, line 11: Really not a fan of this. We should keep the untagged NA/EU/JP as-is, as those directly correspond to the binary itself, and its load address encoded in the cartridge header.

I think it would be a good idea to follow the same format we do for the ITCM (see `arm9/itcm.yml`), which is to record both the unmapped load address (NA/EU/JP), and the relevant mapped runtime address (-WRAM/-SLOW). Since the regions mapped to WRAM and main memory are each contiguous blocks, it would also be great if we could create subregions for them, like with the ITCM.

headers/functions/arm7.h
```diff
@@ -2,11 +2,15 @@
...
 
 void EntryArm7(void);
+void MainArm7(void);
 int ClearIrqFlag(void);
 int EnableIrqFlag(void);
 int SetIrqFlag(int new_value);
 int EnableIrqFiqFlags(void);
 int SetIrqFiqFlags(int new_value);
 int GetProcessorMode(void);
+unsigned long long __divsi3(int dividend, int divisor);
```
UsernameFodder, line 12: Hmm okay I don't think this copy-pasting from arm9 to arm7 is going to be sustainable. Can we delete the duplicate function prototypes from `arm7.h` and change the descriptions in `arm7.yml` to say something like:

```
Copy of the ARM9 function. See arm9.yml for more information.
```
?

UsernameFodder (changes requested):

symbols/arm7.yml
adakite1: So for ITCM the NA/EU/JP is unmapped initial load addr and NA-ITCM/EU-ITCM/JP-ITCM is mapped addr? Got it, I'll update it.

adakite1: Actually should I move the two later areas into separate files like `slow.yml` and `wram.yml`? I'm hesitant to do it this way because if we do this the main `arm7.yml` file is gonna be very small, the entry section only contains 3 functions solely responsible for memory mapping.

headers/functions/arm7.h
adakite1: Sure

symbols/arm7.yml
UsernameFodder: > So for ITCM the NA/EU/JP is unmapped initial load addr and NA-ITCM/EU-ITCM/JP-ITCM is mapped addr? Got it, I'll update it.

Yep, that's right. So I guess in this case there'd be _three_ different variants (e.g., NA, NA-WRAM, NA-SLOW).

(Side note, maybe we should change "-SLOW" to just "-RAM"? I think that better reflects what it actually is. We can add a description on the meanings of each region in the top-level description block)

> 

Actually should I move the two later areas into separate files like slow.yml and wram.yml? I'm hesitant to do it this way because if we do this the main arm7.yml file is gonna be very small, the entry section only contains 3 functions solely responsible for memory mapping. It would make more intuitive sense if the wram part was inside of the main file imo, and I would even say that it'd be nice to have the slow part with the wram part too, just tagged differently, but we could stick to what's done with itcm as well, that can be more straightforward in a way.

Technically we don't have to split anything out into separate files, I just think it's nice organizationally. It also means each region can have its own top-level description block (e.g., a `wram.yml` can have a description specially for ARM7 WRAM).

WRAM is fairly small, right? So is it safe to say that the majority of arm7 will reside in main memory? Maybe we could keep the entry stuff and main memory stuff in `arm7.yml`, and have a dedicated file for WRAM stuff? But if it would make more sense to keep everything in the same file for now, I think that's okay too. We can always change this later.

symbols/arm7.yml
```diff
@@ -1,88 +1,153 @@
...
-        EU: 0x2385F40
-        NA: 0x2385F40
-        JP: 0x2385F40
+        EU: 0x37FDD58
+        NA: 0x37FDD58
+        JP: 0x37FDD58
       description: |-
         Gets the processor's current operating mode.
         
-        See https://problemkaputt.de/gbatek.htm#armcpuflagsconditionfieldcond
```
UsernameFodder, line 85: Is this not applicable? Not sure why we're deleting it. It seems useful to have.

symbols/arm7.yml
adakite1: > (Side note, maybe we should change "-SLOW" to just "-RAM"? I think that better reflects what it actually is. We can add a description on the meanings of each region in the top-level description block)

Ah yeah that's a better name for it.

> WRAM is fairly small, right? So is it safe to say that the majority of arm7 will reside in main memory? Maybe we could keep the entry stuff and main memory stuff in arm7.yml, and have a dedicated file for WRAM stuff? But if it would make more sense to keep everything in the same file for now, I think that's okay too. We can always change this later.

It's true that it's only about 1/3 of the binary, but at the same time it contains some of the most important functions, like the main event loop, the interrupt setters and getters, and the ipc initialization and send/receive functions. In fact for the two arm7 patches I've done so far, in both instances I've never had to even touch the 0x27E0000 area, which makes me think it might be where functions that aren't used a ton are placed. I'll keep everything in one file for now.

symbols/arm7.yml
adakite1: Oh sorry I had to do a bunch of merge conflict resolutions lol, it must've been caught in the cross-fire.

symbols/arm7.yml
```diff
@@ -1,88 +1,153 @@
...
-    EU: 0x2380000
-    NA: 0x2380000
-    JP: 0x2380000
+    EU-ENTRY: 0x2380000
+    NA-ENTRY: 0x2380000
+    JP-ENTRY: 0x2380000
+    EU-WRAM: 0x37F8000
+    NA-WRAM: 0x37F8000
+    JP-WRAM: 0x37F8000
+    EU-SLOW: 0x27E0000
```
UsernameFodder, line 19: > I haven't had a chance to take a closer look at the 0x27E0000 area since most of the code I've had to work with for arm7 patching has been within the 0x37F8000 area, but given that 0xF608 is just under 64kB which is the amount of WRAM reserved specifically for arm7 (see https://problemkaputt.de/gbatek.htm#dsmemorymaps), and the 0x27E0000 area has a much larger size than the WRAM area, I think this is a good hypothesis for now.

Hmm, wait a second, I realized that this offset looked familiar. A long time ago I had a theory that this was the DTCM region, which I wrote down in my [EoS RAM notes](https://docs.google.com/document/d/1_Q_7BGmNx5wJtJ9iJEwlK1WITjiCcEQxE9C82RECJbg/edit#heading=h.jtqakz6si98g). IIRC, this was based on looking at raw memory in DeSmuME. What do you think of this? If you compare the bytes at 0x27E0000 to those at 0x238F7F0 (0xF7F0 relative to arm7.bin), do they really match for 0x17878 bytes? Or does the correspondence stop after 0x4000 bytes as I mentioned in my old notes?

adakite1: Interesting... let me check. Btw do you know how one can do a full memory dump in no$gba? It always says "Huh?" for me whenever I try to dump the whole thing.

AnonymousRandomPerson: I can confirm from the decomp that 0x27E0000 is the DTCM region.

symbols/arm7.yml
adakite1: > Yep, that's right. So I guess in this case there'd be three different variants (e.g., NA, NA-WRAM, NA-SLOW).

I think there's benefits to including NA-ENTRY as a separate variant as well since the area after those entry functions seem to be reused by the game for other things once the map is complete.

symbols/arm7.yml
adakite1: Interestingly, no$gba shows that 0x27E0000 contains the arm7 code I expect when I switch to nds7, but when I switch back to nds9 it shows me a different set of values, so I think the two cpus might have different memory mapped there. Although it's weird how it's the same exact address, does arm7 have something akin to DTCM? If this region is just slow memory I'm not sure why the devs picked the exact same place to load code.

symbols/arm7.yml
UsernameFodder: > I think there's benefits to including NA-ENTRY as a separate variant as well since the area after those entry functions seem to be reused by the game for other things once the map is complete.

This discussion about reuse came up back when we first added the ITCM stuff too. I think the conclusion was that we shouldn't worry about that until we discover what that memory is actually being reused for, and causing a conflict. The advantage of using the plain "NA", etc. is that it'll put all the artifacts for loading symbols in the main "NA", etc. bucket, rather than separate artifact files.

symbols/arm7.yml
UsernameFodder: > Btw do you know how one can do a full memory dump in no$gba? It always says "Huh?" for me whenever I try to dump the whole thing.

Not sure, I don't really use No$ all that much. @End45 would probably know?

> I can confirm from the decomp that 0x27E0000 is the DTCM region.

Thanks. Thinking about this more, it's kind of weird that arm7 is mapping stuff to 0x27E0000... The GBATEK memory map seems to indicate that the DTCM is ARM9-only. So it seems like arm7 should have no business touching it. Unless it's for some weird IPC stuff?

UsernameFodder: > Interestingly, no$gba shows that 0x27E0000 contains the arm7 code I expect when I switch to nds7, but when I switch back to nds9 it shows me a different set of values

Ah okay, I guess that refutes what I just said. So maybe we don't need to think about the DTCM in that case. If the code matches in arm7 mode, let's keep what you have.

adakite1: > Thanks. Thinking about this more, it's kind of weird that arm7 is mapping stuff to 0x27E0000... The GBATEK memory map seems to indicate that the DTCM is ARM9-only. So it seems like arm7 should have no business touching it. Unless it's for some weird IPC stuff?

It's still weird that it maps to 0x27E0000 in a way, but actually DTCM is separate from the 4MB of main memory right? This means that since arm9 maps 0x27E0000 to DTCM, that specific part of the main memory remains unused, which can be utilized by arm7 for placing its own code.

symbols/arm7.yml
adakite1: > This discussion about reuse came up back when we first added the ITCM stuff too. I think the conclusion was that we shouldn't worry about that until we discover what that memory is actually being reused for, and causing a conflict. The advantage of using the plain "NA", etc. is that it'll put all the artifacts for loading symbols in the main "NA", etc. bucket, rather than separate artifact files.

I see, I'll match what you guys did for ITCM for this then.

symbols/arm7.yml
UsernameFodder: > It's still weird that it maps to 0x27E0000 in a way, but actually DTCM is separate from the 4MB of main memory right? This means that since arm9 maps 0x27E0000 to DTCM, that specific part of the main memory remains unused, which can be utilized by arm7 for placing its own code.

But based on what you said earlier, it seems like arm9 and arm7 are allowed to have different views of what data is at a certain memory address (I guess they have their own memory, or there's some virtual memory thing going on?), so it seems like concerns about what memory is used shouldn't actually matter. So this still feels mysterious to me (though I don't think that should stop us from making the additions as we are here).

symbols/arm7.yml
```diff
@@ -3,55 +3,101 @@ arm7:
...
     JP: 0x2380000
+    EU-WRAM: 0x37F8000
+    NA-WRAM: 0x37F8000
+    JP-WRAM: 0x37F8000
+    EU-RAM: 0x27E0000
+    NA-RAM: 0x27E0000
+    JP-RAM: 0x27E0000
   length:
-    EU: 0x27080
+    EU: 0x27080    #0x1E8 long ENTRY block
```
UsernameFodder, line 23: Sorry, `resymgen` (the YAML transpiler this repo relies on) doesn't support YAML comments. Let's just put this in the binary description block.

symbols/arm7.yml
```diff
@@ -3,55 +3,101 @@ arm7:
...
       description: |-
-        The entrypoint for the ARM7 CPU. This is like the "main" function for the ARM7 subsystem.
+        The entrypoint for the ARM7 CPU.
         
+        Handles mapping the ARM7 binary into the various memory areas that the program will be using:
+
+        Memory map: (binary is initially loaded at 0x2380000)
+        0x2380000-0x23801E8 => Contains EntryArm7 and two more methods, all related to memory mapping.
+        0x23801E8-0x238F7F0 => Mapped to 0x37F8000, contains MainArm7 and functions crucial to execution.
+        0x238F7F0-0x23A7068 => Mapped to 0x27E0000, contains everything else that won't fit in the fast WRAM.
```
UsernameFodder, line 52: Can we move these mapping notes to the top-level binary description?

symbols/arm7.yml
adakite1: > though I don't think that should stop us from making the additions as we are here

Agreed, I just have a few more addresses that I need to calculate for unmapped.

> But based on what you said earlier, it seems like arm9 and arm7 are allowed to have different views of what data is at a certain memory address (I guess they have their own memory, or there's some virtual memory thing going on?), so it seems like concerns about what memory is used shouldn't actually matter.

I believe based on what I read [here](https://www.copetti.org/writings/consoles/nintendo-ds/) that the main memory is shared which is where I got this hypothesis, but nowhere does it really talk about the DTCM. Since it's tightly coupled, I almost think it might be hard-coded in the processor perhaps?

symbols/arm7.yml
```diff
@@ -3,55 +3,101 @@ arm7:
...
-    Speaking generally, this is the program run by the Nintendo DS's secondary ARM7TDMI CPU, which handles the audio engine, the touch screen, Wi-Fi functions, cryptography, and more.
+    Speaking generally, this is the program run by the Nintendo DS's secondary ARM7TDMI CPU, which handles the audio I/O, the touch screen, Wi-Fi functions, cryptography, and more.
   functions:
     - name: EntryArm7
       address:
         EU: 0x2380000
         NA: 0x2380000
         JP: 0x2380000
       description: |-
-        The entrypoint for the ARM7 CPU. This is like the "main" function for the ARM7 subsystem.
```
UsernameFodder, line 27: Can we mention how this function calls `MainArm7` after it sets everything up?

symbols/arm7.yml
adakite1: Added

symbols/arm7.yml
adakite1: Moved

symbols/arm7.yml
adakite1: > Can we mention how this function calls `MainArm7` after it sets everything up?

It's done the same way as how `NitroMain` is called for arm9 though with a constant being read containing the address to `MainArm7` and then `bx`, should I update the arm9 too?
(Side note, I didn't name it the same since I can't guarantee this is the same function from the NitroSDK. Another thing is that it would also create a name conflict for two functions with completely different contents.)

symbols/arm7.yml
```diff
@@ -79,10 +133,46 @@ arm7:
...
+      address:
+        EU-WRAM: 0x3806DDC
+        NA-WRAM: 0x3806DDC
+        JP-WRAM: 0x3806DDC
+      description: |-
+        Copy of the ARM9 function. See arm9.yml for more information.
+        
+        r0: dividend
+        r1: divisor
+        return: (quotient) | (remainder << 32)
```
UsernameFodder, line 177: It looks like these exist in the binary? Can we add the EU/NA/JP offsets?

symbols/arm7.yml
adakite1: Actually is `NitroMain` named as such because it matches an official NitroSDK leak? If not, since the call pattern is so similar we could also call this the NitroMain for arm7, afaik Nintendo didn't let anyone add code into arm7 so this is definitely Nintendo's main function at least.

symbols/arm7.yml
adakite1: Yeah I still have to calculate these since Ghidra only shows me WRAM addresses, hold on.

symbols/arm7.yml
UsernameFodder: > It's done the same way as how NitroMain is called for arm9 though with a constant being read containing the address to MainArm7 and then bx, should I update the arm9 too?

Hmm might as well, sounds helpful.

> (Side note, I didn't name it the same since I can't guarantee this is the same function from the NitroSDK. Another thing is that it would also create a name conflict for two functions with completely different contents.)

Yeah agree with the reasoning here.

symbols/arm7.yml
adakite1: Added

symbols/arm7.yml
AnonymousRandomPerson: I believe the NitroSDK name for `MainArm7` is `NitroSpMain`, if you want to follow the ARM9 `NitroMain`.

As for `EntryArm7`, that is required to be called `_start` in the actual codebase, just like the ARM9 entrypoint.

adakite1: > I believe the NitroSDK name for `MainArm7` is `NitroSpMain`, if you want to follow the ARM9 `NitroMain`.
> 
> As for `EntryArm7`, that is required to be called `_start` in the actual codebase, just like the ARM9 entrypoint.

I see, I will change `MainArm7` then. For `_start` though, the signature will match between arm7 and arm9 but the two entry functions contain completely different code, is it ok that we use the same name twice here?

adakite1: > Can we mention how this function calls MainArm7 after it sets everything up?
Added, and also on arm9 entry

UsernameFodder: I would just leave it as `EntryArm7`. We can change it later.

UsernameFodder (approved):
Everything looks good to me now. Neat stuff in this one, thanks!

Will leave it until tomorrow to merge in case there are lingering discussions people want to settle.

symbols/arm7.yml
End45: > Btw do you know how one can do a full memory dump in no$gba? It always says "Huh?" for me whenever I try to dump the whole thing.

It seems like the maximum dump size is 0x1000000, so you probably can't just dump everything at once, unless you manually move to the start of each region and perform separate dumps from there.

About the DTCM stuff: No idea why or how ARM7 uses it, but I know ARM9 holds the stack and the ARM9 interrupt vector there.

PR #207, opened 2023-11-13
Info dump 58-62/64: Music, interrupts, hurling and dungeon struct

End45:

- The Sky Jukebox and internal audio functions
- Hardware interrupts
- Item hurling
- Initialization of one of the fields on the dungeon struct


headers/data/overlay10.h
```diff
@@ -93,6 +93,7 @@ extern int16_t WONDER_CHEST_EXP_BOOST;
 extern int16_t SPAWN_CAP_WITH_MONSTER_HOUSE;
 extern int16_t POISON_DAMAGE_COOLDOWN;
 extern int16_t LEECH_SEED_DAMAGE_COOLDOWN;
+extern int16_t ITEM_HIT_CHANCE;
```
UsernameFodder, line 96: Can we name this `THROWN_ITEM_HIT_CHANCE`? Just "item" seems too generic.

symbols/arm7.yml
```diff
@@ -68,6 +68,63 @@ arm7:
...
+    - name: GetCurrentPlaybackTime
+      address:
+        EU-WRAM: 0x37FD21C
+        NA-WRAM: 0x37FD21C
+      description: |-
+        Returns the time that the current song has been playing for. Might have a more generic purpose.
+        
+        The time is obtained using a couple of RAM counters and the hardware timers for additional precision.
+        The game uses this value to know when a given note should stop being played. It doesn't seem to be used to keep track of the
+        current time instant within the song.
```
UsernameFodder, line 127: Needs a `return` line?

symbols/arm7.yml
```diff
@@ -68,6 +68,63 @@ arm7:
         This main function for the ARM7 subsystem. Contains the main event loop.
         
         No params.
+    - name: HardwareInterrupt
+      address:
+        EU-WRAM: 0x37FB488
```
UsernameFodder, line 73: Can we add the normal (initial load) EU and NA addresses for these too?

UsernameFodder (changes requested):

headers/data/overlay10.h
End45: :+1:

symbols/arm7.yml
End45: Right, thanks.

symbols/arm7.yml
End45: At first I didn't add them because calculating and verifying the offsets is quite tedious, but I guess I can just do a simple subtraction. It should be correct.

adakite1: Can we add JP addresses too? arm7 addresses should be identical across all three versions.

UsernameFodder: Derived them with `symbols_vfill.py` and they are the same. Added them in.

UsernameFodder (approved):
Thanks! This interrupt stuff is pretty cool.

PR #211, opened 2023-11-18
Updated special process enum.

Jawshoeuh:
When updating the symbols for the special processes, I forgot to update the enums. Additionally, I corrected two errors where I incorrectly said one of the functions had 'No params' (making it not match it's definition in the header file) and corrected SetActorTalkMainAndActorTalkSub incorrectly saying it to be SPECIAL_PROC_0x3D when that's SetActorTalkMain


UsernameFodder (approved):
Looks good, thanks!

PR #212, opened 2023-11-22
Filled in missing EU addresses

AnonymousRandomPerson:
With EU support in the decomp (https://github.com/pret/pmd-sky/pull/36), it is now possible to precisely identify the address of any NA/EU symbol when given an address in the other game version. I updated the [syncing script](https://github.com/pret/pmd-sky/blob/main/tools/sync_pmdsky_debug/sync_to_pmdsky_debug.py) within the decomp project to fill in any missing addresses between EU/NA (mostly EU). There were a handful of cases where EU and NA were pointing to different symbols, which I also ironed out.

For data lengths, I simply copied over the NA length to the EU length. It is not feasible to determine the length of a symbol automatically with just the ASM, and there are too many to inspect all of them manually. If the assumption of equal NA/EU lengths is undesirable, I can remove all of the newly added length entries instead.

I noticed that `GUMMI_IQ_STRING_IDS` and `GUMMI_LIKE_STRING_IDS` in overlay 29 are set to the same address. The data at this address has a size of 8, which leans towards `GUMMI_LIKE_STRING_IDS` being the correct entry. I'm not familiar enough with message IDs to quickly find the correct `GUMMI_IQ_STRING_IDS`, so I left both of them as duplicates for now when adding the EU address.


UsernameFodder:
> For data lengths, I simply copied over the NA length to the EU length. It is not feasible to determine the length of a symbol automatically with just the ASM, and there are too many to inspect all of them manually. If the assumption of equal NA/EU lengths is undesirable, I can remove all of the newly added length entries instead.

Yeah this is undesirable. It's not foolproof enough that I wouldn't want it to be automated.

> I noticed that GUMMI_IQ_STRING_IDS and GUMMI_LIKE_STRING_IDS in overlay 29 are set to the same address. The data at this address has a size of 8, which leans towards GUMMI_LIKE_STRING_IDS being the correct entry. I'm not familiar enough with message IDs to quickly find the correct GUMMI_IQ_STRING_IDS, so I left both of them as duplicates for now when adding the EU address.

Can you file a bug?

docs/contributing.md
```diff
@@ -1,5 +1,5 @@
 # Contributing to `pmdsky-debug`
-Contributions are welcome, no matter how small! This can be new symbols, new type definitions, changes to `resymgen`, bug fixes, documentation, etc. Make sure add yourself to [`CREDITS.txt`](../CREDITS.txt) if you're a new contributor!
+Contributions are welcome, no matter how small! This can be new symbols, new type definitions, changes to `resymgen`, bug fixes, documentation, etc. Make sure to add yourself to [`CREDITS.txt`](../CREDITS.txt) if you're a new contributor!
```
UsernameFodder, line 2: For the commit history, separate PR please

headers/functions/arm7.h
```diff
@@ -1,7 +1,7 @@
 #ifndef HEADERS_FUNCTIONS_ARM7_H_
 #define HEADERS_FUNCTIONS_ARM7_H_
 
-void EntryArm7(void);
+void _start(void);
```
UsernameFodder, line 4: Can we name this something else? Like `_start_arm7`? I don't want this to match the arm9 symbol when they aren't copies.

symbols/arm9.yml
```diff
@@ -29,16 +29,19 @@ arm9:
   subregions:
     - itcm.yml
   functions:
-    - name: SvcWaitByLoop
-      address:
-        NA: 0x2000088
-      description: Software interrupt.
     - name: SvcSoftReset
       address:
+        EU: 0x20001A4
```
UsernameFodder, line 34: Can you audit that these addresses are correct? While it's gratifying that my sorting algorithm's conflict resolution procedure is coming into play, it makes me suspicious that we have a genuine instance where the symbol order differs across versions. Especially in the system functions at the start, where I would expect things to be virtually identical.

symbols/arm9.yml
```diff
@@ -3688,16 +3695,25 @@ arm9:
...
+        NA: 0x2017EE8
+        JP: 0x2017F40
+      description: |-
+        Stops the BGM that is being currently played by calling SendAudioCommand.
+        
+        No params.
     - name: PlayBgmByIdVolume
       address:
-        EU: 0x2017F0C
+        EU: 0x201813C
```
UsernameFodder, line 3716: Can we pull the corrections into a separate PR for auditing?

This symbol was added by @End45 from an info dump, and if the sync thinks EU/NA don't agree, I would suspect that NA is wrong rather than EU, given that End's preferred version is EU. Does the sync script have any way to tell _which_ version is wrong if there's a conflict? If not, I think we ought to have such automated overrides behind some extra step (like a command line flag on the script) to make auditing easier.

UsernameFodder (changes requested):

symbols/arm9.yml
AnonymousRandomPerson: Yes, these are correct. All of these interrupt functions are mixed around between US and EU for some reason, and there are calls to them within the regular ARM9 code that resolve correctly for both versions with these specific addresses.

symbols/arm9.yml
AnonymousRandomPerson: The sync script does not resolve conflicts. If there are conflicts in pmdsky-debug, the sync will create duplicates within the decomp that result in compiler errors, and I then resolve the conflicts manually between the two projects.

I can favor the EU version here instead. The two functions are very similar, so it is possible that the US function was misidentified as matching the EU one.

headers/functions/arm7.h
AnonymousRandomPerson: In the actual codebase, it is required that this function is named `_start()` in both ARM9 and ARM7. Though I'm fine with renaming the ARM7 function here.

docs/contributing.md
AnonymousRandomPerson: https://github.com/UsernameFodder/pmdsky-debug/pull/213

AnonymousRandomPerson:
> Can you file a bug?

https://github.com/UsernameFodder/pmdsky-debug/issues/214

symbols/arm9.yml
AnonymousRandomPerson: https://github.com/UsernameFodder/pmdsky-debug/pull/215

AnonymousRandomPerson:
> Yeah this is undesirable. It's not foolproof enough that I wouldn't want it to be automated.

Removed.

symbols/arm9.yml
UsernameFodder: Weird

symbols/arm9.yml
UsernameFodder: > The sync script does not resolve conflicts. If there are conflicts in pmdsky-debug, the sync will create duplicates within the decomp that result in compiler errors, and I then resolve the conflicts manually between the two projects.

I see, that seems fine. As long as there's a forced manual step somewhere to resolve conflicts.

symbols/arm7.yml
```diff
@@ -56,6 +56,16 @@ arm7:
...
+    - name: do_autoload
+      address:
+        EU: 0x2380118
+        NA: 0x2380118
+        JP: 0x2380118
+    - name: StartAutoloadDoneCallback
+      address:
+        EU: 0x2380188
+        NA: 0x2380188
+        JP: 0x2380188
```
UsernameFodder, line 68: Are these also different than the ARM9 versions? If so, we should also tag these with an arm7 suffix in pmdsky-debug.

UsernameFodder (changes requested):

symbols/arm7.yml
AnonymousRandomPerson: Done.

UsernameFodder (approved):

PR #215, opened 2023-11-22
Fixed some incorrect addresses

AnonymousRandomPerson:
While syncing NA and EU with the decomp, I ran into a number of incorrect addresses that conflicted between versions. I manually resolved each of these, attempting to favor the version that the original contributor used.

### DIFF
--- a/symbols/arm9.yml
+++ b/symbols/arm9.yml
@@ -3657,7 +3657,7 @@ arm9:
     - name: PlayBgmByIdVeneer
       address:
         EU: 0x2017BF4
-        NA: 0x2017BD4
+        NA: 0x2017B58
       description: |-
         Likely a linker-generated veneer for PlayBgmById.
         
@@ -3667,7 +3667,7 @@ arm9:
     - name: PlayBgmByIdVolumeVeneer
       address:
         EU: 0x2017C00
-        NA: 0x2017BE0
+        NA: 0x2017B64
       description: |-
         Likely a linker-generated veneer for PlayBgmByIdVolume.
         
@@ -3688,7 +3688,7 @@ arm9:
     - name: PlayBgmById
       address:
         EU: 0x2017E90
-        NA: 0x2018024
+        NA: 0x2017DF4
       description: |-
         Initializes some values and then calls SendAudioCommand to play a BGM track.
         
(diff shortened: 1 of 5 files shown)


End45 (approved):
Manually checked everything except `SKY_ATTACK_DAMAGE_MULTIPLIER` and `CheckTeamMemberIdxVeneer`. The new offsets are correct.

Some of these errors are mine. I looks like they happned because of functions that are almost identical to the one that I was documenting, so I mistook them for the right one when adding the US offsets.
I need to find a better way of doing that.

UsernameFodder:
> Manually checked everything except SKY_ATTACK_DAMAGE_MULTIPLIER and CheckTeamMemberIdxVeneer. The new offsets are correct.

The new `SKY_ATTACK_DAMAGE_MULTIPLIER` is correct (I apparently had this right in my Ghidra project, but transcribed it wrong). `CheckTeamMemberIdxVeneer` points to the new `CheckTeamMemberIdx` as would be expected.

UsernameFodder (approved):
Thanks!

AnonymousRandomPerson:
> I looks like they happned because of functions that are almost identical to the one that I was documenting, so I mistook them for the right one when adding the US offsets.
I need to find a better way of doing that.

We can now use the decomp to match EU/US offsets more reliably than an ASM similarity check. For example, when knowing the EU address:
1. Find the placeholder symbol name within EU's [xMAP file](https://raw.githubusercontent.com/pret/pmd-sky/xmap/pmdskyeu.xMAP) by looking at the address and overlay.
2. Rename the symbol within the decomp.
3. Build the US ROM.
4. Find the symbol in the US xMAP file to get the US address.

I follow changes to pmdsky-debug and sync them to the decomp, so any conflicting addresses will be sorted out by this process as well.

End45:
> > I looks like they happned because of functions that are almost identical to the one that I was documenting, so I mistook them for the right one when adding the US offsets.
> > I need to find a better way of doing that.
> 
> We can now use the decomp to match EU/US offsets more reliably than an ASM similarity check. For example, when knowing the EU address:
> 
>     1. Find the placeholder symbol name within EU's [xMAP file](https://raw.githubusercontent.com/pret/pmd-sky/xmap/pmdskyeu.xMAP) by looking at the address and overlay.
> 
>     2. Rename the symbol within the decomp.
> 
>     3. Build the US ROM.
> 
>     4. Find the symbol in the US xMAP file to get the US address.
> 
> 
> I follow changes to pmdsky-debug and sync them to the decomp, so any conflicting addresses will be sorted out by this process as well.

I might use that if I ever set up the decomp project on my computer. Right now, it's still quicker to check manually.

PR #216, opened 2023-11-26
Info dump 63-65/65: Threading, map generation and misc.

End45:

- The game's threading system
- Map generation
- Other miscellaneous stuff that I hadn't categorized

And with this, the info dump is finally complete! Only took 1.5 years.

### DIFF
--- a/headers/functions/arm9.h
+++ b/headers/functions/arm9.h
@@ -970,8 +970,8 @@ void GeomMtxMult4x3(struct matrix_4x3* matrix);
 void GeomGxFifoSendMtx4x3(struct matrix_4x3* matrix, void* gxfifo);
 int IncrementThreadCount(void);
 void InsertThreadIntoList(struct thread* thread);
-void ThreadStart(struct thread* thread, void* function, struct thread* param_3, void* stack_area,
-                 int stack_size, int param_6);
+void ThreadStart(struct thread* thread, thread_entry_fn_t function, struct thread* param_3,
+                 void* stack_area, int stack_size, int param_6);
 void ThreadExit(void);
 void SetThreadField0xB4(struct thread* thread, int value);
 void InitThread(struct thread* thread, void* function, void* stack_area_minus_4);
--- a/headers/functions/arm9.h
+++ b/headers/functions/arm9.h
@@ -970,7 +970,7 @@ void GeomMtxMult4x3(struct matrix_4x3* matrix);
 void GeomGxFifoSendMtx4x3(struct matrix_4x3* matrix, void* gxfifo);
 int IncrementThreadCount(void);
 void InsertThreadIntoList(struct thread* thread);
-void ThreadStart(struct thread* thread, thread_entry_fn_t function, struct thread* param_3,
+void StartThread(struct thread* thread, thread_entry_fn_t function, struct thread* param_3,
                  void* stack_area, int stack_size, int param_6);
 void ThreadExit(void);
 void SetThreadField0xB4(struct thread* thread, int value);
--- a/headers/functions/arm9.h
+++ b/headers/functions/arm9.h
@@ -969,7 +969,7 @@ void GeomMtxLoad4x3(struct matrix_4x3* matrix);
 void GeomMtxMult4x3(struct matrix_4x3* matrix);
 void GeomGxFifoSendMtx4x3(struct matrix_4x3* matrix, void* gxfifo);
 int IncrementThreadCount(void);
-void InsertThreadOnList(struct thread* thread);
+void InsertThreadIntoList(struct thread* thread);
 void ThreadStart(struct thread* thread, void* function, struct thread* param_3, void* stack_area,
                  int stack_size, int param_6);
 void ThreadExit(void);
--- a/headers/functions/arm9.h
+++ b/headers/functions/arm9.h
@@ -968,6 +968,13 @@ void EnableVramBanksInSet(struct vram_banks_set* vram_banks);
 void GeomMtxLoad4x3(struct matrix_4x3* matrix);
 void GeomMtxMult4x3(struct matrix_4x3* matrix);
 void GeomGxFifoSendMtx4x3(struct matrix_4x3* matrix, void* gxfifo);
+int IncrementThreadCount(void);
+void InsertThreadOnList(struct thread* thread);
+void ThreadStart(struct thread* thread, void* function, struct thread* param_3, void* stack_area,
+                 int stack_size, int param_6);
+void ThreadExit(void);
+void SetThreadField0xB4(struct thread* thread, int value);
+void InitThread(struct thread* thread, void* function, void* stack_area_minus_4);
 uint16_t GetTimer0Control(void);
 int ClearIrqFlag(void);
 int EnableIrqFlag(void);
--- a/headers/types/common/common.h
+++ b/headers/types/common/common.h
@@ -1048,7 +1048,8 @@ struct thread {
     // 0x68: Pointer to the next thread. This forms a linked list sorted in ascending order
     // according to sorting_order
     struct thread* next_thread;
-    int thread_id;           // 0x6C: Seems to be a thread ID that gets incremented for each new thread created.
+    // 0x6C: Seems to be a thread ID that gets incremented for each new thread created.
+    int thread_id;
     int sorting_order;       // 0x70: Unknown. Used to sort threads.
     undefined4 field_0x74;   // Initialized to 0
     undefined4 field_0x78;   // Initialized to 0
--- a/headers/types/common/common.h
+++ b/headers/types/common/common.h
@@ -992,6 +992,11 @@ struct wavi_data {
 };
 ASSERT_SIZE(struct wavi_data, 20); // Likely longer
 
+// Specifies a function called when a thread starts running
+typedef void (*thread_entry_fn_t)(void);
+// Specifies a function called when a thread exits
+typedef void (*thread_exit_fn_t)(void);
+
 // Contains information about a running thread
 struct thread {
     int flags;                // Probably a flags field. Usually 0x1F.
@@ -1013,7 +1018,7 @@ struct thread {
     // aligned to 8 bytes.
     void* usable_stack_pointer;
     // 0x3C: Address of the function to return to once the thread exits. Usually ThreadExit.
-    void* return_address;
+    thread_exit_fn_t exit_function;
     void* function_address_plus_4; // 0x40: Pointer to the function to run + 4
     void* stack_pointer_minus_4;   // 0x44: Pointer to the start of the stack area - 4
     undefined field_0x48;
--- a/headers/types/common/common.h
+++ b/headers/types/common/common.h
@@ -992,7 +992,7 @@ struct wavi_data {
 };
 ASSERT_SIZE(struct wavi_data, 20); // Likely longer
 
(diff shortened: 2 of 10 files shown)


symbols/overlay29.yml
```diff
@@ -876,6 +886,17 @@ overlay29:
         Advances one frame. Does not return until the next frame starts.
         
         r0: ? - Unused by the function
+    - name: DisplayDamageNumbers
```
Jawshoeuh, line 889: I believe this function has a niche use for the Stockpile effect in 0x22E46D4 (NA). I doubt it's worth generalizing the function name, but it's probably worth noting.

headers/types/dungeon_mode/enums.h
```diff
@@ -1391,4 +1391,22 @@ union item_index {
...
+    NUMBER_COLOR_RED_0x1 = 1,
+    NUMBER_COLOR_DARK_GREEN = 2,
+    NUMBER_COLOR_YELLOW = 3,
+    NUMBER_COLOR_UNK_0x4 = 4,
+    NUMBER_COLOR_UNK_0x5 = 5,
+    NUMBER_COLOR_RED_0x6 = 6,
+    NUMBER_COLOR_UNK_0x7 = 7,
+    NUMBER_COLOR_UNK_0x8 = 8,
+    NUMBER_COLOR_UNK_0x9 = 9,
+    NUMBER_COLOR_GREEN_0xA = 0xA,
```
Jawshoeuh, line 1409: Stockpile uses 11
```suggestion
    NUMBER_COLOR_GREEN_0xA = 10,
    NUMBER_COLOR_LIGHT_BLUE = 11,
```

Jawshoeuh (commented):
Minor comment about stockpile and damage numbers.

symbols/ram.yml
```diff
@@ -472,6 +472,14 @@ ram:
         Bitset of enabled VRAM banks
         
         type: vram_banks_set
+    - name: THREAD_INFO_STRUCT
+      address:
+        EU: 0x22B9F88
+        NA: 0x22B9648
+      length:
+        EU: 0x4
+        NA: 0x4
```
UsernameFodder, line 481: Shouldn't this be the size of `struct thread_info`? But also, since that's uncertain, it might be better to just omit the length here.

symbols/overlay29.yml
UsernameFodder: How is it used for Stockpile? If it's just the same damage number effect, I feel that it's covered by the general case and maybe not worth noting. It might make more sense to discuss this in a comment for the `NUMBER_COLOR_LIGHT_BLUE` value you suggested in the headers.

symbols/overlay29.yml
```diff
@@ -876,6 +886,17 @@ overlay29:
...
+      address:
+        EU: 0x22EB0C8
+        NA: 0x22EA718
+      description: |-
+        Displays damage numbers or the "MISS" text above a monster.
+        
+        r0: Damage amount. Can be negative. 9999 displays "MISS" instead.
+        r1: Entity above which the numbers will be displayed
+        r2: True to display a plus or minus sign before the numbers, false to hide it
+        r3: Color of the numbers. -1 to determine it automatically.
```
UsernameFodder, line 899:
```suggestion
        r3: Color of the numbers. NUMBER_COLOR_AUTO to determine it automatically.
```

symbols/arm9.yml
```diff
@@ -9343,6 +9343,71 @@ arm9:
...
         r0: matrix_4x3 pointer
         r1: GXFIFO pointer
+    - name: IncrementThreadCount
+      address:
+        EU: 0x20794E8
+        NA: 0x2079150
+      description: |-
+        Increments thread_info::thread_count by 1 and returns the new value.
+        
+        return: thread_info::thread_count + 1
```
UsernameFodder, line 9353: This phrasing is ambiguous (does `thread_count` refer to the value before or after the increment?). I think it'd be better to just say something like "return: new thread count".

symbols/arm9.yml
```diff
@@ -9343,6 +9343,71 @@ arm9:
...
+    - name: InsertThreadOnList
```
UsernameFodder, line 9354: "On" seems like the wrong verb. Can we call this something like `AddThreadToList` or `InsertThreadIntoList`?

symbols/arm9.yml
```diff
@@ -9343,6 +9343,71 @@ arm9:
...
+      address:
+        EU: 0x2079630
+        NA: 0x2079298
+      description: |-
+        Inserts a new thread on the linked thread list (see thread_info::first_thread).
+        
+        The order of the list will be maintained.
```
UsernameFodder, line 9361: I think this could be misinterpreted as the new thread being inserted at the end, thereby "maintaining" the old list order. It might be clearer to phrase it like:

```suggestion
        The thread is inserted in sorted order.
```

headers/types/common/common.h
```diff
@@ -992,6 +992,140 @@ struct wavi_data {
...
+    undefined field_0x1E;
+    undefined field_0x1F;
+    int thread_count; // 0x20: Total amount of threads (or maybe just the active ones?)
+    undefined field_0x24;
+    undefined field_0x25;
+    undefined field_0x26;
+    undefined field_0x27;
+    undefined* field_0x28;
+    // 0x2C: Points to the first thread on the linked list sorted by thread::sort_order
+    struct thread* first_thread;
```
UsernameFodder, line 1121: Can we maybe call this `thread_list_head`? I think that's slightly clearer, and the `_head` should distinguish it from an array.

symbols/overlay29.yml
End45: Oh right, I didn't remember that. I'll change it.

headers/types/dungeon_mode/enums.h
End45: Thanks!

headers/types/common/common.h
```diff
@@ -992,6 +992,140 @@ struct wavi_data {
...
+    undefined4 field_0x28;    // Initialized to 0
+    undefined4 field_0x2C;    // Initialized to 0
+    undefined4 field_0x30;    // Initialized to 0
+    undefined4 field_0x34;    // Initialized to 0
+    // 0x38: Points to the area of the allocated stack area that can actually be used to store data.
+    // This is equal to stack_pointer - 0x44, with an additional - 4 if the resulting addess is not
+    // aligned to 8 bytes.
+    void* usable_stack_pointer;
+    // 0x3C: Address of the function to return to once the thread exits. Usually ThreadExit.
+    void* return_address;
```
UsernameFodder, line 1016: Is this always a function with no params? Can we define a function pointer type?

```c
typedef void (*thread_exit_fn_t)(void);
```

headers/functions/arm9.h
```diff
@@ -968,6 +968,13 @@ void EnableVramBanksInSet(struct vram_banks_set* vram_banks);
 void GeomMtxLoad4x3(struct matrix_4x3* matrix);
 void GeomMtxMult4x3(struct matrix_4x3* matrix);
 void GeomGxFifoSendMtx4x3(struct matrix_4x3* matrix, void* gxfifo);
+int IncrementThreadCount(void);
+void InsertThreadOnList(struct thread* thread);
+void ThreadStart(struct thread* thread, void* function, struct thread* param_3, void* stack_area,
```
UsernameFodder, line 973: Is the entry function always have no params? Can we define a function pointer type?

```c
typedef void (*thread_entry_fn_t)(void);
```

UsernameFodder (changes requested):
Interesting, I didn't know ARM9 had threading facilities. How are threads even managed without an OS?

symbols/overlay29.yml
UsernameFodder: Alright, new description looks fine.

symbols/ram.yml
End45: Oh, right. I incorrectly wrote this down as a pointer at first but later corrected it. I must have forgotten to change the size.

> since that's uncertain, it might be better to just omit the length here.

Alright

symbols/arm9.yml
End45: Sounds good

symbols/arm9.yml
End45: Right, thanks.

symbols/arm9.yml
End45: Oh, you're right. I didn't think of that. I'll change it.

headers/types/common/common.h
End45: I didn't check. But I don't see a way to pass parameters, so probably?
Where would that definition be placed?

UsernameFodder: Probably in `common.h` with all the other threading stuff. Before `struct thread` maybe?

headers/functions/arm9.h
UsernameFodder: Let's do

```c
typedef void (*thread_entry_fn_t)(void);
typedef void (*thread_exit_fn_t)(void);

// Contains information about a running thread
struct thread {
    ...

**Comment by theCapypara** [comment]
Are these actually threads or is this more like an asynchronous task system?

**Comment by End45** [comment]
> Are these actually threads or is this more like an asynchronous task system?

I don't know how the underlying implementation works. I just know that code is run in parallel, that each thread or whatever it is has a different stack, and that they seem to always have the same return function.

File: headers/functions/arm9.h (hunk 10 above)
Comment (End45) [reply headers/functions/arm9.h]: Changed
Hopefully I defined them right, this C syntax is so counterintuitive to me.

Comment (UsernameFodder) [reply headers/functions/arm9.h]: Yeah function pointer syntax is notorious for being confusing. This looks correct. At least once you get the typedef right, it's simple to use.

File: headers/functions/arm9.h [hunk 11]
```diff
@@ -968,6 +968,13 @@ void EnableVramBanksInSet(struct vram_banks_set* vram_banks);
 void GeomMtxLoad4x3(struct matrix_4x3* matrix);
 void GeomMtxMult4x3(struct matrix_4x3* matrix);
 void GeomGxFifoSendMtx4x3(struct matrix_4x3* matrix, void* gxfifo);
+int IncrementThreadCount(void);
+void InsertThreadIntoList(struct thread* thread);
+void ThreadStart(struct thread* thread, thread_entry_fn_t function, struct thread* param_3,
```
UsernameFodder: I'm gonna change this to `StartThread`, since it does stuff _to/with_ a thread (like `InitThread`) rather than being done _by_ a thread (like `ThreadExit`). Lmk if you object.

headers/types/common/common.h
```diff
@@ -992,6 +992,145 @@ struct wavi_data {
 };
 ASSERT_SIZE(struct wavi_data, 20); // Likely longer
 
+// Specifies an entry function for a thread's execution
```
UsernameFodder, line 995: FYI I rephrased the comment a bit so it couldn't be mistaken for being a "pre-startup" routine that might return quickly (similar to the exit function). This way it's more obvious that this is the thread's "main function". Lmk if that doesn't make sense.

UsernameFodder (approved):
> And with this, the info dump is finally complete! Only took 1.5 years.

🎉 It's super exciting to have everything in one place. Thanks for putting the time into dumping it all!

PR #220, opened 2023-12-09
Added sound envelope functions

AnonymousRandomPerson:
Contributed by Rhokin in the SkyTemple Discord. https://discord.com/channels/710190644152369162/804247227991326730/1182773685317271634

### DIFF
--- a/headers/types/common/common.h
+++ b/headers/types/common/common.h
@@ -897,6 +897,34 @@ struct wan_table {
 };
 ASSERT_SIZE(struct wan_table, 5392);
 
+struct sound_envelope_parameters {
+    uint8_t use_envelope;
+    uint8_t slide_time_multiplier;
+    uint16_t field2_0x2;
+    uint32_t field3_0x4;
+    uint8_t attack_begin;
+    uint8_t attack_time;
+    uint8_t decay_time;
+    uint8_t sustain_level;
+    uint8_t hold_time;
+    uint8_t sustain_time;
+    uint8_t release_time;
+    uint8_t field11_0xe;
+};
+ASSERT_SIZE(struct sound_envelope_parameters, 16);
+
+struct sound_envelope {
+    struct sound_envelope_parameters parameters;
+    int32_t current_volume;
+    int32_t volume_delta;
+    int32_t ticks_left;
+    uint8_t state;
+    uint8_t target_volume;
+    uint8_t update_volume;
+    uint8_t _unused;
+};
+ASSERT_SIZE(struct sound_envelope, 32);
+
 // Store one boolean per vram bank
 struct vram_banks_set {
     bool vram_A : 1;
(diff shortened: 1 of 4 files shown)


headers/types/common/common.h
```diff
@@ -897,6 +897,34 @@ struct wan_table {
...
+    uint8_t release_time;
+    uint8_t field11_0xe;
+};
+ASSERT_SIZE(struct sound_envelope_parameters, 16);
+
+struct sound_envelope {
+    struct sound_envelope_parameters parameters;
+    int32_t current_volume;
+    int32_t volume_delta;
+    int32_t ticks_left;
```
End45, line 920: `int` is more commonly seen throughout the repo, although I'm not sure if it's required since `int32_t` is also used sometimes.

End45 (changes requested):

headers/types/common/common.h
```diff
@@ -897,6 +897,34 @@ struct wan_table {
 };
 ASSERT_SIZE(struct wan_table, 5392);
 
+struct sound_envelope_parameters {
+    uint8_t use_envelope;
+    uint8_t slide_time_multiplier;
+    uint16_t field2_0x2;
+    uint32_t field3_0x4;
```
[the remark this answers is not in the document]
UsernameFodder: This is actually the new style of default field names in newer versions of Ghidra, so I think either convention is okay. Makes it easier for someone porting notes from a Ghidra project.

headers/types/common/common.h
UsernameFodder: Yeah either is fine, I don't think it matters much.

UsernameFodder (changes requested):
Can you update CREDITS.txt?

headers/types/common/common.h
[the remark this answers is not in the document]
End45: Interesting. It seems a bit pointless to include the field number.
But if that's how Ghidra generates it, I'm guessing it would be annoying to change it manually for each field.

headers/types/common/common.h
End45: Alright. I personally prefer `int` since it's simpler, but I guess either works.

headers/types/common/common.h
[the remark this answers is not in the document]
AnonymousRandomPerson: I didn't notice that the other style without the field number was present in this repo. Tbh I prefer not having the field number as well, so I'll change it to that.

headers/types/common/common.h
AnonymousRandomPerson: I'll go ahead and use the more prevalent `int`.

UsernameFodder (approved):
Thanks for porting these over!

UsernameFodder:
(Will merge within the next day if End has no further comments to make)

End45 (approved):

PR #222, opened 2023-12-10
Added LFO functions

AnonymousRandomPerson:
Contributed by Rhokin.


symbols/arm9.yml
```diff
@@ -12904,6 +12997,10 @@ arm9:
         See the struct definitions and End45's dungeon data document for more info.
         
         type: struct map_marker[310]
+    - name: LFO_OUTPUT_VOICE_UPDATE_FLAGS
```
UsernameFodder, line 13000: Please add lengths for the data symbols, seeing as we know their types.

headers/types/common/common.h
```diff
@@ -925,6 +925,57 @@ struct sound_envelope {
 };
 ASSERT_SIZE(struct sound_envelope, 32);
 
+struct dse_lfo;
+
+typedef int (*SoundLfoWaveformCallback)(struct dse_lfo* lfo);
```
UsernameFodder, line 930: Typedef'd function pointer types follow `snake_case` convention like other types.

symbols/arm9.yml
```diff
@@ -9998,6 +9998,11 @@ arm9:
         
         r0: Dungeon ID
         return: Dungeon mode
+    - name: SoundUtilGetRandomNumber
+      address:
+        EU: 0x206CC8C
+        NA: 0x206C8F4
+      description: "return: random number"
```
UsernameFodder, line 10005: Do we know the range of random numbers possible?

symbols/arm9.yml
```diff
@@ -9998,6 +9998,11 @@ arm9:
         
         r0: Dungeon ID
         return: Dungeon mode
+    - name: SoundUtilGetRandomNumber
```
End45, line 10001: So the sound engine uses a separate RNG function? How interesting. Do we know if it differs from the usual RNG?

symbols/arm9.yml
```diff
@@ -10134,6 +10139,94 @@ arm9:
...
+        EU: 0x2075644
+        NA: 0x20752AC
+      description: |-
+        r0: LFO bank pointer
+        r1: Level
+    - name: SoundLfoBankTick
+      address:
+        EU: 0x2075690
+        NA: 0x20752F8
+      description: "r0: LFO bank pointer"
```
UsernameFodder, line 10166: The prototype says this returns a `uint16_t`?

UsernameFodder (changes requested):

symbols/arm9.yml
AnonymousRandomPerson: This is probably a part of the Procyon DSE library rather than the main EoS codebase, which would explain why it's a separate function. It looks like a different implementation, though I haven't traced it far enough to determine what RNG mechanism is being used.

symbols/arm9.yml
AnonymousRandomPerson: Looks like [0, 32767] based on the ASM. I'll add this to the description.

symbols/arm9.yml
AnonymousRandomPerson: Added.

UsernameFodder (approved):

PR #227, opened 2023-12-20
Some tiny menu findings

Adex-8x:
Some small findings related to investigating the menu created by pressing X during an overworld segment.


UsernameFodder (approved):
Thanks! Seeing more window stuff is exciting.

PR #228, opened 2023-12-21
Miscellaneous missed stuff from the last PR

ronikirla:
- The function to update window contents was missing, even though it was referenced in a comment
- Added unknowns to `enum fade_status_dungeon`
- Added two bits to `struct preprocessor_flags`. Current only known purpose is that it's used when creating the play time string.


UsernameFodder (approved):
Looks good, thanks.

PR #234, opened 2023-12-29
Fix enemy_density type

End45:
The enemy density is signed, as explained in the comment.


End45:
Nevermind, it is an unsigned byte in-game, so the behavior explained in the comment never takes place, even though the check is there. I'll make a PR to remove the comment instead.

PR #235, opened 2023-12-29
Update enemy_density comment

End45:
Continuation of #234

### DIFF
--- a/headers/types/dungeon_mode/dungeon_mode.h
+++ b/headers/types/dungeon_mode/dungeon_mode.h
@@ -1121,10 +1121,12 @@ struct floor_properties {
     // 0x5: Controls how many connections will be made between grid cells
     uint8_t floor_connectivity;
     // 0x6: Controls how many enemies will be spawned. If 0, no enemies will spawn, even as the
-    // player walks. The final value will be randomized between density / 2 and density - 1.
-    // The game contains a check that, if the value is < 0, will take absolute value of the density
-    // without adding a random variation; however, the check never passes since the value is
-    // unsigned.
+    // player walks.
+    // The final value will be 0 in the special case that density is 0. Otherwise, the final value
+    // will be randomized between density / 2 and density - 1, but no lower than 1.
+    // The code seems to imply that the value was meant to be signed, since it contains a section
+    // that takes the absolute value of the density while skipping the random variation, similar to
+    // room_density.
     uint8_t enemy_density;
     uint8_t kecleon_shop_spawn_chance;  // 0x7: Percentage chance from 0-100
     uint8_t monster_house_spawn_chance; // 0x8: Percentage chance from 0-100
--- a/headers/types/dungeon_mode/dungeon_mode.h
+++ b/headers/types/dungeon_mode/dungeon_mode.h
@@ -1121,8 +1121,10 @@ struct floor_properties {
     // 0x5: Controls how many connections will be made between grid cells
     uint8_t floor_connectivity;
     // 0x6: Controls how many enemies will be spawned. If 0, no enemies will spawn, even as the
-    // player walks. If > 0, the final value will be randomized between density / 2 and density - 1.
-    // If < 0, its absolute value will be used without adding a random variation.
+    // player walks. The final value will be randomized between density / 2 and density - 1.
+    // The game contains a check that, if the value is < 0, will take absolute value of the density
+    // without adding a random variation; however, the check never passes since the value is
+    // unsigned.
     uint8_t enemy_density;
     uint8_t kecleon_shop_spawn_chance;  // 0x7: Percentage chance from 0-100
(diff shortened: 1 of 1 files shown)


headers/types/dungeon_mode/dungeon_mode.h
```diff
@@ -1121,8 +1121,10 @@ struct floor_properties {
     // 0x5: Controls how many connections will be made between grid cells
     uint8_t floor_connectivity;
     // 0x6: Controls how many enemies will be spawned. If 0, no enemies will spawn, even as the
-    // player walks. If > 0, the final value will be randomized between density / 2 and density - 1.
-    // If < 0, its absolute value will be used without adding a random variation.
+    // player walks. The final value will be randomized between density / 2 and density - 1.
+    // The game contains a check that, if the value is < 0, will take absolute value of the density
+    // without adding a random variation; however, the check never passes since the value is
```
UsernameFodder, line 1126: Where is this check? I don't see one in `MarkEnemySpawns` which is what adds the variation; it only checks for 0 exactly, and a check to make sure it's always at least 1. I'm not even sure it makes sense to say a < 0 check would always fail, since it's not like the asm `lt` and `le` conditionals can tell the difference between a signed and unsigned number. It's all about interpretation after all.

I'm wondering if this comment was just copied from `room_density`, which _does_ allow a negative value in `AssignRooms`.

End45: It's located right at the start of `MarkEnemySpawns`. These are my notes on the matter (EU offsets):
```
0x2342064: Load enemy density
0x2342074: If it's >= 1 {
    0x2342080: RNG(density / 2, density)
        This means the real enemy density will be between half the original value (rounded down) and the original value - 1
} else (0x2342094) {
    The density (r8) is set to its absolute value
} end if at 0x23420A0
```
I just checked the ASM again and they seem correct. Both EU and US have this check.

> it only checks for 0 exactly

Where are you seeing this? Ghidra disassembly? Maybe it sees a signed value and a `>= 1` check and disassembles it as `!= 0`.

> I'm not even sure it makes sense to say a < 0 check would always fail

Well, technically it passes if the value is 0. But the intended effect never takes place. Maybe that part should be reworded?

> it's not like the asm lt and le conditionals can tell the difference between a signed and unsigned number

Yeah, in this case, the problem is that the value gets loaded with `ldrb`, because it's unsigned. If it was loaded with `ldrsb`, the check would work as intended.

UsernameFodder: > Where are you seeing this? Ghidra disassembly? Maybe it sees a signed value and a >= 1 check and disassembles it as != 0.

Oh yeah, you're right. Looking at the asm directly, it's just `cmp r1,#0x1`.

> Well, technically it passes if the value is 0. But the intended effect never takes place. Maybe that part should be reworded?

Yeah can we rephrase this a bit? Since it _is_ unsigned, I don't think it's valid to say `enemy_density` could ever be negative. A value of "-1" would just be interpreted as "0xFF".

End45: I reworded it. Let me know if that's better.

UsernameFodder: Is the check entirely useless? I think it still applies for 0 at least, otherwise the final value would get clamped to 1. The clamping logic means 0 needs to be treated specially. The logic is like this:

```c
  y = (uint)floor_props->enemy_density;
  if (y == 0) {
    iVar3 = abs(0);
  }
  else {
    iVar3 = DungeonRandRange(y / 2,y);
    if (iVar3 < 1) {
      iVar3 = 1;
    }
  }
```

UsernameFodder: Do we really need to specify that the `abs` call is "useless"? How about we just say what the result is? Something like this:

> The final value will be 0 in the special case that density is 0. Otherwise, the final value will be randomized between density / 2 and density - 1, but no lower than 1.

End45: I kinda wanted to state that there's an unused behavior that was meant to be used alongside a signed value. Maybe we could add a second paragraph stating that after yours, like so:

> Controls how many enemies will be spawned. If 0, no enemies will spawn, even as the player walks.
> The final value will be 0 in the special case that density is 0. Otherwise, the final value will be randomized between density / 2 and density - 1, but no lower than 1.
> The code seems to imply that the value was meant to be signed, since it contains a section that takes the absolute value of the density while skipping the random variation, similar to `room_density`.

UsernameFodder: That seems reasonable to me.

End45: Updated

UsernameFodder (approved):
Thanks, I think we got to a good final result.

PR #239, opened 2023-12-30
Make docstring preamble a regular comment instead of a doc comment

tech-ticks:
Ensures that these lines are excluded in some IDEs and doc generation tools.


UsernameFodder (approved):
Thanks!

Issue #238, opened 2023-12-30
Support symbol name aliases


UsernameFodder:
**Is your feature request related to a problem? Please describe.**
Various projects rely on pmdsky-debug symbols, including https://github.com/SkyTemple/pmdsky-debug-py and https://github.com/SkyTemple/c-of-time. When symbols are renamed, it can cause compatibility issues downstream.

**Describe the solution you'd like**
Add support for an optional `aliases` field in the YAML files, like this:
```yaml
- name: SomeFunction
  aliases:
    - SomeFunctionAlias1
    - SomeFunctionAlias2
```
This will make things symbols searchable by more names (including old ones), while making it possible for tooling to decide whether or not to use the extra alias names.

- `resymgen gen` should include aliases in the `.ghidra` (as extra symbols) and `.json` (in a separate `"aliases"` property) output formats, since Ghidra can support alternate symbol names. The `.sym` output format should only use the primary name, since No$GBA only supports one symbol name per address.
- `resymgen fmt` should order alias names alphabetically.
- `resymgen check`:
    - Alias names should be subject to the same naming convention checks (`--function-names` and `--data-names`) as primary symbol names.
    - Alias names should be included in the `--unique-symbols` check so they don't clash with any other symbol or alias names.
- `resymgen merge` should properly deduplicate aliases when merging alias lists.
- The `resymgen` docs and unit tests should be updated accordingly.

Aliases should _not_ be included in the header files, since this would lead to a lot of duplication. Instead, they should be auto-generated for projects that need them (like c-of-time). The `headers-with-docstrings` artifact should include these generated aliases.

The following tooling should also support aliases:
- If possible, the GitHub Actions PR checks should compare the symbol tables in the base commit with the new changes, and fail if a symbol has been renamed without adding the old name as an alias. This is probably doable by seeing if there are any symbol names in the base commit that no longer exist in the new version (if an old symbol is being moved, there's nothing that can be done with aliases).
- `symdiff.py` should detect updates to alias lists.
- `import_symbols_json.py` and `import_symbols_ntr_ghidra.py` should create additional labels for aliases.
- `add_header_docstrings.py` should support adding aliases alongside header docstrings.

**Describe alternatives you've considered**
- Never rename symbols. Doesn't seem practical.
[...]

PR #240, opened 2023-12-30
Revert #236

UsernameFodder:
See the comments on https://github.com/UsernameFodder/pmdsky-debug/pull/236. Hopefully this will only be temporary, and we can add these back in later once we figure out a good way to solve the current issues.


UsernameFodder:
My current thought is that we can add these back as secondary aliases once we do https://github.com/UsernameFodder/pmdsky-debug/issues/238.

End45 (approved):
I think the alias idea is a good solution to this problem. But until then, I agree with reverting the changes.

theCapypara:
I personally have no strong opinion on immediately reverting, as long as we find a somewhat fast implementation that solves the issue, either is fine.

PR #247, opened 2024-01-12
Assortment of arm9 findings

Adex-8x:
Documents various things from arm9, though mostly related to missions.


headers/types/dungeon_mode/enums.h
```diff
@@ -731,6 +731,8 @@ enum mission_type {
     MISSION_ARREST_OUTLAW = 10,
     MISSION_CHALLENGE_REQUEST = 11,
     MISSION_TREASURE_MEMO = 12,
+    MISSION_TYPE_UNK_13 = 13,
```
UsernameFodder, line 734: Usually I prefer to use hex with an explicit `0x` in unknown enum labels to avoid ambiguity on whether a number is decimal or hex. Can we call this `MISSION_TYPE_UNK_0xD`?

headers/functions/arm9.h
```diff
@@ -740,7 +743,9 @@ bool RestoreScriptVariableValues(void* src);
 void InitScenarioScriptVars(void);
 void SetScenarioScriptVar(enum script_var_id id, uint8_t val0, uint8_t val1);
 int GetSpecialEpisodeType(void);
+void SetSpecialEpisodeType(int val);
```
UsernameFodder, line 746: Is `enum special_episode_type` applicable for the parameter?

symbols/arm9.yml
```diff
@@ -6997,6 +7031,16 @@ arm9:
...
         return: special episode type
+    - name: IsSpecialEpisodeOpen
+      address:
+        EU: 0x204CC84
+        NA: 0x204C94C
+        JP: 0x204CCAC
+      description: |-
+        Checks if a special episode is unlocked from the SPECIAL_EPISODE_OPEN script variable.
+        
+        r0: index
```
UsernameFodder, line 7042: Do you know what the index here means? Is it the same as the special episode type?

symbols/arm9.yml
```diff
@@ -9912,22 +10040,34 @@ arm9:
         r1: Pointer to some struct that contains the subtype of the mission to check on its first byte
         r2: Dungeon ID
         return: True if at least one mission meets the specified requirements, false otherwise.
+    - name: GetAllPossibleMonsters
+      address:
+        EU: 0x205FAB4
+        NA: 0x205F738
+        JP: 0x205FA28
+      description: |-
+        Stores MISSION_MONSTER_LIST_PTR into the passed buffer and retreives the number of monsters that can be used in a mission.
```
UsernameFodder, line 10049:
```suggestion
        Stores MISSION_MONSTER_LIST_PTR into the passed buffer and retrieves the number of monsters that can be used in a mission.
```

UsernameFodder (changes requested):

headers/functions/arm9.h
Adex-8x: Oh, completely missed that, yeah it would

symbols/arm9.yml
Adex-8x: Oops yeah, my bad

UsernameFodder (approved):
I changed `IsSpecialEpisodeOpen` to use `enum special_episode_type` for the first param, since you changed the YAML description to indicate it. Looks good otherwise.

PR #249, opened 2024-02-08
Added DSE function headers/structs

AnonymousRandomPerson:
Contributed by Rhokin.

Rhokin updated a couple names. I opted to rename the symbols directly instead of making aliases, since they were just added two weeks ago in https://github.com/UsernameFodder/pmdsky-debug/pull/248 and are essentially part of the same research session. I can change them to aliases if desired.


headers/types/common/common.h
```diff
@@ -1072,6 +1072,517 @@ struct wavi_data {
...
+
+    uint8_t loader_thread_priority;
+    uint8_t field_0x1DF1;
+    int8_t quit_sample_loader;
+    int8_t is_sample_loader_sleeping;
+    uint8_t sample_loader_thread[192]; // Again, OSThread
+    void* sample_loader_thread_stack;
+    struct dse_wavebank* loading_bank;
+    void* loading_bank_pcm_data;
+    // There's more, but nothing labeled.
```
UsernameFodder, line 1582: Does this mean the struct is actually bigger? This comment is unclear.

UsernameFodder (changes requested):

headers/types/common/common.h
AnonymousRandomPerson: Yes. I'll update the comment to match some of the other structs with ambiguous sizes.

headers/types/types.h
```diff
@@ -5,5 +5,6 @@
 #include "dungeon_mode/dungeon_mode.h"
 #include "ground_mode/ground_mode.h"
 #include "sentry_duty/sentry_duty.h"
+#include "sound/sound.h"
```
UsernameFodder, line 8: Can we put this in `common/sound.h`? The sound engine isn't its own mode.

AnonymousRandomPerson: Done.

headers/types/sound/sound.h
```diff
@@ -0,0 +1,671 @@
...
+    uint16_t field_0x1CFC;
+    uint16_t field_0x1CFE;
+    int num_active_voices;
+    int prev_nom_active_voices;
+    int ticks_until_num_active_voices_timeout;
+    uint32_t microseconds_per_driver_tick_2;
+    uint32_t field_1D10;
+    uint32_t field_1D14;
+
+    struct thread* driver_thread[48];
```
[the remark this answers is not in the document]
AnonymousRandomPerson: Oops, that should be the struct itself, not a pointer.

headers/types/common/sound.h
```diff
@@ -0,0 +1,671 @@
+// Types used by the DSE sound library.
+
+#ifndef HEADERS_TYPES_SOUND_H_
```
UsernameFodder, line 3: The define needs updating

headers/types/types.h
```diff
@@ -2,6 +2,7 @@
 #define HEADERS_TYPES_H_
 
 #include "common/common.h"
+#include "common/sound.h"
```
UsernameFodder, line 5: I'm not a fan of this import structure. Can we `#include sound.h` from `common.h`? I think circular imports can be avoided by moving `struct thread` and `struct thread_info` to `util.h` and then including `enums.h`, `file_io.h`, and `util.h` individually in `sound.h` rather than the full `common.h`.

UsernameFodder (changes requested):

headers/types/common/sound.h
AnonymousRandomPerson: I was looking at `window.h`, but I see that this one is also wrong. I'll fixed both.

headers/types/types.h
AnonymousRandomPerson: Yeah, this works.

headers/types/common/sound.h
UsernameFodder: Oops, thanks.

UsernameFodder (approved):
LGTM

PR #251, opened 2024-03-23
Replace GetSleepAnimationId with GetIdleAnimationId

End45:
Better documents `GetSleepAnimationId`, which is actually used to get the animation a monster should play while idling. The function has been renamed to `GetIdleAnimationId`.

Also documents `ChangeMonsterAnimation`.

### DIFF
--- a/headers/functions/overlay29.h
+++ b/headers/functions/overlay29.h
@@ -301,7 +301,9 @@ bool TryDecreaseLevel(struct entity* user, struct entity* target, int n_levels);
 bool LevelUp(struct entity* user, struct entity* target, bool message, bool dialogue);
 void GetMonsterMoves(struct move_id_16* out_moves, enum monster_id monster_id, int level);
 void EvolveMonster(struct entity* user, struct entity* target, enum monster_id new_monster_id);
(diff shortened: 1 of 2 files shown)


UsernameFodder (approved):
Looks fine to me, will wait a day before merging since there's a symbol name being removed.

UsernameFodder:
Oh well I guess an alias is okay too. I figured you didn't want to add one since it's not exactly correct, but I don't mind either way.

End45:
I just added the old name as an alias (I forgot the alias system is a thing now), although I don't know if it would be better to remove it, since the old name is incorrect.

UsernameFodder:
> I just added the old name as an alias (I forgot the alias system is a thing now), although I don't know if it would be better to remove it, since the old name is incorrect.

I kind of doubt many people use this symbol? So if I had to choose I might lean towards no alias, but will defer to you.

End45:
Alright, I'll remove it then. I thought it was mandatory to add an alias since the checks were failing.

UsernameFodder:
> Alright, I'll remove it then. I thought it was mandatory to add an alias since the checks were failing.

It'll only fail for the PR the change is introduced in. So it's an ignorable error.

PR #252, opened 2024-03-26
Fix `ExecuteMonsterAction` return value

tech-ticks:
`ExecuteMonsterAction` is supposed to have a return value, which became apparent after trying to hook calls to this function via a wrapper with a `void` return type and watching every Pokémon attack three times in a row. The actual return value seems to be a `bool` because either 0 or 1 are moved into r0 before the function returns.

### DIFF
--- a/symbols/overlay29.yml
+++ b/symbols/overlay29.yml
@@ -2506,6 +2506,7 @@ overlay29:
         Executes the set action for the specified monster. Used for both AI actions and player-inputted actions. If the action is not ACTION_NOTHING, ACTION_PASS_TURN, ACTION_WALK or ACTION_UNK_4, the monster's already_acted field is set to true. Includes a switch based on the action ID that performs the action, although some of them aren't handled by said swtich.
         
         r0: Pointer to monster entity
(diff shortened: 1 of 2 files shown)


End45:
I took a quick look, it looks like the return value is not used when this function is called for the player or the enemies, but it is checked after executing the action of an ally. If the result is true, the AI is run again for the current ally, and it performs another action. This can happen up to three times.
Quite interesting.

symbols/overlay29.yml
```diff
@@ -2506,6 +2506,7 @@ overlay29:
         Executes the set action for the specified monster. Used for both AI actions and player-inputted actions. If the action is not ACTION_NOTHING, ACTION_PASS_TURN, ACTION_WALK or ACTION_UNK_4, the monster's already_acted field is set to true. Includes a switch based on the action ID that performs the action, although some of them aren't handled by said swtich.
         
         r0: Pointer to monster entity
+        return: ?
```
UsernameFodder, line 2509: Let's update this to what End found?

> If the result is true, the AI is run again for the current ally, and it performs another action. This can happen up to three times.

UsernameFodder (approved):

UsernameFodder:
Thanks!

PR #253, opened 2024-04-15
Document some dungeon tip function + fix some signature type

marius851000:
document DisplayFloorTip and DisplayItemTip, add return for DisplayDungeonTip and specify the exact type for pack_file_id in LoadWteFromFileDirectory


UsernameFodder (approved):
Thanks! I made some minor fixes but this otherwise looks good.

PR #254, opened 2024-04-15
Touchscreen struct and script parameter conversion functions

End45:
This PR contains two minor additions.
First, I've documented a struct that can be used to read data about the current status of the touchscreen, including whether it's currently being pressed or not and the position being pressed. This data, unlike the raw values that can be read from the I/O ports, is expressed in pixels starting from the top left corner of the screen, which should be a better spot to read touchscreen coordinates since it's easier to use and consistent across emulators (seems like the raw values are not).
Second, I've researched the function previously known as `ProcessScriptParam` and its sister function. We now have a better understanding of how scripting opcode parameters work, and how the two special bits they contain affect their value.

A couple of questions:
1. Should I document those two special bits on scripting opcode parameters? Should a new type be created for those parameters? It's just a halfword composed of 2 bit flags + a 14-bit number.
2. I don't know if there's a better way of documenting the type returned by `ScriptParamToDecimal`. I'm not sure what name I could give it, nor how to explain it in a clear and concise way. It's just a number that's stored multiplied by 256, should we add a new type for that?


symbols/overlay11.yml
```diff
@@ -76,15 +76,39 @@ overlay11:
...
+        r0: Parameter to convert
+        return: The input parameter, as a signed integer
+    - name: ScriptParamToDecimal
+      address:
+        EU: 0x22E5220
+        NA: 0x22E48E0
+      description: |-
+        Converts the given opcode parameter to a signed decimal number with 8 decimal bits.
+        
+        The resulting number is encoded as (value) * 256, with the last byte acting as a fraction byte capable of representing multiples of 1/256.
```
UsernameFodder, line 103: Throughout the repo, I refer to this (with varying levels of specificity) as an "_n_-bit (un)signed fixed-point (_f_ fraction bits)". So in this case, a 16-bit signed fixed-point number (8 fraction bits). It's much more precise than the term "decimal".

symbols/overlay11.yml
```diff
@@ -76,15 +76,39 @@ overlay11:
...
         
-        r0: On input, the parameter. On output, the modified parameter if either of the two bits listed above is set, same parameter otherwise.
+        The parameter will be returned unchanged unless one of its two most significant bits (0x8000 and 0x4000) are set, in which case both bit will be cleared and the original value will be modified according to the following two rules:
+        - If the 0x4000 bit is set (negative flag), the value will be set to -16384 + value.
+        - If the 0x8000 bit is set (decimal flag), the value will be set to value / 256, rounded down.
+        Both rules can be applied, in the same order as listed, if both conditions are met.
+        
+        r0: Parameter to convert
+        return: The input parameter, as a signed integer
+    - name: ScriptParamToDecimal
```
UsernameFodder, line 96: Let's call this `ScriptParamToFixedPoint16`

symbols/overlay11.yml
```diff
@@ -76,15 +76,39 @@ overlay11:
...
+      address:
+        EU: 0x22E5220
+        NA: 0x22E48E0
+      description: |-
+        Converts the given opcode parameter to a signed decimal number with 8 decimal bits.
+        
+        The resulting number is encoded as (value) * 256, with the last byte acting as a fraction byte capable of representing multiples of 1/256.
+        
+        The parameter will wither be returned unchanged or modified depending on which of its two most significant bits (0x8000 and 0x4000) are set. Both bits are unset before running the operations listed below:
+        - If the 0x4000 bit is set (negative flag), the value will be set to -16384 + value.
```
UsernameFodder, line 106: "negative flag" is alright with me, though my personal preference would be to use the standard "sign bit" term. It's a little unusual because it's a sign bit for what's effectively a _15-bit_ signed integer, but it's analogous to how the highest bit in a normal (two's complement) signed integer turns it negative in the same way.

symbols/overlay11.yml
```diff
@@ -76,15 +76,39 @@ overlay11:
...
+        - If the 0x8000 bit is set (decimal flag), nothing else happens. If it's not set, the value will be set to value * 256.
```
UsernameFodder, line 107: Can we call this the "fixed-point flag"? If it's set, the raw value is interpreted as being fixed-point already. Otherwise, it's assumed to be a normal integer and is converted to fixed-point by left-shifting it by 8 (moving the integer part its proper place).

UsernameFodder (changes requested):

symbols/ram.yml
```diff
@@ -137,6 +137,16 @@ ram:
         EU: 0x4
         NA: 0x4
       description: "Starts at 0 when the game is first launched, and continuously ticks up once per frame while the game is running."
+    - name: TOUCHSCREEN_STATUS
+      address:
+        EU:
+          - 0x22A3F1C
```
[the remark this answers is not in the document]
End45: Oh, right. I just copy-pasted from the above symbol, that's why there's only one. I'll fix it.

symbols/overlay11.yml
End45: Ah, there we go. I knew there had to be a better name. Thanks!

symbols/overlay11.yml
End45: Oh, you're right, its just a sign bit. That makes more sense, I'll change it.

symbols/overlay11.yml
End45: Changed

symbols/overlay11.yml
```diff
@@ -76,15 +76,39 @@ overlay11:
...
       address:
         EU: 0x22E51EC
         NA: 0x22E48AC
         JP: 0x22E5ED8
       description: |-
-        Checks if the two most significant bits (0x8000 and 0x4000) are set on a given script opcode parameter, and modifies the value if so.
+        Converts the given opcode parameter to a signed integer.
         
-        r0: On input, the parameter. On output, the modified parameter if either of the two bits listed above is set, same parameter otherwise.
+        The parameter will be returned unchanged unless one of its two most significant bits (0x8000 and 0x4000) are set, in which case both bit will be cleared and the original value will be modified according to the following two rules:
```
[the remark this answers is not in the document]
End45: Thanks!

symbols/overlay11.yml
```diff
@@ -76,15 +76,39 @@ overlay11:
...
+    - name: ScriptParamToDecimal
+      address:
+        EU: 0x22E5220
+        NA: 0x22E48E0
+      description: |-
+        Converts the given opcode parameter to a signed decimal number with 8 decimal bits.
+        
+        The resulting number is encoded as (value) * 256, with the last byte acting as a fraction byte capable of representing multiples of 1/256.
+        
+        The parameter will wither be returned unchanged or modified depending on which of its two most significant bits (0x8000 and 0x4000) are set. Both bits are unset before running the operations listed below:
```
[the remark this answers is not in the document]
End45: Thanks!

symbols/overlay11.yml
End45: Right

End45:
Alright, everything is updated and fixed. I also rebased master since there were new commits from Marius' PR. Let me know if it looks good now or if something still needs to be changed.
I did not create new types for the fixed-point numbers, since it seems like describing them is enough.

UsernameFodder (approved):
All looks good now, thanks!

PR #256, opened 2024-05-29
Document debug flag 9

End45:


theCapypara:
Is that flag checked when loading a quick save or when saving one?

End45:
> Is that flag checked when loading a quick save or when saving one?

When loading it. To be precise, when selecting the "continue" option in the main menu.
There's another check when the main menu is loaded, although I haven't noticed any obvious difference there. It's probably related to which options are shown on the menu, since they might depend on whether the quicksave is "used up" or not.

UsernameFodder (approved):
Thanks!

PR #257, opened 2024-05-29
Add new fixed-point types

End45:
There's multiple places in the repo that refer to various fixed-point value formats, but they are usually documented as integers or shorts. I thought it could be a good idea to document them explicitly, so I defined new types for them.
`struct fx64` has also been renamed to `struct fx64_16` for consistency.

I also plan to make use of these new types in SkyTemple, so I need them to be distinguishable from regular ints and shorts.


End45:
I have no idea why that particular run fails, I defined my types next to other common types, so they should all work, right?

headers/pmdsky.h
```diff
@@ -63,6 +63,14 @@ typedef unsigned short wchar_t;
 // whatever stdarg.h defines
 typedef void* va_list;
 
+// Define custom types for fixed-point numbers
+typedef int fx32_16;      // 32-bit signed fixed-point number with 16 fraction bits
```
UsernameFodder, line 67: Can we move these to the same place as `struct fx64_16`? They ought to all live in the same place. Also, `pmdsky.h` is only for the core "system" types and things of that nature. I wouldn't consider these typedefs to fall in that category.

The compile-unsized action fails because this section is skipped in favor of importing the raw types from the real system headers (`stdint.h`, etc.). Which is related to the fact that this file only contains core system-level declarations.

UsernameFodder (changes requested):
This seems reasonable. I've thought about doing this myself, but didn't want to define new structs that were just primitive types. Typedefing seems like a good approach.

headers/functions/arm9.h
```diff
@@ -859,7 +861,7 @@ void ResetPlayTimer(struct play_time* igt);
 void PlayTimerTick(struct play_time* igt);
 uint32_t GetPlayTimeSeconds(struct play_time* igt);
 uint32_t SubFixedPoint(uint32_t val_fp, uint32_t dec_fp);
-uint32_t BinToDecFixedPoint(uint32_t* q16);
+uint32_t BinToDecFixedPoint(fx32_16* q16);
```
UsernameFodder, line 864:
```suggestion
uint32_t BinToDecFixedPoint(ufx32_16* q16);
```

headers/functions/overlay29.h
```diff
@@ -617,14 +617,14 @@ void ExecuteMoveEffect(undefined4* param_1, struct entity* attacker, struct move
...
+               fx32_8 damage_mult, enum item_id item_id);
 int DealDamageWithTypeAndPowerBoost(struct entity* attacker, struct entity* defender,
-                                    struct move* move, int damage_mult_fp, enum item_id item_id,
+                                    struct move* move, fx32_8 damage_mult, enum item_id item_id,
                                     enum type_id attack_type, int16_t power_boost);
 int DealDamageProjectile(struct entity* attacker, struct entity* defender, struct move* move,
-                         int power, int damage_mult_fp, enum item_id item_id);
+                         int power, fx32_8 damage_mult, enum item_id item_id);
 int DealDamageWithType(struct entity* attacker, struct entity* defender, enum type_id attack_type,
-                       struct move* move, int damage_mult_fp, enum item_id item_id);
```
UsernameFodder, line 627: Can we still keep all the `_fp` suffixes? It makes it easier to read decompiled code in Ghidra.

UsernameFodder (changes requested):

headers/pmdsky.h
End45: Sure, I'll place them in `util.h`

headers/functions/arm9.h
End45: Fixed, thanks!

headers/functions/overlay29.h
End45: I thought they would no longer be necessary now that the proper type is used for the parameter, but sure.

UsernameFodder (approved):
Thanks! Not sure if this is all the conversions, but some is better than none.

End45:
I searched for "fixed-point" and updated all symbols that mentioned the fixed-point format, so as long as it was documented in text, it should be updated.

PR #258, opened 2024-05-29
Add instruction immediates to header files

End45:
[literals.yml](https://github.com/UsernameFodder/pmdsky-debug/blob/master/symbols/literals.yml) lists values that are embedded into instructions as literals. However, those values are not documented anywhere in the header files. I think it could be helpful to have those explicitly declared with their own type in the headers.
I named the type `sh_imm`, short for "shifted immediate", which I thought was too long. Suggestions for alternatives are welcome though.

I also added the `DEBUG_TEXT_SPEED` literal and renamed `TEXT_SPEED` to `REGULAR_TEXT_SPEED`. Sorry, I saw that only one of the two was documented and couldn't resist.

This PR is created on top of #257, so it can't be merged yet. I'll update it once that one is merged.

TODO: Sort the newly added symbols in the header files. Calculating their exact position is annoying since I have to compare the definitions in literals.yml with the ones in <current binary>.yml, so I'm letting the automated check do it for me.

### DIFF
--- a/headers/data/arm9.h
+++ b/headers/data/arm9.h
@@ -196,10 +196,10 @@ extern int32_t MUSIC_DURATION_LOOKUP_TABLE_2[128];
 extern sound_lfo_waveform_callback LFO_WAVEFORM_CALLBACKS[16];
 extern bool IS_DISP_ON;
 extern uint32_t GXI_DMA_ID;
-extern struct data_processing_instruction JUICE_BAR_NECTAR_IQ_GAIN;
-extern struct data_processing_instruction DEBUG_TEXT_SPEED;
-extern struct data_processing_instruction REGULAR_TEXT_SPEED;
-extern struct data_processing_instruction HERO_START_LEVEL;
-extern struct data_processing_instruction PARTNER_START_LEVEL;
+extern const struct data_processing_instruction JUICE_BAR_NECTAR_IQ_GAIN;
+extern const struct data_processing_instruction DEBUG_TEXT_SPEED;
+extern const struct data_processing_instruction REGULAR_TEXT_SPEED;
+extern const struct data_processing_instruction HERO_START_LEVEL;
+extern const struct data_processing_instruction PARTNER_START_LEVEL;
 
 #endif
--- a/headers/data/arm9.h
+++ b/headers/data/arm9.h
@@ -196,10 +196,5 @@ extern int32_t MUSIC_DURATION_LOOKUP_TABLE_2[128];
 extern sound_lfo_waveform_callback LFO_WAVEFORM_CALLBACKS[16];
 extern bool IS_DISP_ON;
 extern uint32_t GXI_DMA_ID;
-extern const struct data_processing_instruction JUICE_BAR_NECTAR_IQ_GAIN;
-extern const struct data_processing_instruction DEBUG_TEXT_SPEED;
-extern const struct data_processing_instruction REGULAR_TEXT_SPEED;
-extern const struct data_processing_instruction HERO_START_LEVEL;
-extern const struct data_processing_instruction PARTNER_START_LEVEL;
 
 #endif
--- a/headers/data/arm9.h
+++ b/headers/data/arm9.h
@@ -196,5 +196,10 @@ extern int32_t MUSIC_DURATION_LOOKUP_TABLE_2[128];
 extern sound_lfo_waveform_callback LFO_WAVEFORM_CALLBACKS[16];
 extern bool IS_DISP_ON;
 extern uint32_t GXI_DMA_ID;
+extern struct data_processing_instruction JUICE_BAR_NECTAR_IQ_GAIN;
+extern struct data_processing_instruction DEBUG_TEXT_SPEED;
+extern struct data_processing_instruction REGULAR_TEXT_SPEED;
+extern struct data_processing_instruction HERO_START_LEVEL;
+extern struct data_processing_instruction PARTNER_START_LEVEL;
 
 #endif
--- a/headers/data/literals.h
+++ b/headers/data/literals.h
@@ -11,10 +11,10 @@ extern const struct data_processing_instruction PARTNER_START_LEVEL;
 // Overlay 0
 extern const struct data_processing_instruction TOP_MENU_MUSIC_ID;
 
-// Overlay 9
-extern const struct data_processing_instruction TOP_MENU_RETURN_MUSIC_ID;
-
 // Overlay 29
 extern const struct data_processing_instruction NECTAR_IQ_BOOST;
+
+// Overlay 9
+extern const struct data_processing_instruction TOP_MENU_RETURN_MUSIC_ID;
 
 #endif
--- a/headers/data/literals.h
+++ b/headers/data/literals.h
@@ -1,0 +1,20 @@
+#ifndef HEADERS_DATA_LITERALS_H_
+#define HEADERS_DATA_LITERALS_H_
+
+// Arm9
+extern const struct data_processing_instruction JUICE_BAR_NECTAR_IQ_GAIN;
+extern const struct data_processing_instruction DEBUG_TEXT_SPEED;
+extern const struct data_processing_instruction REGULAR_TEXT_SPEED;
+extern const struct data_processing_instruction HERO_START_LEVEL;
+extern const struct data_processing_instruction PARTNER_START_LEVEL;
+
+// Overlay 0
+extern const struct data_processing_instruction TOP_MENU_MUSIC_ID;
+
+// Overlay 9
+extern const struct data_processing_instruction TOP_MENU_RETURN_MUSIC_ID;
+
+// Overlay 29
+extern const struct data_processing_instruction NECTAR_IQ_BOOST;
+
+#endif
--- a/headers/data/literals.h
+++ b/headers/data/literals.h
@@ -1,7 +1,7 @@
 #ifndef HEADERS_DATA_LITERALS_H_
 #define HEADERS_DATA_LITERALS_H_
 
-// Arm9
+// ARM9
 extern const struct data_processing_instruction JUICE_BAR_NECTAR_IQ_GAIN;
 extern const struct data_processing_instruction DEBUG_TEXT_SPEED;
 extern const struct data_processing_instruction REGULAR_TEXT_SPEED;
--- a/headers/data/overlay00.h
+++ b/headers/data/overlay00.h
@@ -1,6 +1,6 @@
 #ifndef HEADERS_DATA_OVERLAY00_H_
 #define HEADERS_DATA_OVERLAY00_H_
 
-extern struct data_processing_instruction TOP_MENU_MUSIC_ID;
+extern const struct data_processing_instruction TOP_MENU_MUSIC_ID;
 
 #endif
--- a/headers/data/overlay00.h
+++ b/headers/data/overlay00.h
@@ -1,0 +1,6 @@
+#ifndef HEADERS_DATA_OVERLAY00_H_
+#define HEADERS_DATA_OVERLAY00_H_
+
+extern struct data_processing_instruction TOP_MENU_MUSIC_ID;
+
+#endif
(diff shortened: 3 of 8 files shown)


End45:
Alright, sync-check doesn't like that symbols in the .h file are not defined in the corresponding .yml file. I guess we'll have to figure out how to deal with that.
Maybe just get rid of literals.yml entirely and include these values as regular symbols? The new type can be used to know they are immediate values, so separating them by file might not be necessary anymore.

UsernameFodder (commented):
I'm not a fan of this change. The reason I cordoned all the immediate values to their own separately symbol table file is so that they wouldn't appear in the generated symbol tables, since they aren't "real"/valid symbols that any code analysis tool (Ghidra, No$ debugger) knows how to deal with properly.

Semantically, it doesn't make sense to label these as a "C type" at all. I don't think any compiler would see these declarations and generate valid code if you tried to reference the actual data they link to. So that would have negative implications for c-of-time (and any other project that relies on the C headers being _valid_), for example.

I can't think of a way to achieve this kind of labeling that isn't fragile. The representation of any given immediate value is instruction-dependent, isn't it? An immediate for a `mov` instruction might be encoded in a different way than that for an `add` or a `ldr` instruction. Does your code just assume they're all `mov`s?

For your changes, is it not possible to just assume every symbol defined in `literals.yml` is a shifted immediate type, and use that to implicitly interpret those addresses as the `sh_imm` type without needing to bake that into the C headers?

End45:
> I'm not a fan of this change. The reason I cordoned all the immediate values to their own separately symbol table file is so that they wouldn't appear in the generated symbol tables, since they aren't "real"/valid symbols that any code analysis tool (Ghidra, No$ debugger) knows how to deal with properly.

> Semantically, it doesn't make sense to label these as a "C type" at all. I don't think any compiler would see these declarations and generate valid code if you tried to reference the actual data they link to. So that would have negative implications for c-of-time (and any other project that relies on the C headers being valid), for example.

Why do you consider the symbols to be invalid? They are embedded into an instruction, but they are still valid 2-byte data (or, well, 1.5-byte data, I guess). Does anything really break just for defining these?

> I can't think of a way to achieve this kind of labeling that isn't fragile. The representation of any given immediate value is instruction-dependent, isn't it? An immediate for a mov instruction might be encoded in a different way than that for an add or a ldr instruction. Does your code just assume they're all movs?

I don't think so. All instructions that have an immediate as their value use the last 3 nibbles to encode it. That includes `mov`, `add`, `sub`, `tst`, etc. Instructions that use immediates for addressing (such as `ldr rX, [rY, offset]`) don't use that format, but the immediate there does not carry any meaningful data.

> For your changes, is it not possible to just assume every symbol defined in literals.yml is a shifted immediate type, and use that to implicitly interpret those addresses as the sh_imm type without needing to bake that into the C headers?

That's the other possibility, yes. I thought about adding the info to the C headers in case it was helpful, but if it's going to cause more harm than good, I can just implement this at the pmdsky-debug-py level. I wouldn't mind.

UsernameFodder:
> > I can't think of a way to achieve this kind of labeling that isn't fragile. The representation of any given immediate value is instruction-dependent, isn't it? An immediate for a mov instruction might be encoded in a different way than that for an add or a ldr instruction. Does your code just assume they're all movs?
> 
> I don't think so. All instructions that have an immediate as their value use the last 3 nibbles to encode it. That includes `mov`, `add`, `sub`, `tst`, etc. Instructions that use immediates for addressing (such as `ldr rX, [rY, offset]`) don't use that format, but the immediate there does not carry any meaningful data.

Hmm, I guess this is more uniform than I thought. The manual classifies these all as "data processing instructions" that use the same encoding scheme for all fields, not just the immediate value.

> Why do you consider the symbols to be invalid? They are embedded into an instruction, but they are still valid 2-byte data (or, well, 1.5-byte data, I guess). Does anything really break just for defining these?

Well, 1.5 bytes is very different from 2 bytes. The latter can be mapped to a C `short`, whereas the former needs to be encoded more specially.

I think I would be okay with this if we defined the type to more precisely map to the actual machine code bytes. And we should declare all instructions as `const`. Something like this:

```c
// A raw ARMv5 data-processing instruction, such as MOV, ADD, AND, CMP, etc.
// See the ARMv5 Architecture Reference Manual, Section A3.4.1
// https://developer.arm.com/documentation/ddi0100/latest/
struct data_processing_instruction {
    // second source operand, either a shifted immediate value or a register, see Section A5.1
    uint32_t shifter_operand : 12;
    uint32_t rd : 4; // destination register
    uint32_t rn : 4; // first source operand register
    uint32_t s : 1; // status flag, set if the instruction updates the status registers
    uint32_t opcode : 4; // see Section A3.4, Table A3-2
    uint32_t i : 1; // immediate flag, set if shifter_operand represents an immediate
    uint32_t _zero : 2; // always 0
    uint32_t cond : 4; // condition code, see Section A3.2
};
ASSERT_SIZE(struct data_processing_instruction, 4);

extern const struct data_processing_instruction JUICE_BAR_NECTAR_IQ_GAIN;
...
```

We should also go over all the addresses defined in `literals.yml` and make sure they correspond to the offset of the start of the _instruction_, and not one of the bytes halfway through. Though, given that all the addresses are 4-byte aligned, I suspect this is already the case? Probably because `shifter_operand` happens to come first. We should also make sure the lengths are specified as 4 bytes, since we'll be typing them with a 4-byte struct. I see that `JUICE_BAR_NECTAR_IQ_GAIN` has a length of 1, so we should update that.

End45:
> I think I would be okay with this if we defined the type to more precisely map to the actual machine code bytes.
That could work. Although the symbol wouldn't be 100% accurate anymore. For example, `JUICE_BAR_NECTAR_IQ_GAIN` no longer refers to the immediate value on that address, but rather to the whole instruction. Does it make sense to document that as a symbol?

> And we should declare all instructions as `const`.
Well, technically, they can be modified while in RAM. No one is stopping you from doing that.

> We should also go over all the addresses defined in literals.yml and make sure they correspond to the offset of the start of the instruction, and not one of the bytes halfway through. Though, given that all the addresses are 4-byte aligned, I suspect this is already the case? Probably because shifter_operand happens to come first.
Yeah, the operand comes first, so I don't think any changes are needed here.

> I see that JUICE_BAR_NECTAR_IQ_GAIN has a length of 1, so we should update that.
Huh, you're right. It's the only literal that has a lenght field. That should be changed.

UsernameFodder:
> Although the symbol wouldn't be 100% accurate anymore. For example, JUICE_BAR_NECTAR_IQ_GAIN no longer refers to the immediate value on that address, but rather to the whole instruction. Does it make sense to document that as a symbol?

I think this is fine. It's already conveyed that these are _literals_ anyway, not normal symbols. And since the value can involve a fractional number of bytes, trying to convey the exact location of the immediate value isn't possible in all cases anyway. The type will be documented in the C headers.

> Well, technically, they can be modified while in RAM. No one is stopping you from doing that.

Well, by this logic, C shouldn't have a `const` keyword. Anything can be modified in RAM, but that doesn't mean they _are_, and doesn't mean a hacker _should_ modify them in most cases. Labeling something as `const` doesn't actually do anything at the machine code level, it's purely a construct for the compiler to enforce restrictions on what C code is allowable. In these C headers, I think it's mainly useful as documentation. These instruction offsets aren't "variables" in the usual sense.

End45:
Alright, I'm okay with this implementation then. I'll change it when I have some time.

End45:
Alright, the new implementation is ready. The synchronization script needs to be updated to account for the fact that symbols might be defined in literals.yml reather than the yml file corresponding to the current binary.
The sort order check might need to be changed too.

headers/data/arm9.h
```diff
@@ -196,5 +196,10 @@ extern int32_t MUSIC_DURATION_LOOKUP_TABLE_2[128];
 extern sound_lfo_waveform_callback LFO_WAVEFORM_CALLBACKS[16];
 extern bool IS_DISP_ON;
 extern uint32_t GXI_DMA_ID;
+extern struct data_processing_instruction JUICE_BAR_NECTAR_IQ_GAIN;
```
UsernameFodder, line 199: Can we make these `const`?

headers/data/overlay00.h
```diff
@@ -0,0 +1,6 @@
+#ifndef HEADERS_DATA_OVERLAY00_H_
```
UsernameFodder, line 1: I think we should put these in a new header file, `headers/data/literals.h` to match the structure of the YAML files. I'm pretty sure the test scripts should "just work" (or at least only require minor changes) if we do this.

UsernameFodder (changes requested):

headers/data/arm9.h
End45: Right. I knew I was forgetting someting.

headers/data/overlay00.h
End45: Not sure if I like that approach. I think it has multiple downsides:

- It's not possible to know what binary each symbol belongs to just by looking at the header file
- It's very hard to know what the sort order for new symbols should be just by looking at the header file
- Symbols in the header file would be sorted by their addresses, even though they belong to different binaries. You could get completely unrelated symbols next to each other, multple related symbols split into separate groups, two unrelated symbols with the same address, etc.
- This creates yet another .sym file to import for the runtime debugging setup. It also means that you're forced to choose between importing all literals or none of them, regardless of which binary they belong to. Multiple literals with the same address would also be a problem in this scenario.

I think separating them by binary on the header files works better.

UsernameFodder: > * It's not possible to know what binary each symbol belongs to just by looking at the header file
> 
> * It's very hard to know what the sort order for new symbols should be just by looking at the header file

I feel like this could be solved with comments. We could add comments in `literals.h` labeling each binary.

> * Symbols in the header file would be sorted by their addresses, even though they belong to different binaries. You could get completely unrelated symbols next to each other, multple related symbols split into separate groups, two unrelated symbols with the same address, etc.

This isn't how `symbol_check.py` works. When it parses the YAML files it preserves the order of blocks. So all the arm9 symbols will come first in address order, then the overlay0 symbols, etc.

>     * This creates yet another .sym file to import for the runtime debugging setup. It also means that you're forced to choose between importing all literals or none of them, regardless of which binary they belong to. Multiple literals with the same address would also be a problem in this scenario.

Not really. This is entirely controlled by the YAML files. The header files have nothing to do with it.

> I think separating them by binary on the header files works better.

I would be okay with this approach if all the literals were put in their own section at the end of the header file. But that would require larger changes to `symbol_check.py` that I'm not really keen on making.

End45:
It seems like sync-check follows strictly alphabetical order rather than numeric order for overlays. That should probably be changed.

headers/data/literals.h
```diff
@@ -0,0 +1,20 @@
+#ifndef HEADERS_DATA_LITERALS_H_
+#define HEADERS_DATA_LITERALS_H_
+
+// Arm9
```
UsernameFodder, line 4: ARM is all caps.

```suggestion
// ARM9
```

UsernameFodder (changes requested):

UsernameFodder:
> It seems like sync-check follows strictly alphabetical order rather than numeric order for overlays. That should probably be changed.

It's mainly because the blocks are named, e.g., `overlay0` rather than `overlay00`. I've been worried about changing it and possibly breaking things that rely on the old names, so I haven't gotten around to it. I'm not sure it's worth changing it just for the sake of the literals table anyway.

headers/data/literals.h
End45: 👍

End45:
Isn't it possible to just change the sorting algorithm without renaming them?

UsernameFodder:
> Isn't it possible to just change the sorting algorithm without renaming them?

No. The sort order comes from the YAML, and the YAML sorting needs to work for arbitrary strings.

End45:
> > Isn't it possible to just change the sorting algorithm without renaming them?
> 
> No. The sort order comes from the YAML, and the YAML sorting needs to work for arbitrary strings.

I see. Oh well, I'll just reorder it then. Not like reordering them again would be an issue should you ever change the naming convention.

UsernameFodder (approved):
Looks good now, thanks!

PR #259, opened 2024-06-01
Add flag to mark aliases as deprecated to header augmentation script

tech-ticks:
Adds a new flag `--mark_aliases_deprecated` to `augment_headers.py` that generates warnings such as:

```
c-of-time/src/main.c:17:3: warning: 'ProcessScriptParam' is deprecated: Renamed to 'ScriptParamToInt' [-Wdeprecated-declarations]
   17 |   int x = ProcessScriptParam(123);
      |   ^~~
```

I'm opening this PR as a draft because the formatter generates very weird looking code:
```c
__attribute((deprecated("Renamed to 'ScriptParamToInt'"))) int16_t
// THIS DOCSTRING WAS GENERATED AUTOMATICALLY
/**
 * Converts the given opcode parameter to a signed integer.
 * 
 * The parameter will be returned unchanged unless one of its two most significant bits (0x8000 and 0x4000) are set, in which case both bits will be cleared and the original value will be modified according to the following two rules:
 * - If the 0x4000 bit is set (sign bit), the value will be set to -16384 + value.
 * - If the 0x8000 bit is set (fixed-point flag), the value will be set to value / 256, rounded down.
 * Both rules can be applied, in the same order as listed, if both conditions are met.
 * 
 * r0: Parameter to convert
 * return: The input parameter, as a signed integer
 */
ProcessScriptParam(uint16_t parameter);
```

This issue doesn't seem to occur if I set the formatter to `None`, but I don't have enough insight into how it works to fix it.

### DIFF
--- a/headers/augment_headers.py
+++ b/headers/augment_headers.py
@@ -160,7 +160,7 @@ class HeaderAugmenter:
             if in_c_style_comment and "*/" in line:
                 in_c_style_comment = False
 
-    def add_aliases(self) -> int:
+    def add_aliases(self, mark_as_deprecated: bool) -> int:
         add_count = 0
         symbol_names = self.load_symbol_names()
         with open(self.input_header_file(), "r") as f:
@@ -203,6 +203,9 @@ class HeaderAugmenter:
                     if ";" in line.split("//", 1)[0]:
                         # Symbol declaration is finished; add aliases
                         for alias in aliases:
+                            if mark_as_deprecated:
+                                f.write(f"__attribute((deprecated(\"Renamed to '{aliased_symbol}'\"))) ")
+
                             f.write(
                                 symbol_declaration[0].replace(aliased_symbol, alias, 1)
                             )
@@ -253,6 +256,7 @@ def add_header_content(
     extension: str,
     *,
     aliases: bool = True,
+    mark_aliases_as_deprecated: bool = True,
     docstrings: bool = True,
     formatter: Optional[Formatter] = None,
     filter: Optional[str] = None,
@@ -270,7 +274,7 @@ def add_header_content(
             continue
 
         if aliases:
-            count = augmenter.add_aliases()
+            count = augmenter.add_aliases(mark_aliases_as_deprecated)
             if verbosity >= 2 or (verbosity >= 1 and count > 0):
                 print(f"Added {count} alias declaration(s) to {header_file}")
         if docstrings:
@@ -288,6 +292,13 @@ if __name__ == "__main__":
         "--aliases",
         action="store_true",
         help="add alias declarations",
+    )
+    parser.add_argument(
+        "-ad",
+        "--mark_aliases_deprecated",
+        action="store_true",
+        help="mark aliases as deprecated (if --aliases is set)",
+        default=True
     )
     parser.add_argument(
         "-d",
@@ -328,6 +339,7 @@ if __name__ == "__main__":
         FunctionList,
         args.extension,
         aliases=args.aliases,
+        mark_aliases_as_deprecated=args.mark_aliases_deprecated,
         docstrings=args.docstrings,
         formatter=formatter,
         filter=args.filter,
@@ -337,6 +349,7 @@ if __name__ == "__main__":
         DataList,
         args.extension,
         aliases=args.aliases,
+        mark_aliases_as_deprecated=args.mark_aliases_deprecated,
         docstrings=args.docstrings,
         formatter=formatter,
         filter=args.filter,
--- a/headers/augment_headers.py
+++ b/headers/augment_headers.py
@@ -136,9 +136,10 @@ class HeaderAugmenter:
         return set(self.slist.names_from_c_header(self.input_header_file()))
 
     @staticmethod
-    def _input_header_lines(lines) -> Iterator[Tuple[str, bool, bool]]:
+    def _input_header_lines(lines) -> Iterator[Tuple[str, bool, bool, bool]]:
         in_docstring = False
         in_c_style_comment = False
+        is_attribute = False
         for line in lines:
             in_cpp_style_comment = False
             if line == HeaderAugmenter.DOCSTRING_PREAMBLE_LINE:
@@ -155,7 +156,9 @@ class HeaderAugmenter:
             if not in_comment:
                 in_docstring = False
 
-            yield line, in_comment, in_docstring
+            is_attribute = not in_comment and line.startswith("__attribute__((")
+
+            yield line, in_comment, in_docstring, is_attribute
 
             if in_c_style_comment and "*/" in line:
                 in_c_style_comment = False
@@ -174,7 +177,11 @@ class HeaderAugmenter:
             ignored_symbol = None
             prev_aliases = []
 
-            for line, in_comment, _in_docstring in self._input_header_lines(lines):
+            for line, in_comment, _in_docstring, is_attribute in self._input_header_lines(lines):
+                if is_attribute:
+                    # Skip existing attributes (we ensure that line with attributes don't contain symbol names)
+                    continue
+
                 if not in_comment and aliased_symbol is None and ignored_symbol is None:
                     match = self.slist.NAME_REGEX.search(line)
                     if match:
@@ -204,7 +211,9 @@ class HeaderAugmenter:
                         # Symbol declaration is finished; add aliases
                         for alias in aliases:
                             if mark_as_deprecated:
-                                f.write(f"__attribute((deprecated(\"Renamed to '{aliased_symbol}'\"))) ")
+                                # Attributes should end with a comment and newline; this is a somewhat hacky workaround
+                                # to ensure that attributes are not matched by `NAME_REGEX`.
+                                f.write(f"__attribute__((deprecated(\"Renamed to '{aliased_symbol}'\"))) // alias\n")
 
                             f.write(
                                 symbol_declaration[0].replace(aliased_symbol, alias, 1)
@@ -226,10 +235,15 @@ class HeaderAugmenter:
     def add_docstrings(self) -> int:
         add_count = 0
         symbol_names = self.load_symbol_names()
+        buffered_attribute = None
         with open(self.input_header_file(), "r") as f:
             lines = f.readlines()
         with open(self.output_header_file(), "w") as f:
-            for line, in_comment, in_docstring in self._input_header_lines(lines):
+            for line, in_comment, in_docstring, is_attribute in self._input_header_lines(lines):
+                if is_attribute:
+                    # Write attributes after docstrings
+                    buffered_attribute = line
+                    continue
                 if in_docstring:
                     # Skip over docstrings
                     continue
@@ -245,7 +259,13 @@ class HeaderAugmenter:
                             if docstring is not None:
                                 f.write(HeaderAugmenter.DOCSTRING_PREAMBLE_LINE)
                                 f.write(docstring)
+
+                                if buffered_attribute:
+                                    f.write(buffered_attribute)
+                                    buffered_attribute = None
+
                                 add_count += 1
+
                 f.write(line)
         self.commit_header_file()
         return add_count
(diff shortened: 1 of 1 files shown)


headers/augment_headers.py
```diff
@@ -289,6 +293,13 @@ def add_header_content(
         action="store_true",
         help="add alias declarations",
     )
+    parser.add_argument(
+        "-ad",
```
UsernameFodder, line 297: I didn't know you could have two-character short flags? Won't this interfere with the usual meaning of "-ad" which is "-a -d"? Let's just remove this and only leave the long name, argparse can handle that.

headers/augment_headers.py
```diff
@@ -289,6 +293,13 @@ def add_header_content(
...
+        "--mark_aliases_deprecated",
```
UsernameFodder, line 298: This name seems a bit long. Also I usually like CLI flags to be dash-separated, not underscore-separated. How about `--deprecate-aliases`?

headers/augment_headers.py
```diff
@@ -289,6 +293,13 @@ def add_header_content(
...
+        action="store_true",
+        help="mark aliases as deprecated (if --aliases is set)",
+        default=True
```
UsernameFodder, line 301: I'd rather we not set a default (i.e., default to None/False). The syntax used isn't actually part of standard C, it's a GCC extension. Also, I think if you have a `store_true` flag that defaults to true, there's no way to actually turn it off?

headers/augment_headers.py
```diff
@@ -289,6 +293,13 @@ def add_header_content(
         action="store_true",
         help="add alias declarations",
     )
+    parser.add_argument(
+        "-ad",
+        "--mark_aliases_deprecated",
+        action="store_true",
+        help="mark aliases as deprecated (if --aliases is set)",
```
UsernameFodder, line 300: Let's mention the syntax specifically:

```suggestion
        help="mark aliases with __attribute__((deprecated)) if --aliases is set",
```

headers/augment_headers.py
```diff
@@ -203,6 +203,9 @@ def add_aliases(self) -> int:
                     if ";" in line.split("//", 1)[0]:
                         # Symbol declaration is finished; add aliases
                         for alias in aliases:
+                            if mark_as_deprecated:
+                                f.write(f"__attribute((deprecated(\"Renamed to '{aliased_symbol}'\"))) ")
```
UsernameFodder, line 207: I'm not entirely sure why `__attribute` works, but I believe the official syntax has both leading and trailing underscores, i.e. `__attribute__((deprecated(...)))`

UsernameFodder (changes requested):

UsernameFodder:
> I'm opening this PR as a draft because the formatter generates very weird looking code:

Hmm, I see the problem. There's actually multiple interacting problems 😅.

The augmenter works in multiple passes. The first pass goes in and adds aliases by essentially copy-pasting aliased functions. The second pass prepends docstrings.

Both passes process files line-by-line, searching for symbol names with a regex. Whenever content is added, the pass will just inject extra lines into the file stream. After each pass, the formatter is run.

The function regex is:
```py
    # Look for words preceding an open single parenthesis. Not perfect, but
    # good enough. This should work for everything except functions
    # with function pointer parameters, but function pointer parameters
    # should really be typedef'd for readability anyway.
    NAME_REGEX = re.compile(r"\b(\w+)\s*\([^(]")
```

So here's the problem(s): First, aliases are added. With your changes, an alias declaration might look something like this:

```c
__attribute__((deprecated("Renamed to 'DataTransferInit'"))) void FileRom_InitDataTransfer(void);
```

Or if the line is too long, the formatter will break the line:

```c
__attribute__((deprecated("Renamed to 'FileInitVeneer'"))) void
FileRom_Veneer_FileInit(struct file_stream* file);
```

Then docstrings are added. In the first case, the presence of the `__attribute__(` and `deprecated(` fool the regex, so it doesn't find the right function name, and no docstring is written at all. In the second case, the function name is found, but the docstring is inserted on the line before it, which ends up in between the attribute specifier and the actual function name. So it's broken in both cases.

So I think a correct implementation might take a tad more work. Maybe adding a third pass at the end? It seems sort of tricky to get right; I'd have to think about it a bit more.

UsernameFodder:
So right now the script is (mostly) idempotent (you can run it multiple times repeatedly without breaking the headers horribly), which is a property I'd like to keep in case somebody tries to run the script again by accident. That makes things a bit hard, but I think I thought of a reasonable (somewhat hacky) approach that might work:

- Always write `__attribute__(deprecated(...))) // alias` on its own line to make it easier to parse without interfering with the existing regexes. Adding a short comment at the end of the line is a bit hacky, but it ensures the formatter keeps the attribute specifier on its own line.
- Add another output to `_input_header_lines` called `is_attribute`, which should be pretty easy to look for; if we're not in a comment, and the line starts with `__attribute__((`, then the line is an attribute.
- In `add_aliases`, skip any existing line that's an attribute (also skip scanning the line for symbol names), and insert the deprecated attribute on its own line when injecting a new alias declaration.
- In `add_docstrings`, if we see an attribute line, buffer it until we see the next line that matches a symbol name, and then write out the attribute line right after writing the docstring.

I think this approach should ensure that attribute lines are written in the write place, and that they don't get duplicated if you run the script multiple times.

tech-ticks:
Sorry for the late reply. I tried to implement your workaround, but with the formatter enabled I now end up with:

```c
int16_t
    // THIS DOCSTRING WAS GENERATED AUTOMATICALLY
    /**
     * Converts the given opcode parameter to a signed integer.
     *
     * The parameter will be returned unchanged unless one of its two most significant bits (0x8000
     * and 0x4000) are set, in which case both bits will be cleared and the original value will be
     * modified according to the following two rules:
     * - If the 0x4000 bit is set (sign bit), the value will be set to -16384 + value.
     * - If the 0x8000 bit is set (fixed-point flag), the value will be set to value / 256, rounded
     * down. Both rules can be applied, in the same order as listed, if both conditions are met.
     *
     * r0: Parameter to convert
     * return: The input parameter, as a signed integer
     */
    __attribute__((deprecated("Renamed to 'ScriptParamToInt'"))) // alias
    ProcessScriptParam(uint16_t parameter);
  ```

It only seems to separate the type like this if the formatter is enabled. Not sure why this happens in the first place, I'd have to wrap my head around how exactly the formatter works.

**Comment by UsernameFodder** [comment]
> Sorry for the late reply. I tried to implement your workaround, but with the formatter enabled I now end up with:
> 
> ```c
> int16_t
>     // THIS DOCSTRING WAS GENERATED AUTOMATICALLY
>     /**
>      * Converts the given opcode parameter to a signed integer.
>      *
>      * The parameter will be returned unchanged unless one of its two most significant bits (0x8000
>      * and 0x4000) are set, in which case both bits will be cleared and the original value will be
>      * modified according to the following two rules:
>      * - If the 0x4000 bit is set (sign bit), the value will be set to -16384 + value.
>      * - If the 0x8000 bit is set (fixed-point flag), the value will be set to value / 256, rounded
>      * down. Both rules can be applied, in the same order as listed, if both conditions are met.
>      *
>      * r0: Parameter to convert
>      * return: The input parameter, as a signed integer
>      */
>     __attribute__((deprecated("Renamed to 'ScriptParamToInt'"))) // alias
>     ProcessScriptParam(uint16_t parameter);
> ```
> 
> It only seems to separate the type like this if the formatter is enabled. Not sure why this happens in the first place, I'd have to wrap my head around how exactly the formatter works.

Ugh, it appears that this is a bug in clang-format (or GCC attributes are just unsupported). I found [this SO post](https://stackoverflow.com/questions/76898417/make-clang-format-break-after-attribute) where someone was trying to do exactly what we are, and there's also some open GitHub issues for clang-format about it. One of the comments mentions a workaround where you can define a macro. I tested it and this seems to work with the formatter on my machine:

```c
#ifndef DEPRECATED
#define DEPRECATED(name) __attribute__((deprecated("Renamed to '" name "'")))
#endif

DEPRECATED("ScriptParamToInt")
int16_t ProcessScriptParam(uint16_t parameter);
```

It's a little annoying, but I guess we can have the script insert the macro in every header file right after the include guard (I suggest matching with the regex `#define HEADERS_[A-Z0-9_]+_H_?` to be somewhat robust). And play the same game with it for idempotence: inject it, and then skip over any copies from previous runs (anything between `#ifndef DEPRECATED` and the next `#endif`).

tech-ticks:
Alright thanks, that seemed to fix the issue.

headers/augment_headers.py
```diff
@@ -96,6 +96,7 @@ def format_file(self, filename: str):
 class HeaderAugmenter:
     DOCSTRING_PREAMBLE_LINE = "// THIS DOCSTRING WAS GENERATED AUTOMATICALLY\n"
     CPP_COMMENT_RE = re.compile(r"\s*//")
+    DEPRECATED_MACRO_GUARD_RE = re.compile(r"#define HEADERS_[A-Z0-9_]+_H_?")
```
UsernameFodder, line 99: Let's call this `INCLUDE_GUARD_RE`. It's not matching against the DEPRECATED macro, it's matching against the normal include guard.

headers/augment_headers.py
```diff
@@ -128,17 +129,19 @@ def input_header_file(self):
     def output_header_file(self):
         return self.slist.header_file + self.out_extension
 
-    def commit_header_file(self):
-        self.formatter.format_file(self.output_header_file())
+    def commit_header_file(self, format = True):
```
UsernameFodder, line 132: `format` needs a type annotation:

```suggestion
    def commit_header_file(self, format: bool = True):
```

headers/augment_headers.py
```diff
@@ -214,6 +226,37 @@ def add_aliases(self) -> int:
                         symbol_declaration = []
         self.commit_header_file()
         return add_count
+    
+    def add_deprecated_macro(self):
+        # Adds the `DEPRECATED(message)` macro to the header file if it doesn't exist yet.
```
UsernameFodder, line 231: Let's make this an actual triple-quote docstring.

headers/augment_headers.py
```diff
@@ -155,12 +158,14 @@ def _input_header_lines(lines) -> Iterator[Tuple[str, bool, bool]]:
             if not in_comment:
                 in_docstring = False
 
-            yield line, in_comment, in_docstring
+            is_deprecated_macro = not in_comment and line.startswith("DEPRECATED(")
```
UsernameFodder, line 161: Let's `lstrip` the line of leading whitespace before checking for `DEPRECATED(`

```suggestion
            is_deprecated_macro = not in_comment and line.lstrip().startswith("DEPRECATED(")
```

headers/augment_headers.py
```diff
@@ -155,12 +158,14 @@ def _input_header_lines(lines) -> Iterator[Tuple[str, bool, bool]]:
...
+
+            yield line, in_comment, in_docstring, is_deprecated_macro
```
UsernameFodder, line 163: I know this is getting a bit unwieldy, but can we also add an `in_directive` check for preprocessor directives? Otherwise we'll try to match the `#define DEPRECATED(` line against the function regex and it'll probably yield the false symbol name "DEPRECATED". So something like

```py
in_directive = not in_comment and line.lstrip().startswith("#")
is_deprecated_macro = not in_comment and line.lstrip().startswith("DEPRECATED(")

yield line, in_comment, in_docstring, in_directive, is_deprecated_macro
```

headers/augment_headers.py
UsernameFodder, line 231:
```suggestion
        # Adds the `DEPRECATED(name)` macro to the header file if it doesn't exist yet.
```

headers/augment_headers.py
```diff
@@ -214,6 +226,37 @@ def add_aliases(self) -> int:
...
+        with open(self.output_header_file(), "w") as f:
+            in_guard = False
+            for line in lines:
+                if line.strip() == "#ifndef DEPRECATED":
+                    in_guard = True
+
+                if not in_guard:
+                    f.write(line)
+                
+                if line.strip() == "#endif" and in_guard:
```
UsernameFodder, line 253: Let's put `in_guard` first so we don't need to do string operations unless we're actually within the guard.

headers/augment_headers.py
UsernameFodder, line 99: Hmm, just to be safe let's add a `\s*` at the beginning, like with `CPP_COMMENT_RE`.

UsernameFodder (commented):
Glad this seems to be functional now. Just some minor comments.

UsernameFodder (approved):
Looks good now, thanks!

PR #264, opened 2024-07-26
Document sprite data struct

Frostbyte0x70:


headers/data/arm9.h
```diff
@@ -121,7 +121,7 @@ extern struct guest_monster GUEST_MONSTER_DATA[18];
 extern struct rankup_table_entry RANK_UP_TABLE[13];
 extern struct monster_id_16 DS_DOWNLOAD_TEAMS[56];
 extern enum monster_id UNOWN_SPECIES_ADDITIONAL_CHARS[28];
-extern undefined MONSTER_SPRITE_DATA[1200];
+extern struct monster_sprite_data_entry MONSTER_SPRITE_DATA[1200];
```
UsernameFodder, line 124: Shouldn't the array have length 600?

UsernameFodder (changes requested):

headers/data/arm9.h
Frostbyte0x70: Oh, right. The game does do a mod 600 operation before reading from the array. Looks like that value was wrong.

UsernameFodder: It wasn't wrong, it was defining the length in bytes before. Now the struct is 2 bytes.

UsernameFodder (approved):

PR #265, opened 2024-08-14
Document Perish Song turn range

Frostbyte0x70:
Also documents DungeonRandRange's special behavior when the upper and lower range is the same (such as when Perish Song is used).


symbols/overlay10.yml
```diff
@@ -1497,6 +1497,18 @@ overlay10:
       description: |-
         The turn range for the Whiffer status inflicted by Smokescreen, [1, 4).
         
+        type: int16_t[2]
+    - name: PERISH_SONG_TURN_RANGE
+      address:
+        EU: 0x22C5128
+        NA: 0x22C47D0
+        JP: 0x22C5EB8
+      length:
```
UsernameFodder, line 1506: Should there be an EU length?

Frostbyte0x70: Right. I copied it from above. Strange that none of the other fields have it.

UsernameFodder (approved):

PR #266, opened 2024-08-22
Some new functions related to dungeon AI behavior

assidion2:
First commit. Hopefully this is acceptable?

### DIFF
--- a/headers/functions/overlay29.h
+++ b/headers/functions/overlay29.h
@@ -274,7 +274,9 @@ bool CanAiMonsterMoveInDirection(struct entity* monster, enum direction_id direc
                                  bool* out_monster_in_target_position);
 bool ShouldMonsterRunAway(struct entity* monster);
 bool ShouldMonsterRunAwayVariation(struct entity* monster, undefined param_2);
-enum monster_relation GetRelationBetweenMonsters(struct entity* entity1, struct entity* entity2, bool param_3, bool param_4);
+enum monster_relation GetRelationBetweenMonsters(struct entity* entity1, struct entity* entity2,
+                                                 bool see_invisible_targets,
+                                                 bool ignore_petrified_targets);
 bool SafeguardIsActive(struct entity* user, struct entity* target, bool log_message);
 bool LeafGuardIsActive(struct entity* user, struct entity* target, bool log_message);
 bool IsProtectedFromStatDrops(struct entity* user, struct entity* target, bool log_message);
--- a/headers/functions/overlay29.h
+++ b/headers/functions/overlay29.h
@@ -867,5 +867,8 @@ void DisplayMessageInternal(int message_id, bool wait_for_input, struct portrait
 void OpenMenu(undefined param_1, undefined param_2, bool open_bag);
 int OthersMenuLoop(void);
 undefined OthersMenu(void);
+int GetAiUseItemProbability(struct entity* item_consumer, struct item* used_item, int unknown);
+bool IsAdjacentToEnemy(struct entity* entity);
+enum relation GetRelationBetweenMonsters(struct entity* entity_1, struct entity* entity_2, bool unknown_1, bool unknown_2);
 
 #endif
--- a/headers/functions/overlay29.h
+++ b/headers/functions/overlay29.h
@@ -275,8 +275,8 @@ bool CanAiMonsterMoveInDirection(struct entity* monster, enum direction_id direc
 bool ShouldMonsterRunAway(struct entity* monster);
 bool ShouldMonsterRunAwayVariation(struct entity* monster, undefined param_2);
 enum monster_treatment GetTreatmentBetweenMonsters(struct entity* entity1, struct entity* entity2,
-                                                 bool see_invisible_targets,
-                                                 bool ignore_petrified_targets);
+                                                   bool see_invisible_targets,
+                                                   bool ignore_petrified_targets);
 bool SafeguardIsActive(struct entity* user, struct entity* target, bool log_message);
 bool LeafGuardIsActive(struct entity* user, struct entity* target, bool log_message);
 bool IsProtectedFromStatDrops(struct entity* user, struct entity* target, bool log_message);
--- a/headers/functions/overlay29.h
+++ b/headers/functions/overlay29.h
@@ -274,7 +274,7 @@ bool CanAiMonsterMoveInDirection(struct entity* monster, enum direction_id direc
                                  bool* out_monster_in_target_position);
 bool ShouldMonsterRunAway(struct entity* monster);
 bool ShouldMonsterRunAwayVariation(struct entity* monster, undefined param_2);
-enum monster_relation GetRelationBetweenMonsters(struct entity* entity1, struct entity* entity2,
+enum monster_treatment GetTreatmentBetweenMonsters(struct entity* entity1, struct entity* entity2,
                                                  bool see_invisible_targets,
                                                  bool ignore_petrified_targets);
 bool SafeguardIsActive(struct entity* user, struct entity* target, bool log_message);
--- a/headers/functions/overlay29.h
+++ b/headers/functions/overlay29.h
@@ -274,6 +274,7 @@ bool CanAiMonsterMoveInDirection(struct entity* monster, enum direction_id direc
                                  bool* out_monster_in_target_position);
 bool ShouldMonsterRunAway(struct entity* monster);
 bool ShouldMonsterRunAwayVariation(struct entity* monster, undefined param_2);
+enum monster_relation GetRelationBetweenMonsters(struct entity* entity1, struct entity* entity2, bool param_3, bool param_4);
 bool SafeguardIsActive(struct entity* user, struct entity* target, bool log_message);
 bool LeafGuardIsActive(struct entity* user, struct entity* target, bool log_message);
 bool IsProtectedFromStatDrops(struct entity* user, struct entity* target, bool log_message);
@@ -576,6 +577,8 @@ void ApplyViaSeedEffect(struct entity* user, struct entity* target);
 void ApplyGravelyrockEffect(struct entity* user, struct entity* target);
 void ApplyGonePebbleEffect(struct entity* user, struct entity* target);
 void ApplyGracideaEffect(struct entity* user, struct entity* target);
+int GetAiUseItemProbability(struct entity* item_consumer, struct item* item, uint32_t flags);
+bool IsAdjacentToEnemy(struct entity* entity);
 bool ShouldTryEatItem(enum item_id item_id);
 int GetMaxPpWrapper(struct move* move);
 bool MoveIsNotPhysical(enum move_id move_id);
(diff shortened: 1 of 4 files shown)


headers/functions/overlay29.h
```diff
@@ -867,5 +867,8 @@ void DisplayMessageInternal(int message_id, bool wait_for_input, struct portrait
 void OpenMenu(undefined param_1, undefined param_2, bool open_bag);
 int OthersMenuLoop(void);
 undefined OthersMenu(void);
+int GetAiUseItemProbability(struct entity* item_consumer, struct item* used_item, int unknown);
```
Frostbyte0x70, line 870: > struct item* used_item

I think `used_item` is a bit misleading. It seems to imply that the item has already been used. Why not just name it `item`?

> int unknown

I think that parameter should either be documented as a bitfield or specified as `undefined`, since saying that it's an int is not 100% true.

symbols/overlay29.yml
```diff
@@ -8500,6 +8500,44 @@ overlay29:
...
+        EU: 0x2302188
+        NA: 0x230157C
+        JP: 0x23027FC
+      description: |-
+        Called to check if two monsters are allies, enemies, or neither (one is a shopkeeper, decoy, etc.)
+
+        r0: Pointer to entity 1
+        r1: Pointer to entity 2
+        r2: ? (If == 0, will make function return 2 if r1 is invisible and r0 can see invisible monsters)
+        r3: ? (If != 0, will make function return 2 if r0 is a team member and r1 has some sort of status)
```
Frostbyte0x70, line 8538:
```suggestion
        r2: ? (If false, will make function return RELATION_OTHER if entity 2 is invisible and entity 1 can see invisible monsters)
        r3: ? (If true, will make function return RELATION_OTHER if entity 1 is a team member and entity 2 has the STATUS_PETRIFIED status)
```

Frostbyte0x70 (changes requested):
Hi! Thanks for taking the time to contribute. I only have a couple of minor suggestions, see below.

Make sure to take a look at the failing checks too. In particular,
- overlay29.yml seems to be underindented starting on line 8517
- A couple of lines in overlay29.h are too long (the maximum is 100 characters)

symbols/overlay29.yml
```diff
@@ -8500,6 +8500,44 @@ overlay29:
...
+      address:
+        EU: 0x231EAC4
+        NA: 0x231DEB8
+        JP: 0x231F138
+      description: |-
+        Called to get the probability of an item being used or thrown by an AI on the current turn.
+
+        r0: Pointer to either the user if it is an item used by the AI or the target if it is an item thrown by the AI
+        r1: Pointer to item
+        r2: Unknown 2-bit bitvector. Bit 1 influences only the Warp Seed case, while Bit 0 influences various items.
```
AnonymousRandomPerson, line 8513: In the [equivalent function in Rescue Team](https://github.com/pret/pmd-red/blob/master/src/dungeon_ai_item_weight.c#L14), r2 represents the AI's current target int context of deciding the item probability.
* If bit 0 is set, the AI is throwing the item to another Pokémon (friend or foe). If not set, the AI is consuming the item.
* If bit 1 is set, the AI is targeting an ally. If not set, the AI is targeting an enemy.

The description says bit 1 is only for the Warp Seed, which is consistent with Rescue Team and leads me to believe the field works the same in both games.

symbols/overlay29.yml
```diff
@@ -8500,6 +8500,44 @@ overlay29:
...
+        r0: Pointer to entity
+        
+        return: True if yes, false if no
+      - name: GetRelationBetweenMonsters
+      address:
+        EU: 0x2302188
+        NA: 0x230157C
+        JP: 0x23027FC
+      description: |-
+        Called to check if two monsters are allies, enemies, or neither (one is a shopkeeper, decoy, etc.)
```
AnonymousRandomPerson, line 8533: This looks like a [similar function in Rescue Team](https://github.com/pret/pmd-red/blob/master/src/dungeon_ai_targeting.c#L463). There's more nuance to the return values than it may initially seem.

To be more precise, this function returns whether the first entity should _treat_ the second Pokémon as an ally, enemy, or ignore them. The complexity here is with decoy handling, because Pokémon prioritize targeting a decoy over other Pokémon. In addition, decoy handling is handled somewhat differently if a teammate or a wild Pokémon becomes a decoy. Wild decoys will be targeted by other wild Pokémon, while teammates won't be targeted by other teammates (thankfully). The decoy targeting is accomplished by returning different values from this function if a decoy is present, such that Pokémon target the decoy and ignore other Pokémon that they would normally target.

Aside from a few (well, quite a few) conditions at the beginning, most of the decisions are made by indexing into a 4-D array based on a number of conditions: the "decoy targeting mode" of the Pokémon, whether entity 1 is an enemy, whether entity 2 is an enemy, and whether entity 2 is a decoy. The array enumerates all possible combinations of these conditions, with the resulting targeting relation to use for each combination.

symbols/overlay29.yml
```diff
@@ -8500,6 +8500,44 @@ overlay29:
...
+
+        r0: Pointer to entity 1
+        r1: Pointer to entity 2
+        r2: ? (If == 0, will make function return 2 if r1 is invisible and r0 can see invisible monsters)
```
AnonymousRandomPerson, line 8537: r2 is true if entity 1 should not take into account the target's invisibility (i.e., target them anyway even when invisible).

r3 is true if entity 1 should ignore a petrified target. This is used to ensure that your allies don't attack petrified targets, since attacking a petrified target will un-petrify them.

UsernameFodder (changes requested):
Thanks for contributing!

Your formatting wasn't quite correct. I added a formatting commit.

assidion2:
All right, I've fixed the formatting and added AnonymousRandomPerson's new info (I didn't know Rescue Team had been decompiled, that seems like a really useful reference).

symbols/overlay29.yml
assidion2: "while teammates won't be targeted by other teammates (thankfully)."

Is this really the case? I definitely remember an enemy using Substitute or something similar on me and having my whole team target me lol

AnonymousRandomPerson: If your team has AOE moves then they may hit friendly decoys accidentally. I believe they don't target you directly with single-target moves.

symbols/overlay29.yml
assidion2: So uint8_t freeze in statuses is 6 if the Pokémon is petrified? I suspected as much but it's weird that isn't documented, it only says that it 1 means a freeze status condition.

headers/functions/overlay29.h
```diff
@@ -274,6 +274,7 @@ bool CanAiMonsterMoveInDirection(struct entity* monster, enum direction_id direc
                                  bool* out_monster_in_target_position);
 bool ShouldMonsterRunAway(struct entity* monster);
 bool ShouldMonsterRunAwayVariation(struct entity* monster, undefined param_2);
+enum monster_relation GetRelationBetweenMonsters(struct entity* entity1, struct entity* entity2, bool param_3, bool param_4);
```
AnonymousRandomPerson, line 277: Since params 3 and 4 are known, they could be named more descriptively.

symbols/overlay29.yml
AnonymousRandomPerson: The statuses are grouped into a number of different categories such that you can only be inflicted with one status out of each category at a time. The frozen field encompasses several different statuses like freeze, wrap, ingrain, and petrified, which are indicated by different values on the frozen field. Petrified corresponds to 6.

headers/functions/overlay29.h
assidion2: Done

symbols/overlay29.yml
assidion2: Is there any documentation for what values correspond to what?

assidion2: Never mind I found it

headers/types/dungeon_mode/enums.h
```diff
@@ -1442,4 +1442,11 @@ enum fade_status_dungeon {
     DUNGEON_FADE_0x5 = 5,
 };
 
+// Used to indicate whether two monsters are allies, enemies, or neither.
+enum monster_relation {
+    RELATION_ALLY = 0,
+    RELATION_ENEMY = 1,
+    RELATION_OTHER = 2,
+};
```
Frostbyte0x70, line 1450: Since we now know this is actually used to determine how one pokémon should treat another, how about renaming it?

```suggestion
// Used to indicate how a given monster should target another.
enum monster_treatment {
    TREATMENT_TREAT_AS_ALLY = 0,
    TREATMENT_TREAT_AS_ENEMY = 1,
    TREATMENT_IGNORE = 2,
};
```
@AnonymousRandomPerson let me know if those values make sense, since you seem to be familiar with the function already.

symbols/overlay29.yml
```diff
@@ -2982,6 +2982,19 @@ overlay29:
...
+        NA: 0x230157C
+        JP: 0x23027FC
+      description: |-
+        Called to check if two monsters are allies, enemies, or neither.
+        (Examples of the neither case: target is a shopkeeper, there is a decoy on the floor, etc.)
+        
+        r0: Pointer to entity 1
+        r1: Pointer to entity 2
+        r2: If true, entity 1 will target enemies even if they are invisible
+        r3: If true and entity 1 is a team member, entity 1 will ignore petrified targets
```
Frostbyte0x70, line 2997: I see you changed this part after AnonymousRandomPerson's comment, but I think it's a bit confusing now, since the name of the function implies that it only returns the relation between monsters, but these parameters mention targeting and ignoring. I liked it more when it explained how the parameters affected the return of the function.

Here's my suggestion (see also my rename proposal above):
```suggestion
    - name: GetTreatmentBetweenMonsters
      address:
        EU: 0x2302188
        NA: 0x230157C
        JP: 0x23027FC
      description: |-
        Called to check if a monster should treat another as an ally, enemy, or ignore it.
        (Examples of the "ignore" case: target is a shopkeeper, there is a decoy on the floor, etc.)
        
        r0: Pointer to entity 1
        r1: Pointer to entity 2
        r2: If false, return TREATMENT_IGNORE if entity 2 is invisible and entity 1 cannot see invisible monsters
        r3: If true, return TREATMENT_IGNORE if entity 1 is a team member and entity 2 has the STATUS_PETRIFIED status
        return: Treatment that monster 1 should apply to monster 2
```

Frostbyte0x70 (changes requested):

headers/types/dungeon_mode/enums.h
AnonymousRandomPerson: Yep, names like these make sense.

assidion2:
I've added the new names.

Frostbyte0x70 (approved):
Looks good, thanks!

UsernameFodder (approved):
Thanks!

PR #267, opened 2024-08-22
Add dungeon sprite buffer size

irdkwia:
Adds two symbols related to dungeon sprite buffer size, which controls how many bytes are used in the main heap to store sprite file data in dungeon mode.


UsernameFodder (approved):
Thanks!

PR #268, opened 2024-08-24
More new functions related to AI

assidion2:


UsernameFodder:
Could you resynchronize the branch with this repo? As is all the commits from your last pull request are being included in this one as well.

If you aren't sure how to do it, start with your fork's `master` branch as described in [this tutorial](https://github.com/firstcontributions/first-contributions/blob/main/additional-material/git_workflow_scenarios/keeping-your-fork-synced-with-this-repository.md). Then once you've done that, you can update this branch by doing the following. (If you aren't comfortable with git, I'd recommend saving a copy of your files to somewhere outside of your project directory, in case something goes wrong.)

```
git checkout assidion-branch
git rebase master
```

assidion2:
Done

symbols/overlay29.yml
```diff
@@ -8536,6 +8536,75 @@ overlay29:
         Called when the in-dungeon "others" menu is open. Does not return until the menu is closed.
         
         return: Always 0
+    - name: CheckVariousStatuses
+      address:
+        EU: 0x23016DC
+        NA: 0x2300AD0
+        JP: 0x2301D50
```
UsernameFodder, line 8543: These NA/JP addresses aren't correct. I checked them in Ghidra, and they end up in the middle of random functions.

I can see that you computed the NA/JP symbols here by adding the EU offset (0x23016DC) to the (NA - EU == 0x234FA24 - 0x2350630) and (JP - EU == 0x2350CA4 - 0x2350630) shifts from `OthersMenu`. In general, naively trying to compute the offsets of symbols in a different version like this is not reliable. The different versions can sometimes change/insert/delete code in between your symbol and the "reference" symbol you're using, which means the interversion shifts will not be the same between the two symbols.

We have more reliable ways of automatically determining the proper offsets for different versions, so please don't fill these in with "guesses" if you aren't sure they're correct. Having less information is better than having wrong information, and we can autofill the other versions afterwards. `pmdsky-debug` allows you to omit the addresses for versions you aren't sure about, so you can just add the address in the version you're working on (EU). For example:

```yml
    - name: CheckVariousStatuses
      address:
       EU: 0x23016DC
```

is a perfectly valid entry on its own.

symbols/overlay29.yml
```diff
@@ -8536,6 +8536,75 @@ overlay29:
...
+    - name: IsChargingAnyTwoTurnMove
+      address:
+        EU: 0x2325084
+        NA: 0x2324478
+        JP: 0x23256F8
+      description: |-
+        Returns a boolean indicating whether or not the given entity is charging any two-turn move.
+
+        r0: Entity pointer
+        r1: Unused boolean
```
UsernameFodder, line 8571: Are you sure this is unused? Looking at the code, I see its value being checked in a conditional.

AnonymousRandomPerson: In [Rescue Team](https://github.com/pret/pmd-red/blob/master/src/charge_move.c#L116), if this flag is true, the function will also return true if the Pokémon has the Charge status (i.e., the Electric move that boosts other Electric moves).

assidion2: This function goes through an array which contains all of the values of the bide status for which it should return true. The charge status is not listed in the array. The conditional will only be activated if iVar1 (the current index in the array) is > 99, and since the loop breaks upon reaching a 0 value in the array, it will never be activated as the first 0 occurs much sooner.

AnonymousRandomPerson: Aha, good point. So the bool is _supposed_ to include the Charge status but it doesn't actually work as intended.

symbols/overlay29.yml
assidion2: Sorry about that, I've removed them.

symbols/overlay29.yml
UsernameFodder: Can we document this? Let's name the variable `charge_check_unused` or something, and then mention in the description what it's supposed to do, and how it's unused.

symbols/overlay29.yml
```diff
@@ -8536,6 +8536,63 @@ overlay29:
...
+    - name: GetPossibleAiItemTargets
+      address:
+        EU: 0x230FAA0
+      description: |-
+        Gets the positions of all targets that an AI can hit with a thrown item such as a Gravelerock.
+        The number of positions in the array will be stored at 0x023542FC.
+
+        r0: Entity pointer
+        r1: Item struct pointer
+        r2: Array of size 20 for storing position structs
```
UsernameFodder, line 8594: Let's follow the convention for marking output variables:
```suggestion
        r2: [output] Array of size 20 for storing position structs
```

symbols/overlay29.yml
```diff
@@ -8536,6 +8536,63 @@ overlay29:
...
+        return: bool
+    - name: MonsterCanThrowItems
+      address:
+        EU: 0x2302D94
+      description: |-
+        Returns a boolean indicating whether or not the given monster can throw items based on its monster id.
+
+        r0: Monster pointer
+        return: bool
+    - name: GetPossibleAiItemTargets
```
UsernameFodder, line 8585: Can we specify _thrown_ items instead of just "items"? I.e. `GetPossibleAiThrownItemTargets`

symbols/overlay29.yml
```diff
@@ -8536,6 +8536,63 @@ overlay29:
...
+      address:
+        EU: 0x230FAA0
+      description: |-
+        Gets the positions of all targets that an AI can hit with a thrown item such as a Gravelerock.
+        The number of positions in the array will be stored at 0x023542FC.
```
UsernameFodder, line 8590: Can you define a data symbol for this instead of referencing the raw offset? 0x23542FC is specific to EU.

UsernameFodder (changes requested):

assidion2:
Made the requested changes.

headers/data/overlay29.h
```diff
@@ -102,5 +102,6 @@ extern struct tile DEFAULT_TILE;
 extern bool HIDDEN_STAIRS_SPAWN_BLOCKED;
 extern void* FIXED_ROOM_DATA_PTR;
 extern struct dungeon_fades* DUNGEON_FADES_PTR;
+extern uint32_t POSITIONS_COUNTER;
```
UsernameFodder, line 105: For a global variable, this name is sort of generic. Can we qualify it a little more? How about `AI_THROWN_ITEM_TARGET_POSITIONS_COUNT`?

UsernameFodder (changes requested):

headers/data/overlay29.h
assidion2: I actually discovered that this counter is also used in FUN_0230faa0 for multiple arrays (one being direction enums and another being probabilities). I don't know what non-generic name could encompass all of these cases, do you have any ideas?

UsernameFodder: It looks like `FUN_0230f388` is a higher-level function for deciding...something, and it calls into (and is the only caller of) both `FUN_0230f824` and `GetPossibleAiArcItemTargets`. Additionally, the first call site of the subfunctions within the higher-level one is inside the block that checks `MonsterCanThrowItems`. So, I think `FUN_0230f824` must also be related to thrown items. Based on what you said about it dealing with directions and probabilities, perhaps it's a `THROWN_LINE` counterpart to the `THROWN_ARC` function you found. Additionally, this `POSITIONS_COUNTER` value is only accessed by the aforementioned three functions.

So, to me it seems like maybe it counts the number of possible actions the AI can take with a thrown item in a given category. With arc-thrown items, that can be encoded by a list of positions of targets. With line-thrown items, it seems plausible that that might be encoded by a list of directions to throw, and the corresponding probability of throwing in each direction? So maybe something like `AI_THROWN_ITEM_ACTION_CHOICE_COUNT`. We can add what we know in the description for the symbol.

AnonymousRandomPerson: All of this is consistent with Rescue Team.

`FUN_0230f824` ([FindStraightThrowableItems](https://github.com/pret/pmd-red/blob/master/src/dungeon_ai_items.c#L274) in RT) is the `THROWN_LINE` counterpart to `GetPossibleAiArcItemTargets`.

`FUN_0230f388` ([DecideUseItem](https://github.com/pret/pmd-red/blob/master/src/dungeon_ai_items.c#L64) in RT) is the top-level function for deciding if an AI should use its held item and where to throw it (if applicable).

`POSITIONS_COUNTER` ([gNumPotentialTargets](https://github.com/pret/pmd-red/blob/master/src/dungeon_ai_items.c#L37) in RT) is used by the thrown line/arc functions to count the number of targets the AI has found for throwing its item. Line-thrown items are tracked with two parallel arrays of direction + probability, while arc-thrown items track the positions of targets.

assidion2: I also made these findings yesterday. I've added them, this should be everything related to AI item use.

headers/data/overlay29.h
```diff
@@ -102,5 +102,9 @@ extern struct tile DEFAULT_TILE;
 extern bool HIDDEN_STAIRS_SPAWN_BLOCKED;
 extern void* FIXED_ROOM_DATA_PTR;
 extern struct dungeon_fades* DUNGEON_FADES_PTR;
+extern uint32_t AI_THROWN_ITEM_ACTION_CHOICE_COUNT;
+extern bool DIRECTION_BOOLS[8];
```
UsernameFodder, line 106: Can we make this less generic? `AI_THROWN_ITEM_DIRECTION_IS_USED` or something

headers/types/dungeon_mode/dungeon_mode.h
```diff
@@ -314,8 +314,8 @@ struct action_data {
     struct direction_id_8 direction; // 0x2: Direction in which the action will be performed
     undefined field_0x3;
     struct action_parameter action_parameters[2]; // 0x4: Parameters for the action
-    int16_t field_0x10;
-    int16_t field_0x12;
+    int16_t arc_item_target_pos_x; // 0x10: Used for storing the target position of an item thrown in an arc, such as a Gravelerock
+    int16_t arc_item_target_pos_y; // 0x12: Used for storing the target position of an item thrown in an arc, such as a Gravelerock
```
UsernameFodder, line 318:
```suggestion
    struct position arc_item_target_pos; // 0x10: Used for storing the target position of an item thrown in an arc, such as a Gravelerock
```

symbols/overlay29.yml
```diff
@@ -8536,6 +8536,102 @@ overlay29:
...
+        The size of the arrays will be stored in AI_THROWN_ITEM_ACTION_CHOICE_COUNT.
+        The caller function will select the direction to throw the item by iterating through the array(s), rolling the probability, and then throwing in that direction if the roll succeeds.
+        Nothing will be thrown if all rolls fail.
+
+        r0: Entity pointer
+        r1: Integer in {1, 2}. If 1, target allies; if 2, target enemies.
+        r2: Item struct pointer
+        r3: If false, will call GetAiUseItemProbability to get the probability of throwing in a certain direction.
+        If true, the added probability will always be 100.
+    - name: SetDecoyTracker
```
UsernameFodder, line 8611: Can we call this `SetDecoyAiTracker` to match the field name?

symbols/overlay29.yml
```diff
@@ -8536,6 +8536,102 @@ overlay29:
...
+        r0: Entity pointer
+    - name: TestItemAiFlag
+      address:
+        EU: 0x200EB48
+      description: |-
+        Returns a boolean indicating whether the item is consumable, throwable at an ally, or throwable at an enemy.
+        The table used for this is inaccessible in the code, as it is loaded from a file in the ROM at runtime.
+
+        r0: item_id enum
+        r1: Integer in {0, 1, 2}. If 0, test if consumable; if 1, test if throwable at ally; if 2, test if throwable at enemy.
```
UsernameFodder, line 8634:
```suggestion
        r1: Integer in {0, 1, 2}. If 0, test if consumable; if 1, test if throwable at ally; if 2, test if throwable at enemy.
        return: bool
```

symbols/overlay29.yml
```diff
@@ -8536,6 +8536,102 @@ overlay29:
...
+
+        r0: Entity pointer
+    - name: AiDecideUseItem
+      address:
+        EU: 0x230F388
+      description: |-
+        Decides whether or not an AI should use its held item and updates its action_data fields accordingly.
+
+        r0: Entity pointer
+    - name: TestItemAiFlag
```
UsernameFodder, line 8626: Can we add a function prototype?

symbols/overlay29.yml
UsernameFodder, line 8634: Can we add an enum in `dungeon_mode/enums.h`? How about:

```c
enum item_flag {
    ITEM_FLAG_CONSUMABLE = 0,
    ITEM_FLAG_THROWABLE_ALLY = 1,
    ITEM_FLAG_THROWABLE_ENEMY = 2,
};
```

UsernameFodder (changes requested):

symbols/overlay29.yml
assidion2: This is a function in arm9 that already has a header, I put this in overlay29 by mistake.

symbols/overlay29.yml
```diff
@@ -9669,3 +9757,27 @@ overlay29:
...
       description: "[Runtime] Pointer to the dungeon fades struct that maintains the status of screen fades in dungeon mode."
+    - name: AI_THROWN_ITEM_ACTION_CHOICE_COUNT
+      address:
+        EU: 0x23542FC
+      length:
+        EU: 0x4
+      description: "[Runtime] Used to store the number of positions output by GetPossibleAiArcItemTargets and the number of directions/probabilities output by GetPossibleAiThrownItemDirections."
+    - name: AI_THROWN_ITEM_DIRECTION_IS_USED
+      address:
+        EU: 0x237D5D0
```
UsernameFodder, line 9768: These look to be failing the checks. I think this belongs in `ram.yml`?

UsernameFodder (changes requested):
Mostly looks good, we just need to make pipeline fixes (formatting/tests) now.

symbols/overlay29.yml
assidion2: Moved it

assidion2: The counter actually seems to be located in overlay29, which is weird.

UsernameFodder (approved):
I fixed the formatting and checks for you, but for future reference, you can view the errors yourself in the result logs. For formatting issues, you can also download the formatted files with the automatic fixes, as described in the contribution docs for [symbols](https://github.com/UsernameFodder/pmdsky-debug/blob/master/symbols/README.md#follow-style-conventions) and [headers](https://github.com/UsernameFodder/pmdsky-debug/blob/master/headers/README.md#follow-style-conventions).

PR #274, opened 2024-09-02
Some hidden stairs functions

assidion2:


UsernameFodder (approved):
I added param names to the prototypes, but otherwise looks good, thanks.

PR #276, opened 2024-09-06
Monster initialization functions

assidion2:


symbols/overlay29.yml
```diff
@@ -492,6 +492,10 @@ overlay29:
         r0: Entity pointer
         r1: Target position
         return: True if the specified monster can target the target position, false otherwise.
+    - name: PopulateActiveMonsterPtrs
+      address:
+        EU: 0x22E32B8
+      description: Populates active_monster_ptrs with all valid monsters in the monster_slot_headers array.
```
UsernameFodder, line 498: Do you know that there's no params? Not mentioning anything is meant for when you aren't sure. If you are, put:

```suggestion
      description: |-
        Populates active_monster_ptrs with all valid monsters in the monster_slot_headers array.
        
        No params.
```

headers/functions/overlay29.h
```diff
@@ -34,6 +34,7 @@ bool ShouldDisplayEntityWrapper(struct entity* entity);
 bool CanSeeTarget(struct entity* user, struct entity* target);
 bool CanTargetEntity(struct entity* user, struct entity* target);
 bool CanTargetPosition(struct entity* monster, struct position* position);
+void PopulateActiveMonsterPtrs();
```
UsernameFodder, line 37: If there's no params, do

```suggestion
void PopulateActiveMonsterPtrs(void);
```

Putting () is only correct in C++; in C it technically means something else.

UsernameFodder (changes requested):

UsernameFodder (approved):
Thanks!

PR #280, opened 2024-10-01
Overlay31 research

assidion2:
Some overlay31 findings related to the stairs menu.


symbols/overlay29.yml
```diff
@@ -248,7 +248,7 @@ overlay29:
         
         Floor types:
           0 appears to mean the current floor is "normal"
-          1 appears to mean the current floor is a fixed floor
+          1 means the current floor is a fixed floor with Is Free Layout enabled
```
Frostbyte0x70, line 251: "Is free layout" is not a vanilla setting, it's added by a patch. So the actual meaning has to be something else.

assidion2: In my testing, it returned 1 when it was a fixed floor with a Warp Zone and 0 otherwise; I assumed they were one and the same.

symbols/ram.yml
```diff
@@ -685,3 +685,10 @@ ram:
         This data is populated as the dungeon floor is generated.
         
         type: struct floor_generation_status
+    - name: STAIRS_MENU_PTR
+      address:
+        EU: 0x238AEC4
+      length:
+        NA: 0x4
```
UsernameFodder, line 692: It seems weird that there's no EU length here. Is there a reason to omit it?

assidion2: Oops. I ctrl v'ed it from somewhere and didn't check.

PR #282, opened 2024-10-08
Some functions related to badges/ranks

assidion2:


symbols/arm9.yml
```diff
@@ -4804,6 +4804,15 @@ arm9:
         NA: 0x2024428
         JP: 0x2024478
       description: "Note: unverified, ported from Irdkwia's notes"
+    - name: GetRankString
+      address:
+        EU: 0x2024D88
+      description: |-
+        Gets the string corresponding to the player's current explorer rank.
+        
+        r0: [output] str
```
UsernameFodder, line 4813: Can you add more description? If this is the same as the later functions:

```suggestion
        r0: [output] Pointer to the buffer where the string will be written
```

UsernameFodder (changes requested):

UsernameFodder (approved):
Thanks!

PR #292, opened 2024-11-20
Various Dungeon Mode Findings

Jawshoeuh:
Includes snatch, explorer maze, shadows, fixed rooms, warp scarf, traps.


headers/types/dungeon_mode/dungeon_mode.h
```diff
@@ -990,9 +993,10 @@ struct dungeon_generation_info {
     bool dough_seed_extra_poke_flag;
     // 0x5: Room index of Monster House on the floor. 0xFF if there's no Monster House
     uint8_t monster_house_room;
-    // 0x6: Related to when a monster from a fixed room faints. Maybe to check if the floor
-    // should be over after knocking them out?
-    undefined unk_fixed_room_static_monster_tracker;
+    // 0x6: It's set to 0 when it's not a fixed room, 1 when it's a fixed room with no boss fight,
+    // and 2 when it's a fixed room with a boss fight. If this tracker is 0, 1, or 3 the floor
```
UsernameFodder, line 997: So is 3 also a legitimate value? The first sentence makes it seem like 0, 1, and 2 are the only allowed values.

symbols/overlay29.yml
```diff
@@ -2244,6 +2260,23 @@ overlay29:
         r0: Entity pointer. Must be a monster, otherwise the function does nothing.
         r1: ?
         r2: ?
+    - name: ReevaluateSnatchMonster
+      address:
+        NA: 0x22F9970
+      description: |-
+        Checks if any monsters on the floor has the snatch status and stores the corresponding info
```
UsernameFodder, line 2267: What happens if there are multiple monsters with the snatch status? This description makes it seem ambiguous.

symbols/overlay29.yml
```diff
@@ -3442,6 +3475,26 @@ overlay29:
         
         r0: pointer to entity
         return: animation ID
+    - name: DetermineAllMonsterShadow
+      address:
+        NA: 0x2304B64
+      description: |-
+        Change all monster's shadow to be approppriate for their size and the tile they're
```
UsernameFodder, line 3482:
```suggestion
        Change all monsters' shadows to be appropriate for their sizes and the tile they're
```

symbols/overlay29.yml
```diff
@@ -3442,6 +3475,26 @@ overlay29:
...
+        standing on. It simply calls DetermineMontersShadow for all monsters in the dungeon.
+        
+        No params.
+    - name: DetermineMonsterShadow
+      address:
+        NA: 0x2304BAC
+      description: |-
+        Changes the monster's shadow to be approppriate for its size and the tile its standing on.
```
UsernameFodder, line 3490:
```suggestion
        Changes the monster's shadow to be appropriate for its size and the tile it's standing on.
```

UsernameFodder (changes requested):

symbols/overlay29.yml
Jawshoeuh: If there are multiple monsters with snatch, it would set it to the last one in the list. When trying to inflict snatch using TryInflictSnatchStatus, it will remove snatch from any other monsters first to prevent multiple from having it. Would you like me to specify that normally multiple monsters can not have snatch, but if multiple do it selects the last one in the list of monsters?

headers/types/dungeon_mode/dungeon_mode.h
Jawshoeuh: The game specifically checks for 3 when passing the floors, but I couldn't find anywhere the game sets it to 3. Could be related to a debug thing or I just missed another location which does set it to 3.

symbols/overlay29.yml
Jawshoeuh: Correction: The first one in the list.

headers/types/dungeon_mode/dungeon_mode.h
UsernameFodder: I see, I guess this is fine.

symbols/overlay29.yml
UsernameFodder: Yeah I think that extra info would be good

symbols/overlay29.yml
```diff
@@ -2244,6 +2260,25 @@ overlay29:
         r0: Entity pointer. Must be a monster, otherwise the function does nothing.
         r1: ?
         r2: ?
+    - name: ReevaluateSnatchMonster
+      address:
+        NA: 0x22F9970
+      description: |-
+        Checks if any monsters on the floor has the snatch status and stores the corresponding info
```
UsernameFodder, line 2267:
```suggestion
        Checks if any monster on the floor has the snatch status and stores the corresponding info
```

UsernameFodder (changes requested):

UsernameFodder (approved):

PR #294, opened 2024-11-23
Document HasTactic

AnonymousRandomPerson:


symbols/overlay29.yml
```diff
@@ -3279,6 +3279,17 @@ overlay29:
...
+    - name: HasTactic
+      address:
+        EU: 0x230294C
+        NA: 0x2301F20
+        JP: 0x2303470
+      description: |-
+        Returns whether a monster is set to use a specific tactic.
+        
+        r0: Entity pointer
+        r1: Enum tactic_id
```
UsernameFodder, line 3291: This doesn't match the prototype?

UsernameFodder (changes requested):

symbols/overlay29.yml
AnonymousRandomPerson: Oops, fixed.

symbols/overlay29.yml
```diff
@@ -3279,6 +3279,17 @@ overlay29:
         
         r0: Entity pointer
         return: True if the monster can see invisible monsters.
+    - name: HasTactic
```
UsernameFodder, line 3282: I feel like "has" could potentially be interpreted as "is unlocked", which might be confusing. How about something like `IsTacticSet`?

AnonymousRandomPerson: Valid, I've renamed it.

UsernameFodder (approved):

PR #296, opened 2024-12-04
Random tile, text palette, ITCM functions

assidion2:


symbols/arm9.yml
```diff
@@ -5070,6 +5070,18 @@ arm9:
         r1: x offset within window
         r2: y offset within window
         r3: text to draw
+    - name: GetColorCodePaletteOffset
+      address:
+        EU: 0x20268A8
+        NA: 0x20265C4
+        JP: 0x202669C
+      description: |-
+        Gets the offset of a text color symbol's 2-byte RGB5 color in the palette stored in VRAM at 0x6882000.
```
UsernameFodder, line 5079: Is this VRAM offset the same for all versions?

headers/functions/arm9/itcm.h
```diff
@@ -1,6 +1,7 @@
 #ifndef HEADERS_FUNCTIONS_ARM9_ITCM_H_
 #define HEADERS_FUNCTIONS_ARM9_ITCM_H_
 
+void CopyAndInterleave(int* dst, int* src, int len, int val);
```
UsernameFodder, line 4: Are these really 4-byte signed `int` values? The description makes me think `dst` should be `uint32_t*`, `src` should be `uint16_t*`, and `val` should be `uint8_t`, though I haven't looked at what the access pattern looks like.

headers/functions/arm9/itcm.h
```diff
@@ -26,6 +27,7 @@ int GetKeyM2N(enum monster_id monster_id);
 int GetKeyM2NBaseForm(enum monster_id monster_id);
 void HardwareInterrupt(void);
 void ReturnFromInterrupt(void);
+void InitDmaTransfer_Standard(int channel, int* src, int* dst, uint16_t word_count);
```
UsernameFodder, line 30: Should the pointers be `void*`? DMA seems more generic than specifically integers

headers/functions/arm9/itcm.h
assidion2: The accesses are all ldrh/strh so I suppose both should be uint16_t*

symbols/arm9.yml
UsernameFodder: Ok seems reasonable

UsernameFodder (approved):
Thanks!

PR #299, opened 2024-12-21
Small assortment of arm9 stuff

Adex-8x:
Some arm9 shenanigans involving the system clock, firmware user settings, etc.


UsernameFodder (approved):
Thanks!

PR #301, opened 2024-12-28
Add status class enums, fill in language addresses

AnonymousRandomPerson:


headers/data/overlay29.h
```diff
@@ -77,6 +77,7 @@ extern struct fx64_16 DAMAGE_FORMULA_MAX_BASE;
 extern struct fx64_16 WONDER_GUARD_MULTIPLIER;
 extern struct fx64_16 DAMAGE_FORMULA_MIN_BASE;
 extern struct damage_negating_exclusive_eff_entry TYPE_DAMAGE_NEGATING_EXCLUSIVE_ITEM_EFFECTS[28];
+extern uint8_t TWO_TURN_STATUSES[11];
```
UsernameFodder, line 80: Should this be `struct status_id_8`?

AnonymousRandomPerson: No, it's not the combined `status_id` enum. It's compared to the value of [monster.statuses.bide](https://github.com/UsernameFodder/pmdsky-debug/blob/master/headers/types/dungeon_mode/dungeon_mode.h#L132), which is a regular `uint8_t`. The value matches the enum [status_two_turn_id](https://github.com/UsernameFodder/pmdsky-debug/blob/master/headers/types/dungeon_mode/enums.h#L276), except that this enum is 16-bit while `TWO_TURN_STATUSES` has 8-bit values.

UsernameFodder: Let's make an 8-bit version of the enum and use that.

AnonymousRandomPerson: Since the other status classes are handled similarly, I made enums for all those as well as this.

symbols/ram.yml
```diff
@@ -270,17 +278,44 @@ ram:
...
+      length:
+        NA: 0x4
+      description: Pointer to SOUND_MEMORY_ARENA.
+    - name: WINDOW_LIST
+      address:
+        EU: 0x22A921C
+        NA: 0x22A88DC
+        JP: 0x22AA0E8
+      length:
+        EU: 0x8
```
UsernameFodder, line 301: This length is wrong, it should be 0x1180

symbols/ram.yml
```diff
@@ -428,9 +474,11 @@ ram:
       address:
         EU: 0x22AC258
         NA: 0x22AB918
+        JP: 0x22AD0D4
       length:
         EU: 0xC
         NA: 0xC
+        JP: 0x8
```
UsernameFodder, line 481: Is this supposed to be shorter?

Also I just noticed that the struct declaration says this is 10 bytes long, which is somehow different than all the listed lengths? Can we fix this? We can use the `PMDSKY_VERSION` macro

symbols/ram.yml
```diff
@@ -524,6 +576,11 @@ ram:
       address:
         EU: 0x22B9F88
         NA: 0x22B9648
+        JP: 0x22BAE00
+      length:
+        EU: 0xC
```
UsernameFodder, line 581: This length is wrong, it's unknown but definitely not 0xC

symbols/ram.yml
```diff
@@ -549,18 +606,24 @@ ram:
...
-        1 byte per lock. Exact size isn't confirmed, it could potentially be longer.
+        1 byte per lock.
     - name: GROUND_MEMORY_ARENA_1
       address:
+        EU: 0x2325B00
         NA: 0x2324FC0
+        JP: 0x2326520
       length:
-        NA: 0x1C
+        EU: 0x64000
```
UsernameFodder, line 624: This is wrong, the old length was correct

UsernameFodder (changes requested):

symbols/ram.yml
AnonymousRandomPerson: The description said "this arena occupies 0x64000 bytes of space". Is the 0x1C only for the metadata?

symbols/ram.yml
AnonymousRandomPerson: I put 0xC here because there is another symbol at NA 0x022B9654, so the current assumption that all the fields in `thread_info` are part of one struct may be incorrect. Though I don't have more context on this field so I'll remove the length for now.

symbols/ram.yml
AnonymousRandomPerson: Yes, JP allocates less space for this symbol than the other versions.

symbols/ram.yml
UsernameFodder: As is stated in the symbol description, see `struct mem_arena`

AnonymousRandomPerson: IMO between the symbol name and description, it's easy to confuse this symbol as representing the whole memory arena rather than just the arena metadata. It'd be more clear if the symbol and struct were named `MEMORY_ARENA_1_METADATA` and `mem_arena_metadata`. But I'm not that invested in this symbol, so I'm fine leaving it as is.

symbols/ram.yml
```diff
@@ -428,30 +474,36 @@ ram:
...
+        JP: 0x22AD0D4
       length:
         EU: 0xC
         NA: 0xC
+        JP: 0x8
       description: |-
         The team name.
         
-        A null-terminated string, with a maximum length of 10. Presumably encoded with the ANSI/Shift JIS encoding the game typically uses.
+        A null-terminated string, with a maximum length of 12 (or 8 in JP). Presumably encoded with the ANSI/Shift JIS encoding the game typically uses.
```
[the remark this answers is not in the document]
AnonymousRandomPerson: I took a closer look at the ASM here. There are 12 bytes allocated for this symbol in BSS for EU/NA and 8 for JP, but the ASM limits the string to size 10 or 5 (respectively). The extra bytes in BSS are probably just for alignment. I've adjusted the description and size to match what the ASM has.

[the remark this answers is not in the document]
UsernameFodder: Ah I see, I guess that's where the 12 came from. That seems fine, thanks for looking into it

UsernameFodder (approved):
All looks good now, thanks!

Issue #246, opened 2024-01-10
Document script engine: Transfer Resarch from SkyTemple SSB Debugger & PsyCommando's notes


theCapypara:
In early 2020 I made a lot of research based on Psy's research on the game's script engine. A lot of this made it into SkyTemple's script engine debugger and also pmdsky-debug.

Some of the most important undocumented things are related to the script engine thread state structs and other entities (actors, events, map, objects, performers, global script):

https://github.com/SkyTemple/skytemple-ssb-emulator/blob/main/src/script_runtime.rs
https://github.com/SkyTemple/skytemple-ssb-debugger/blob/master/skytemple_ssb_debugger/model/script_runtime_struct.py
https://github.com/SkyTemple/skytemple-ssb-debugger/blob/master/skytemple_ssb_debugger/model/ground_state/global_script.py
https://github.com/SkyTemple/skytemple-ssb-debugger/blob/master/skytemple_ssb_debugger/model/ground_state/ssb_file_in_ram.py
https://github.com/SkyTemple/skytemple-ssb-debugger/blob/master/skytemple_ssb_debugger/model/ground_state/actor.py
https://github.com/SkyTemple/skytemple-ssb-debugger/blob/master/skytemple_ssb_debugger/model/ground_state/event.py
https://github.com/SkyTemple/skytemple-ssb-debugger/blob/master/skytemple_ssb_debugger/model/ground_state/map.py
https://github.com/SkyTemple/skytemple-ssb-debugger/blob/master/skytemple_ssb_debugger/model/ground_state/object.py
https://github.com/SkyTemple/skytemple-ssb-debugger/blob/master/skytemple_ssb_debugger/model/ground_state/performer.py 

Additionally I took notes from Psy and made my own, mostly connecting the dots between his research and traces I did with the emulator integration.

I uploaded all my notes I have here: https://github.com/theCapypara/pmdsky-ground-state-notes

theCapypara:
Some more notes start before and after here: https://discord.com/channels/710190644152369162/712341699292037121/1194771802916917331

Adex-8x:
To add onto this, Psy's notes: https://github.com/PsyCommando/pmd_2_doc/tree/main/doc/script_engine

PR #302, opened 2025-01-06
Ground entity (actor/object/performer) info

Adex-8x:
Finally tackling more script engine research! Related to https://github.com/UsernameFodder/pmdsky-debug/issues/246.

Documents information on operations related to ground entities, AKA actors/objects/performers that can be manipulated in the script engine. Will ultimately end up documenting:
- [x] Actor operations
- [x] Object operations
- [x] Performer operations
- [x] An attempt at partially documenting the type of `RunNextOpcode`'s parameter


headers/functions/overlay11.h
```diff
@@ -46,10 +46,35 @@ void AnimRelatedFunction(struct animation* anim, undefined4 param_2, undefined4
...
+int GetEffectStatusLiveActor(struct live_actor* actor);
+void GetAttributeBitfieldLiveActor(struct live_actor* actor, uint32_t* attribute_bitfield);
+void SetAttributeBitfieldLiveActorWrapper(struct live_actor* actor, uint32_t attribute_bitfield);
+void ResetAttributeBitfieldLiveActorWrapper(struct live_actor* actor, uint32_t attribute_bitfield);
+void SetBlinkLiveActor(struct live_actor* actor, uint16_t interval_blink_in,
+                       uint16_t interval_blink_out);
+void SetPositionOffsetLiveActor(struct live_actor* actor, struct vec2* offset);
+void InitPartnerFollowData(void);
+void GetDirectionLiveObject(undefined object, struct direction_id_8* target);
+void SetDirectionLiveObject(undefined object, struct direction_id_8 direction);
```
UsernameFodder, line 77: Let's use the raw `enum direction_id` for an input param; since it's a register value it's always going to have 4 bytes of capacity anyway, and it looks nicer in Ghidra decomps. I guess we can change `SetDirectionLiveActor` too

headers/functions/overlay11.h
```diff
@@ -46,10 +46,35 @@ void AnimRelatedFunction(struct animation* anim, undefined4 param_2, undefined4
...
+int GetAnimationStatusLiveActor(struct live_actor* actor);
+int GetEffectStatusLiveActor(struct live_actor* actor);
+void GetAttributeBitfieldLiveActor(struct live_actor* actor, uint32_t* attribute_bitfield);
+void SetAttributeBitfieldLiveActorWrapper(struct live_actor* actor, uint32_t attribute_bitfield);
+void ResetAttributeBitfieldLiveActorWrapper(struct live_actor* actor, uint32_t attribute_bitfield);
+void SetBlinkLiveActor(struct live_actor* actor, uint16_t interval_blink_in,
+                       uint16_t interval_blink_out);
+void SetPositionOffsetLiveActor(struct live_actor* actor, struct vec2* offset);
+void InitPartnerFollowData(void);
+void GetDirectionLiveObject(undefined object, struct direction_id_8* target);
```
UsernameFodder, line 76: I assume you'll define the object struct in a bit? But if not an unknown struct pointer should be `undefined*`, assuming that's what `object` is here

Adex-8x: Ah yeah, I'll fill out the `live_object` struct

Adex-8x:
Having filled out what I know about the `live_object` struct now, I notice how it shares a lot of its initial fields with that of `live_actor` (starting from `id` and ending with the `attribute_bitfield`). I'm sure performers will also follow this trend, so these fields might be able to be condensed into another struct? Unsure if that'd be fine, along with what it'd be named

UsernameFodder:
> Having filled out what I know about the `live_object` struct now, I notice how it shares a lot of its initial fields with that of `live_actor` (starting from `id` and ending with the `attribute_bitfield`). I'm sure performers will also follow this trend, so these fields might be able to be condensed into another struct? Unsure if that'd be fine, along with what it'd be named 

If they're shared, that'd be fine. Whatever the name is, it should reflect the purpose of those shared fields. Maybe something referencing the generic script entity.

headers/data/overlay11.h
```diff
@@ -14,7 +14,10 @@ extern struct dungeon_id_16 RECRUITMENT_TABLE_LOCATIONS[22];
 extern int16_t RECRUITMENT_TABLE_LEVELS[22];
 extern struct monster_id_16 RECRUITMENT_TABLE_SPECIES[22];
 extern struct level_tilemap_list_entry LEVEL_TILEMAP_LIST[81];
+extern struct ground_entity_function_table ACTOR_FUNCTION_POINTER_TABLE;
```
UsernameFodder, line 17: Let's remove the POINTER from the symbol names too? Seems more consistent.

headers/functions/overlay11.h
```diff
@@ -46,10 +46,63 @@ void AnimRelatedFunction(struct animation* anim, undefined4 param_2, undefined4
...
+int16_t GetIdLiveActor(struct live_actor* actor);
+void GetCollisionBoxLiveActor(struct live_actor* actor, struct uvec2* collision_box);
+void SetPositionInitialLiveActor(struct live_actor* actor, struct vec2* offset);
+void SetMovementRangeLiveActor(struct live_actor* actor, struct vec2* limit_min_pos,
+                               struct vec2* limit_max_pos);
+void GetCollisionBoxCenterLiveActor(struct live_actor* actor, struct uvec2* collision_box_center);
+void SetPositionLiveActorVeneer(struct live_actor* actor, struct vec2* pos);
+void GetHeightLiveActor(struct live_actor* actor, int* height, int* second_height);
+void SetHeightLiveActor(struct live_actor* actor, int height);
+void GetDirectionLiveActor(struct live_actor* actor, enum direction_id* target);
```
UsernameFodder, line 64: We need to be careful about the enum vs. struct here, since it's a pointer referencing actual memory (which has a size) rather than a register value (always 4 bytes). Is the output value written 1 byte or 4 bytes?

headers/data/overlay11.h
Adex-8x: Oops, yeah it's better to remove

headers/functions/overlay11.h
Adex-8x: The output here would be stored as 1 byte

headers/types/ground_mode/ground_mode.h
```diff
@@ -287,38 +287,118 @@ struct animation_data {
...
+    undefined field_0x5e;
+    undefined field_0x5f;
+    undefined4 field_0x60[4];
+};
+ASSERT_SIZE(struct script_opcode_statuses, 112);
+
+// A common struct used to hold data surrounding a live entity
+struct live_entity_info {
+    struct ground_entity_function_table* function_table;
+    void* ground_entity;                   // Actor, object, or performer
```
UsernameFodder, line 363: Is this a reference to the parent entity? Or some other structure?

Adex-8x: The parent entity; so either a pointer to `live_actor`, `live_object`, or `live_performer`. I don’t know why it’s self-referential, but that’s how it gets initialized as in `InitLiveEntityInfo` anyway

Adex-8x:
Oh, looking at the callsites of `FuncThatCallsRunNextOpcode`, I realize now that its parameter and the parameter of `RunNextOpcode` was actually a pointer to `script_opcode_statuses` lol
Welp, that makes this last bit easier, though I guess I'd rename the struct to `script_engine_state`

headers/types/ground_mode/ground_mode.h
UsernameFodder: Let's add a comment clarifying what this is

headers/functions/overlay11.h
```diff
@@ -43,13 +49,104 @@ struct animation_data GetIdleAnimationType(enum monster_id, undefined* param_2);
 void LoadObjectAnimData(struct animation* anim, int16_t object_id, uint32_t flags);
 void InitAnimDataFromOtherAnimDataVeneer(struct animation* dst, struct animation* src);
 void AnimRelatedFunction(struct animation* anim, undefined4 param_2, undefined4 param_3);
+bool UnlockMainRoutine(int16_t lock_id);
 void AllocAndInitPartnerFollowDataAndLiveActorList(void);
 void InitPartnerFollowDataAndLiveActorList(void);
+int GetLiveActorIdxFromScriptEntityId(enum script_entity_id);
```
UsernameFodder, line 55:
```suggestion
int GetLiveActorIdxFromScriptEntityId(enum script_entity_id entity_id);
```

UsernameFodder (changes requested):

UsernameFodder (approved):
Looks good, thanks! It's great to have this huge chunk of information recorded!

PR #303, opened 2025-01-12
Add a few functions previously documented only in CoT

Adex-8x:
Adds a few functions that were [only documented in CoT](https://github.com/SkyTemple/c-of-time/blob/8a1af65e4fc8d672ce458e690ab42b88c738403b/src/menus.c#L9-L13), mainly to then follow up by editing CoT to rely on the pmdsky-debug names.
~~`ScriptCaseProcess` isn't yet in CoT, but I was planning on using it for adding more custom ground instruction support, so why not get it out of the way now~~

### DIFF
--- a/symbols/arm9.yml
+++ b/symbols/arm9.yml
@@ -6735,6 +6735,18 @@ arm9:
         stack[0]: total number of selectable items
         stack[1]: number of selectable items per page
         stack[2]: ?
+    - name: SetupAndShowKeyboard
+      address:
+        EU: 0x2036AA8
+        NA: 0x20367B4
+        JP: 0x2036ACC
+      description: |-
+        Calls a function that seems to sets up info for the keyboard menu before ending with a call to ShowKeyboard.
+        
+        r0: menu type
+        r1: buffer1
+        r2: buffer2
+        return: ?
     - name: ShowKeyboard
       address:
         EU: 0x2036AE4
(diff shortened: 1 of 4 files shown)


symbols/arm9.yml
```diff
@@ -6735,6 +6735,18 @@ arm9:
         stack[0]: total number of selectable items
         stack[1]: number of selectable items per page
         stack[2]: ?
+    - name: SetupAndShowKeyboard
+      address:
+        EU: 0x2036AA8
+        NA: 0x20367B4
+        JP: 0x2036ACC
+      description: |-
+        Calls a function that seems to sets up info for the keyboard menu before ending with a call to ShowKeyboard.
```
UsernameFodder, line 6744:
```suggestion
        Calls a function that seems to set up info for the keyboard menu before ending with a call to ShowKeyboard.
```

symbols/arm9.yml
```diff
@@ -6743,7 +6755,7 @@ arm9:
       description: |-
         Note: unverified, ported from Irdkwia's notes
```
UsernameFodder, line 6756: Would you say this is still unverified, or can we get rid of this note?

UsernameFodder (changes requested):

symbols/arm9.yml
Adex-8x: It's fine to keep the note, I haven't really delved into that function too much to confidently say what it does

UsernameFodder: Okay sounds good.

UsernameFodder (approved):

PR #307, opened 2025-02-05
Challenge letter research

assidion2:
(Credit to Mond for finding the file the mission_rescue_bin structs are located in, as well as several of their fields)


headers/types/common/common.h
```diff
@@ -795,11 +795,13 @@ struct mission {
     int description_id;
     uint8_t unique_map_id; // 0xC: for challenges/treasure hunts/certain outlaws
     undefined field_0xd;
-    struct monster_id_16 client;                // 0xE
-    struct monster_id_16 target;                // 0x10
-    struct monster_id_16 outlaw_backup_species; // 0x12
-    struct item_id_16 item_wanted;              // 0x14
-    struct mission_reward_type_8 reward_type;   // 0x16
+    struct monster_id_16 client; // 0xE
+    // 0x10: Repurposed for storing the second team member of a normal challenge mission
```
UsernameFodder, line 799: This wording kind of de-emphasizes the primary use of the field, which is confusing. Can we change the wording of all these comments (in `struct mission` and `struct mission_rescue_bin`) to "For challenge letter missions, ..."?

symbols/arm9.yml
```diff
@@ -10284,10 +10292,10 @@ arm9:
         NA: 0x205F0D8
         JP: 0x205F3D8
       description: |-
-        Note: unverified, ported from Irdkwia's notes
+        Gets the mission struct corresponding to a certain mission number in the player's job list.
         
-        r0: mission_id
-        return: mission
+        r0: mission id in player's job list
+        return: mission struct pointer
```
UsernameFodder, line 10298: The signature says the return type is `int`? Which is correct?

symbols/arm9.yml
```diff
@@ -10393,11 +10401,29 @@ arm9:
...
+        Given a mission struct, clears some of its fields.
         
         In particular, mission::status is set to mission_status::MISSION_STATUS_INVALID, mission::dungeon_id is set to -1, mission::floor is set to 0 and mission::reward_type is set to mission_reward_type::MISSION_REWARD_MONEY.
         
         r0: Pointer to the mission to clear
+    - name: ValidateNormalChallengeMission
+      address:
+        EU: 0x2060B80
+      description: |-
+        Validates a normal challenge letter mission by checking whether its client (leader), target (second member), and outlaw_backup_species (third member) match those of some mission_rescue_bin struct (alongside some other conditions).
```
UsernameFodder, line 10413: Can we clarify what "normal" means?

```suggestion
        Validates a normal (non-legendary) challenge letter mission by checking whether its client (leader), target (second member), and outlaw_backup_species (third member) match those of some mission_rescue_bin struct (alongside some other conditions).
```

headers/functions/arm9.h
```diff
@@ -1082,6 +1083,8 @@ void DeleteAllPossibleDungeonsList(void);
 int GenerateAllPossibleDeliverList(void);
 void DeleteAllPossibleDeliverList(void);
 void ClearMissionData(struct mission* mission);
+bool ValidateNormalChallengeMission(struct mission_rescue_bin* param_1, struct mission* mission);
```
UsernameFodder, line 1086: It seems like we could have a better name than `param_1`, given that we know what it is?

symbols/overlay29.yml
```diff
@@ -8541,6 +8555,15 @@ overlay29:
         
         r0: mission destination info pointer
         return: bool
+    - name: MissionIsActiveOnFloor
```
UsernameFodder, line 8558: Since this returns the mission pointer itself rather than a bool, I think we should use a different naming convention, e.g. `GetMissionIfActive`

UsernameFodder (changes requested):

symbols/arm9.yml
assidion2: The second one, don't know how it was documented like that

UsernameFodder (approved):

PR #308, opened 2025-02-08
Document Quiz Addr

Jawshoeuh:


UsernameFodder (commented):
Are these simple integers/pointers to ints? Or more complex structs? Or still unknown?

Jawshoeuh:
> Are these simple integers/pointers to ints? Or more complex structs? Or still unknown?

They're just ints. It may be more accurate to call the 'timer' a counter instead though. I have double checked what happens to the addresses in no$gba during the quiz.

UsernameFodder:
> > Are these simple integers/pointers to ints? Or more complex structs? Or still unknown?
> 
> 
> 
> They're just ints. It may be more accurate to call the 'timer' a counter instead though. I have double checked what happens to the addresses in no$gba during the quiz.

Okay, can you add variable declarations in the headers then?

UsernameFodder (approved):

PR #311, opened 2025-03-05
Fill in languages addresses, document TREATMENT_DATA

AnonymousRandomPerson:


symbols/overlay29.yml
```diff
@@ -9723,6 +9728,23 @@ overlay29:
         NA: 0x2
         JP: 0x2
       description: The additional amount by which belly is decreased every turn when inside walls (fractional thousandths)
+    - name: TREATMENT_DATA
```
UsernameFodder, line 9731: Can we call this `MONSTER_TREATMENT_DATA`?

UsernameFodder (changes requested):

symbols/overlay29.yml
AnonymousRandomPerson: Done.

UsernameFodder (approved):

PR #312, opened 2025-04-22
Fix anim_ctrl embed in animation + wan fragment + some animation_control fields

marius851000:
Some small fix and improvement for what I’m working on.

Note that I’m still not 100% sure of how to properly handle offset_x, but it work in my implementation.

### DIFF
--- a/headers/types/files/wan.h
+++ b/headers/types/files/wan.h
@@ -7,7 +7,7 @@ enum wan_sprite_type {
     WAN_SPRITE_PROPS_UI = 0, // for object and UI objects
     WAN_SPRITE_CHARA = 1,    // for monster sprites
     WAN_SPRITE_UNK_2 = 2,
-    WAN_SPRITE_UNK_3 = 3,
+    WAN_SPRITE_UNK_3 = 3, // use the 3D engine for rendering. Used by a few sprites in effect.bin.
 };
 
 // This is usually stored as a 16-bit integer
@@ -16,7 +16,7 @@ ENUM_16_BIT(wan_sprite_type);
 #pragma pack(pop)
 
 struct wan_animation_header {
-    void* frames;
+    struct wan_fragment** frames; // first level is a frame, pointing to a list of fragments
     struct wan_offset* frame_offsets;
     struct wan_animation_group* animations;
     uint16_t nb_animation_groups;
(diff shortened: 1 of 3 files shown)


headers/types/files/wan.h
```diff
@@ -95,4 +95,34 @@ struct wan_header {
...
+    int8_t unk1;
+    int8_t unk2;
+
+    // 2 bytes
+    int8_t offset_y : 8;
+    bool unk3 : 1;
+    bool unk4 : 1;
+    uint8_t likely_padding_1 : 3;
+    bool is_mosaic : 1;
+    uint8_t shape_indice : 2; // as used in OBJ attribute 1
```
UsernameFodder, line 110: Is this supposed to be `shape_index`?

headers/types/files/wan.h
```diff
@@ -95,4 +95,34 @@ struct wan_header {
...
+
+    // 2 bytes
+    uint16_t offset_x : 9; // The value to be used is this - 256, which can end up negative.
+    bool likely_padding_2 : 1;
+    bool unk5 : 1;
+    bool is_last : 1;
+    bool h_flip : 1;
+    bool v_flip : 1;
+    uint8_t size_indice : 2; // as used in OBJ attribute 0
```
UsernameFodder, line 119: Is this supposed to be `size_index`?

headers/types/files/wan.h
```diff
@@ -95,4 +95,34 @@ struct wan_header {
 #pragma pack(pop)
 ASSERT_SIZE(struct wan_header, 10);
 
+struct wan_fragment {
+    int16_t fragment_bytes_index; // negative mean using the previous defined fragment bytes (or to
+                                  // not update it?)
```
UsernameFodder, line 100: Let's move this comment above the field so we don't need to break the comment

UsernameFodder (changes requested):

headers/types/files/wan.h
marius851000: No. indice is to be understood what the dictionary tell me is an obsolete form of indication. I think this terminology in this context come from ppmdu_tools of psy_commando (of which I derive many of my wan knowledge).
I... would prefer to keep it like that.

headers/types/files/wan.h
marius851000: Done

headers/types/files/wan.h
UsernameFodder: I see. Okay, we'll keep it as is then

UsernameFodder (approved):

PR #318, opened 2025-05-16
Create overlay00.h + misc. functions

assidion2:


symbols/arm9.yml
```diff
@@ -9952,6 +9952,15 @@ arm9:
...
         r0: dungeon ID
+    - name: RevertGiratinaAndShaymin
+      address:
+        EU: 0x20589F0
+        NA: 0x2058674
+      description: |-
+        Reverts Giratina and Shaymin party members to their standard forms.
+        
+        r0: team member index in party
+        r1: ?
```
UsernameFodder, line 9963: The C prototype doesn't show a second parameter. Is there one or not?

symbols/overlay00.yml
```diff
@@ -15,5 +15,12 @@ overlay0:
     Likely contains supporting data and code related to the top menu.
     
     This is loaded together with overlay 1 while in the top menu. Since it's in overlay group 2 (together with overlay 10, which is another "data" overlay), this overlay probably plays a similar role. It mentions several files from the BACK folder that are known backgrounds for the top menu.
-  functions: []
+  functions:
+    - name: SelectRandomBackground
+        address:
```
UsernameFodder, line 20: These are indented too far. `address` and `description` should be at the same level as `name`

symbols/arm9.yml
assidion2: There is, fixed

UsernameFodder (approved):

PR #325, opened 2025-07-12
Corrected input of GetMonsterDisplayNameType

Chesyon:
Was listed as an entity pointer; should be a monster pointer.


UsernameFodder (changes requested):
I don't think you're in `CREDITS.txt`, can you add yourself?

UsernameFodder (approved):

PR #327, opened 2025-07-29
Document PM_ForceToPowerOff

Chesyon:


Frostbyte0x70:
Some additional context: This function can also be found in the decomps of other games, such as [here](https://github.com/pret/pokediamond/blob/31df5f825e7e579b08f1ebbd8237e73851b2586a/arm9/asm/SPI_pm.s#L572), for example.

Maybe we could also document the two functions contained within (`PM_ForceToPowerOffAsync` and `PMi_WaitBusy`), while we're at it.

symbols/arm9/libs.yml
```diff
@@ -2181,6 +2181,16 @@ libs:
...
         r0: Overlay info struct
+    - name: PM_ForceToPowerOff
+      address:
+        EU: 0x20823A4
+        NA: 0x208200C
+        JP: 0x20822F4
+      description: |-
+        Powers off the system.
+        
+        No params.
```
UsernameFodder, line 2192: The return value counts as a `param`, so remove this line

UsernameFodder (changes requested):

UsernameFodder:
> Maybe we could also document the two functions contained within (`PM_ForceToPowerOffAsync` and `PMi_WaitBusy`), while we're at it.

I have a branch that tried to correlate all the stuff from the NitroSDK to Sky, and it worked pretty well, I just never got around to porting most of it because it was a huge pain. If somebody wanted to do that for me, that'd be cool 🙂.

UsernameFodder (approved):

PR #329, opened 2025-08-07
Added AI symbols and audio structs

AnonymousRandomPerson:
Not sure if the new `EntityIsValid` in overlay 31 will cause problems by conflicting with the existing one in overlay 29. It can be renamed if needed.

### DIFF
--- a/headers/functions/arm9.h
+++ b/headers/functions/arm9.h
@@ -32,6 +32,7 @@ struct mem_arena* CreateMemArena(struct iovec* mem, uint32_t max_blocks);
 void* MemLocateSet(struct mem_arena* arena, uint32_t len, uint32_t flags);
 void MemLocateUnset(struct mem_arena* arena, void* ptr);
 int RoundUpDiv256(int x);
+int sin_4096(int x);
 bool UFixedPoint64CmpLt(int32_t x_upper, uint32_t x_lower, int32_t y_upper, uint32_t y_lower);
 int MultiplyByFixedPoint(int x, fx32_8 mult_fp);
 uint32_t UMultiplyByFixedPoint(uint32_t x, ufx32_8 mult_fp);
--- a/headers/functions/arm9.h
+++ b/headers/functions/arm9.h
@@ -32,7 +32,7 @@ struct mem_arena* CreateMemArena(struct iovec* mem, uint32_t max_blocks);
 void* MemLocateSet(struct mem_arena* arena, uint32_t len, uint32_t flags);
 void MemLocateUnset(struct mem_arena* arena, void* ptr);
 int RoundUpDiv256(int x);
-int sin_4096(int x);
+int SinAbs4096(int x);
 bool UFixedPoint64CmpLt(int32_t x_upper, uint32_t x_lower, int32_t y_upper, uint32_t y_lower);
 int MultiplyByFixedPoint(int x, fx32_8 mult_fp);
 uint32_t UMultiplyByFixedPoint(uint32_t x, ufx32_8 mult_fp);
--- a/headers/functions/overlay29.h
+++ b/headers/functions/overlay29.h
@@ -678,6 +678,7 @@ int PerformDamageSequence(struct entity* attacker, struct entity* defender, stru
                           struct damage_data* damage_out, union damage_source damage_source);
 bool CanHitWithRegularAttack(struct entity* attacker, struct entity* defender);
 bool StatusCheckerCheck(struct entity* attacker, struct move* move);
+void StatusCheckerCheckOnTarget(struct entity* attacker, struct entity* target, struct move* move);
 enum weather_id GetApparentWeather(struct entity* entity);
 void TryWeatherFormChange(struct entity* entity);
 void ActivateSportCondition(bool water_sport);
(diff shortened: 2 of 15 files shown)


headers/functions/arm9.h
```diff
@@ -32,6 +32,7 @@ struct mem_arena* CreateMemArena(struct iovec* mem, uint32_t max_blocks);
 void* MemLocateSet(struct mem_arena* arena, uint32_t len, uint32_t flags);
 void MemLocateUnset(struct mem_arena* arena, void* ptr);
 int RoundUpDiv256(int x);
+int sin_4096(int x);
```
UsernameFodder, line 35: This should follow the normal PascalCase naming convention. It's not a builtin

headers/functions/overlay29.h
```diff
@@ -360,6 +360,7 @@ void EndMuzzledStatus(struct entity* user, struct entity* target);
 void EndMiracleEyeStatus(struct entity* user, struct entity* target);
 void EndMagnetRiseStatus(struct entity* user, struct entity* target);
 bool TransferNegativeBlinkerClassStatus(struct entity* user, struct entity* target);
+bool TryEndPetrifiedOrSleepStatus(struct entity* pokemon, struct entity* target);
```
UsernameFodder, line 363: Is `user` appropriate for the first param? Be consistent with everything else

headers/functions/overlay29.h
```diff
@@ -856,6 +858,9 @@ void AddHeldItemToBag(struct monster* monster);
 void RemoveEmptyItemsInBagWrapper(void);
 void GenerateItem(struct item* item, enum item_id item_id, uint16_t quantity,
                   enum gen_item_stickiness sticky_type);
+void HandleCurvedProjectileThrow(struct entity* thrower, struct item* item,
```
UsernameFodder, line 861: Can we call the first param `user`?

headers/functions/overlay31.h
```diff
@@ -7,6 +7,7 @@ void DungeonMenuSwitch(int idx);
 void CreateStairsMenuState(struct entity* entity);
 void StairsSubheadingCallback(int window_id);
 int HandleStairsMenu(void);
+bool EntityIsValid(struct entity* entity);
```
UsernameFodder, line 10: Yes this is going to cause problems for c-of-time. See `EntityIsValidMoveEffects`, this should similarly have a unique name

symbols/arm9.yml
```diff
@@ -292,6 +292,16 @@ arm9:
         
         r0: number
         return: number // 256
+    - name: sin_4096
```
UsernameFodder, line 295: This isn't `sin(x)`, it's `sin(abs(x))` so we should name it accordingly

UsernameFodder (changes requested):

symbols/arm9.yml
AnonymousRandomPerson: Renamed to `SinAbs4096`.

headers/functions/overlay31.h
AnonymousRandomPerson: Renamed to `EntityIsValidOverlay31`.

symbols/arm9/libs.yml
```diff
@@ -2667,19 +2667,21 @@ libs:
         r0: a
         r1: b
         return: a + b
-    - name: _dgeq
+    - name: _dgr
+      aliases:
+        - _dgeq
```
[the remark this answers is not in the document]
AnonymousRandomPerson: Removed. I also removed the alias `IsAffectedByTaunt` for `IsUsableWhileTaunted`, as that was a similar case where I previously renamed an incorrect symbol.

symbols/ram.yml
```diff
@@ -829,6 +829,12 @@ ram:
         NA: 0x4
         JP: 0x4
       description: "[Runtime] The index into METRONOME_TABLE for the next usage of Metronome."
+    - name: DEFAULT_TILE_COPY
```
UsernameFodder, line 832: This needs a length

symbols/overlay31.yml
```diff
@@ -71,6 +71,19 @@ overlay31:
...
+    - name: EntityIsValidOverlay31
+      address:
+        EU:
+          - 0x2384258
+        NA:
+          - 0x2383658
+        JP:
+          - 0x23848D8
+      description: |-
+        Checks if an entity pointer points to a valid entity (not entity type 0, which represents no entity).
```
UsernameFodder, line 83: Include a reference to the main one in overlay29

symbols/arm9.yml
```diff
@@ -3394,8 +3404,6 @@ arm9:
         r0: move
         return: bool
     - name: IsUsableWhileTaunted
-      aliases:
-        - IsAffectedByTaunt
```
UsernameFodder, line 3398: I'd rather keep this one even though it's wrong. It's not an internal system function so it's more likely someone could be using it.

headers/functions/overlay29.h
```diff
@@ -677,6 +678,7 @@ int PerformDamageSequence(struct entity* attacker, struct entity* defender, stru
                           struct damage_data* damage_out, union damage_source damage_source);
 bool CanHitWithRegularAttack(struct entity* attacker, struct entity* defender);
 bool StatusCheckerCheck(struct entity* attacker, struct move* move);
+void StatusCheckerCheckOnTarget(struct entity* attacker, struct entity* target, struct move* move);
```
UsernameFodder, line 681: The description says there's a return value

UsernameFodder (changes requested):

UsernameFodder (approved):

PR #333, opened 2025-08-11
rescue.bin updates + synth.bin

HappyLappy1:
There are a handful of things I need to update/correct for rescue.bin, and in addition, I've taken a closer look at synth.bin, which seems to store data for the Croagunk Swap Shop. Here's hoping the second commit isn't as rocky as the first one...


headers/functions/arm9.h
```diff
@@ -196,6 +196,7 @@ enum move_id GetItemMoveId(enum item_id item_id);
 bool TestItemAiFlag(enum item_id item_id, enum item_flag flag);
 bool IsItemInTimeDarkness(enum item_id item_id);
 bool IsItemValidVeneer(enum item_id item_id);
+struct item_id_16 ReturnEggExclusiveItem(struct monster_id_16);
```
UsernameFodder, line 199: There's no point in using the narrowed structs in functions since params are always passed back and forth in 4-byte registers/stack entries, so it's simpler to just use the bare enums. The width only matters in struct fields.

symbols/arm9.yml
```diff
@@ -3138,12 +3146,30 @@ arm9:
...
+        Searches synth_template in SYNTH/synth.bin for a matching exclusive item.
+        r0: item_id of an exclusive item.
+        return: pointer to synth_template containing that item. 0 if no match was found.
+    - name: GetValidSynthsForSpecies
+      address:
+        EU: 0x201334C
+      description: |-
+        Fills out monster_synth_data for the specified monster, containing the synth templates that apply to said monster.
+        r0: monster_id of the monster
+        r1: Pointer to an empty monster_synth_data, to be filled by the function.
```
UsernameFodder, line 3170: Per convention in some other places, let's do

```suggestion
        r1: [output] Pointer to an empty monster_synth_data, to be filled by the function.
```

headers/types/common/enums.h
```diff
@@ -3966,16 +3966,25 @@ ENUM_8_BIT(mission_status);
...
     // mission_template_item_data_1 is ignored, and mission_template_item_data_2 is an item_id
     ITEM_CASE_SPECIFIC_ITEM = 1,
     // mission_template_item_data_1 and mission_template_item_data_2 are indices to a table
-    ITEM_CASE_USE_ITEM_TABLE = 2,
-    // Has a distinct case menu, unused in template.
-    ITEM_CASE_UNK_3 = 3,
-    // mission_template_item_data_1 and mission_template_item_data_2 are ignored
-    ITEM_CASE_NO_ITEM = 4
+    ITEM_CASE_TABLE_AVAILABLE_ITEM = 2,
+    // similar to ITEM_CASE_TABLE_ITEM, but without a check for availability. Unused in template.
```
UsernameFodder, line 3977:
```suggestion
    // similar to ITEM_CASE_TABLE_AVAILABLE_ITEM, but without a check for availability. Unused in template.
```

headers/types/common/common.h
```diff
@@ -755,6 +755,50 @@ struct adventure_log {
...
+    // Only used for the Water Float (6) and the Phione/Manaphy items (7).
+    int16_t min_balance_flag;
+};
+ASSERT_SIZE(struct synth_template, 8);
+
+// Structure of SYNTH/synth.bin file, which is used for the Croagunk Swap Shop.
+struct synth_file {
+    struct synth_template synth_templates[780];
+    // The last 4 templates appear to be empty, with 0 weight.
+    struct synth_template unused_templates[4];
```
UsernameFodder, line 798: There's no point in separating out the array just to denote the empty entries, considering they're all the same type. It's simpler to just make everything one big array, with a comment

headers/types/common/common.h
```diff
@@ -755,6 +755,50 @@ struct adventure_log {
...
+    int32_t num_items_looped;
+    // 0x4: Sum of all num_to_trade in stored templates
+    int32_t total_num_to_trade;
+    // 0x8: synth_templates with exclusive items applicable to the species
+    struct synth_template* applicable_templates[5];
+};
+ASSERT_SIZE(struct monster_synth_data, 28);
+
+struct type_synth_data {
+    // 0x0: A bitfield of which types have species exclusive items.
```
UsernameFodder, line 769: A "bitfield" has a specific meaning in C/C++ and refers to struct fields that are smaller than the width of the integer type and can be packed together with other bitfields (e.g., `uint8_t bitfield : 1` is a 1-bit bitfield). I don't think this is what you mean here? We should probably use different terminology here.

UsernameFodder (changes requested):

headers/functions/arm9.h
HappyLappy1: I thiiiiink I understand? because the params are 4-byte, the inputs don't need to be compressed. but the actual return of the function still needs to be item_id_16, or is that ALSO full width, because the output (r0) is full width?

headers/types/common/common.h
HappyLappy1: Yeah that was probably not a great choice of words. Fixed.

headers/functions/arm9.h
```diff
@@ -196,6 +196,7 @@ enum move_id GetItemMoveId(enum item_id item_id);
 bool TestItemAiFlag(enum item_id item_id, enum item_flag flag);
 bool IsItemInTimeDarkness(enum item_id item_id);
 bool IsItemValidVeneer(enum item_id item_id);
+enum item_id ReturnEggExclusiveItem(enum monster_id);
```
UsernameFodder, line 199: Still give the param a name though, e.g.

```suggestion
enum item_id ReturnEggExclusiveItem(enum monster_id monster_id);
```

headers/types/common/common.h
```diff
@@ -755,6 +755,50 @@ struct adventure_log {
...
+    int32_t num_items_looped;
+    // 0x4: Sum of all num_to_trade in stored templates
+    int32_t total_num_to_trade;
+    // 0x8: synth_templates with exclusive items applicable to the species
+    struct synth_template* applicable_templates[5];
+};
+ASSERT_SIZE(struct monster_synth_data, 28);
+
+struct type_synth_data {
+    // 0x0: A bitfield of which types can have exclusive items.
```
UsernameFodder, line 769: Still needs rewording?

UsernameFodder (changes requested):

PR #337, opened 2025-08-24
Documented some ground BG functions

AnonymousRandomPerson:
These symbols came from @DizzyEggg from the decomp.


AnonymousRandomPerson:
Just got word that this research is still in progress, so I'll reopen this when ready.

PR #338, opened 2025-08-26
Documented some ground BG functions, filled in language addresses

AnonymousRandomPerson:
The ground BG symbols came from @DizzyEggg from the decomp.

### DIFF
--- a/headers/types/ground_mode/ground_mode.h
+++ b/headers/types/ground_mode/ground_mode.h
@@ -774,4 +774,198 @@ struct swap_shop_inventory_data {
 };
 ASSERT_SIZE(struct swap_shop_inventory_data, 18132);
 
+struct ground_bg_sub_struct_4 {
+    int16_t field_0x0;
+    int16_t field_0x2;
+    const void* unk4;
+    const void* unk8;
+};
+ASSERT_SIZE(struct ground_bg_sub_struct_4, 12);
+
+struct ground_bg_sub_struct_c4 {
+    uint8_t field_0x0;
+    uint8_t field_0x1;
+    int16_t field_0x2;
+    int16_t field_0x4;
+    uint16_t field_0x6;
+    struct iovec bpa_file;
+    const struct bpa_header* unk10;
+    const void* unk14;
+    const void* unk18;
+    const void* unk1c;
+    void* unk20;
+    void* field_0x24;
+    uint32_t field_0x28;
+};
+ASSERT_SIZE(struct ground_bg_sub_struct_c4, 44);
+
+struct ground_bg_sub_struct_194 {
+    uint8_t field_0x0;
+    uint8_t field_0x1;
+    uint8_t field_0x2;
+    uint8_t field_0x3;
+    uint8_t field_0x4;
+    uint8_t field_0x5;
+    uint8_t field_0x6;
+    uint8_t field_0x7;
+    int field_0x8;
+};
+ASSERT_SIZE(struct ground_bg_sub_struct_194, 12);
+
+struct ground_bg_sub_struct_1a0 {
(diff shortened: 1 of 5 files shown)


headers/types/ground_mode/ground_mode.h
```diff
@@ -774,4 +774,198 @@ struct swap_shop_inventory_data {
 };
 ASSERT_SIZE(struct swap_shop_inventory_data, 18132);
 
+struct ground_bg_sub_struct_4 {
+    int16_t field_0x0;
+    int16_t field_0x2;
+    const void* unk4;
```
UsernameFodder, line 780: How do we know this is const? Also, unknown pointers should be `undefined*` and use the Ghidra naming convention `field_0x4`

headers/types/ground_mode/ground_mode.h
```diff
@@ -774,4 +774,198 @@ struct swap_shop_inventory_data {
...
+ASSERT_SIZE(struct ground_bg_sub_struct_4, 12);
+
+struct ground_bg_sub_struct_c4 {
+    uint8_t field_0x0;
+    uint8_t field_0x1;
+    int16_t field_0x2;
+    int16_t field_0x4;
+    uint16_t field_0x6;
+    struct iovec bpa_file;
+    const struct bpa_header* unk10;
```
UsernameFodder, line 792: This isn't unknown if we have the type. It should have a better name

headers/types/ground_mode/ground_mode.h
```diff
@@ -774,4 +774,198 @@ struct swap_shop_inventory_data {
...
+    int16_t num_layers;
+    int16_t field_0x6;
+    int16_t field_0x8;
+    int16_t field_0xa;
+    int16_t field_0xc;
+    int16_t field_0xe[2];
+    int16_t field_0x12;
+    int16_t field_0x14;
+    int16_t field_0x16;
+    void (*unk18)(void*, const void*, void*, int);
```
UsernameFodder, line 917: This should have a better name

symbols/overlay11.yml
```diff
@@ -390,6 +390,34 @@ overlay11:
         
         r0: [output] The entry
         r1: background ID
+    - name: GroundBg_Init
```
UsernameFodder, line 393: There shouldn't be underscores in the name. That's allowed for known original function names, but not ones we just make up

symbols/overlay11.yml
```diff
@@ -414,6 +442,19 @@ overlay11:
...
         r3: additional_info
+    - name: BmaLayerNrlDecompressor
+      address:
+        EU: 0x22EEC14
+        NA: 0x22EE2D4
+        JP: 0x22EF908
+      description: |-
+        Decompresses the NRL-compressed data in a .bma file.
+        
+        r0: Array to hold the decompressed data
```
UsernameFodder, line 453:
```suggestion
        r0: [output] Array to hold the decompressed data
```

UsernameFodder (changes requested):

headers/types/ground_mode/ground_mode.h
AnonymousRandomPerson: Now that you mention it, `const` isn't really needed in `pmdsky-debug`, so I'll just remove it from my changes.

headers/types/ground_mode/ground_mode.h
AnonymousRandomPerson: It's not yet known what this function does, so I don't have a better name for it. Decomp makes it easier to figure out the function parameter arguments are without knowing conceptually what the function does, and this is one of those cases.

UsernameFodder (approved):

PR #344, opened 2025-09-14
Message log research + misc.

assidion2:


symbols/overlay29.yml
```diff
@@ -8845,31 +8871,105 @@ overlay29:
...
+        EU: 0x234BBD0
+      description: Frees the heap-allocated alert_box_info struct in MESSAGE_LOG_INFO.
+    - name: SetMessageLogGroupStartFlag
+      address:
+        EU: 0x234BC24
+      description: |-
+        Sets whether the next message in the message log should be the start of a new group separated by a horizontal line.
+        
+        r0: bool
+    - name: GetGlobalPreprocessorArgs
```
UsernameFodder, line 8889: These names feel too generic for something specific to the message log. Can we add "MessageLog" in the names somehow? E.g. `GetMessageLogPreprocessorArgs`

It looks like you wanted to keep with the convention for the existing `SetPreprocessorArgsIdVal`, but since that seems to be the only one, let's just rename it too and add an alias

PR #371, opened 2026-04-02
Document bitstream and related functions

Chesyon:
This idea of bitstreams has actually [already been known since at least 2013](https://projectpokemon.org/home/docs/mystery-dungeon-nds/explorers-of-sky-save-structure-r62/) (see "Byte Storage"). Save file storage is optimized down to the bit level, so there's no unused bits. It does this using bitstreams, which are responsible for reading/writing data from/to a buffer, and keeping track of location down to a bit level. Perhaps that's not the best way to explain it, but just take a look at `CopyBitsToStream` and `CopyBitsFromStream`, and you'll see what I mean. The code at the previous Project Pokemon link also demonstrates this functionality.
Anyway, it seems most compression of information for the save file is done through bitstreams, including a bunch of functions from Irdkwia's notes that I've updated accordingly. There are dozens more of these where I still need to figure out what they do, so expect those in a future PR.


symbols/arm9.yml
```diff
@@ -8429,46 +8451,62 @@ arm9:
...
+    - name: GetProgressInfoForSave
+      aliases:
+        - CopyProgressInfoFromScratchTo
       address:
         EU: 0x204D880
         NA: 0x204D548
         JP: 0x204D8A8
       description: |-
-        Note: unverified, ported from Irdkwia's notes
+        Compresseses progress info using CopyProgressInfoToStream, writing the compressed data to the provided destination.
```
UsernameFodder:
```suggestion
        Compresses progress info using CopyProgressInfoToStream, writing the compressed data to the provided destination.
```

symbols/arm9.yml
```diff
@@ -11013,47 +11104,62 @@ arm9:
...
+    - name: GetMonsterInfoForSave
+      aliases:
+        - WriteMonsterInfoToSave
       address:
         EU: 0x2059494
         NA: 0x2059118
         JP: 0x2059414
       description: |-
-        Note: unverified, ported from Irdkwia's notes
+        Compresseses monster info from the team_member_table struct using CopyMonsterToStream, writing the compressed data to the provided destination.
```
UsernameFodder:
```suggestion
        Compresses monster info from the team_member_table struct using CopyMonsterToStream, writing the compressed data to the provided destination.
```

UsernameFodder (changes requested):

UsernameFodder (approved):

PR #378, opened 2026-05-11
Heal block duration symbols

Xaklor:
added symbols for HEAL_BLOCK_TURN_RANGE, EU and NA addresses only, no idea about JP


symbols/overlay10.yml
```diff
@@ -1918,6 +1918,18 @@ overlay10:
         JP: 0x4
       description: |-
         The turn range for the Gastro Acid status, [4, 10).
+
+      - name: HEAL_BLOCK_TURN_RANGE
+      address:
+        EU: 0x22C50CC 
+        NA: 0x22C4774
+        JP: 0x???????
```
AnonymousRandomPerson: The JP address is 0x22C5E5C.

symbols/overlay10.yml
```diff
@@ -1918,6 +1918,18 @@ overlay10:
         JP: 0x4
       description: |-
         The turn range for the Gastro Acid status, [4, 10).
+
+      - name: HEAL_BLOCK_TURN_RANGE
+      address:
```
UsernameFodder, line 1924: The indentation of the YAML here's off, make sure these are flush with the `name` key

UsernameFodder (changes requested):

symbols/overlay10.yml
Xaklor: thank you!

symbols/overlay10.yml
Xaklor: sorry about that! I've fixed it now

CREDITS.txt
```diff
@@ -36,6 +36,8 @@ other parts of the project, feel free to add yourself to this file!
 - Chritchy
 - happylappy
 - Chesyon
+- Xaklor
+
```
UsernameFodder: Always happy to see new names in this file! But let's remove the extra newline; I'd rather keep the format consistent

symbols/overlay10.yml
```diff
@@ -1918,6 +1918,18 @@ overlay10:
         JP: 0x4
       description: |-
         The turn range for the Gastro Acid status, [4, 10).
+
+    - name: HEAL_BLOCK_TURN_RANGE
+      address:
+        EU: 0x22C50CC
```
UsernameFodder:
```suggestion
        EU: 0x22C50CC
```

symbols/overlay10.yml
```diff
@@ -1918,6 +1918,18 @@ overlay10:
...
+        NA: 0x22C4774
+        JP: 0x22C5E5C
+      length:
+        EU: 0x4
+        NA: 0x4
+        JP: 0x4
+      description: |-
+        The turn range for the Heal Block status, [4, 10).
         
         type: int16_t[2]
```
UsernameFodder, line 1935: Looks like this was part of the description of the preceding symbol, so let's preserve that

UsernameFodder (changes requested):

symbols/overlay10.yml
Xaklor: okay I think I got everything, sorry about the mess

UsernameFodder: Yeah looks good. Don't worry, that's the point of review!

UsernameFodder (approved):
Thanks!

PR #381, opened 2026-05-14
Label SDK/DWC Functions

CosneyMegundal:


symbols/overlay00.yml
```diff
@@ -154,12 +934,535 @@ overlay0:
...
+    - name: Cpsi_md5_init
+      address:
+        EU: 0x22D38D0
+    - name: Cpsi_md5_calc
+      address:
+        EU: 0x22D3918
+    - name: Cpsi_md5_result
+      address:
+        EU: 0x22D39D0
+    - name: R
```
UsernameFodder: Is this really the name of this function?

symbols/overlay00.yml
```diff
@@ -168,6 +1471,803 @@ overlay0:
...
+    - name: Dwci_BackuplGetWifi
+      address:
+        EU: 0x22EE820
+    - name: Dwc_BM_Init
+      address:
+        EU: 0x22EE82C
+    - name: checkAp
+      address:
+        EU: 0x22EEB34
+    - name: init
```
UsernameFodder: This name seems really generic... I'm worried it could clash with something else that might be more deserving of the name than this one in overlay 0. Any idea what this function does?

UsernameFodder (changes requested):
Tests are failing, some cleanup is needed

symbols/overlay00.yml
CosneyMegundal: Yes. They named it R, for some reason. It's exclusively used by Cpsi_md5_result though, if it needs a rename.

symbols/overlay00.yml
CosneyMegundal: In my personal notes I add lib-specific suffixes to vague function names like this. Would "Init_DWC" work better?

CosneyMegundal:
I will work on fixing the failing tests tomorrow.

symbols/overlay00.yml
UsernameFodder: Yeah that'd be better. Or maybe `Dwc_Init` so it looks more like the other system functions

symbols/overlay00.yml
UsernameFodder: Yeah let's rename it. Not sure what to, maybe something with "helper" in it?

UsernameFodder (commented):
Feel free to add yourself to `CREDITS.txt` if you want

CosneyMegundal:
Alright, I think it's ready now. I can add headers for the new functions I've labeled, but they're pretty much just going to be whatever Ghidra decided they should be. If you'd like them nonetheless I can do that, though. 

Also, I'm currently sitting on all of overlay_02, which outside of the very first function in the overlay, is exclusively sdk/dwc functions. Should I append that to this PR, or make a second one after this clears?

symbols/overlay00.yml
CosneyMegundal: I decided to go with Init_Dwc, since I don't want anyone to think "DWC_Init" verbatim is a function, and be unable to find it. The suffix should make it more clear that this was an edit in post, rather than part of the original function name.

UsernameFodder:
> I can add headers for the new functions I've labeled, but they're pretty much just going to be whatever Ghidra decided they should be. If you'd like them nonetheless I can do that, though.

Don't worry about it; the tooling allows missing headers for this reason, so you can ignore that particular check. The other direction (anything in the headers has to be in the symbols) is required though; I see that check is also failing.

> Also, I'm currently sitting on all of overlay_02, which outside of the very first function in the overlay, is exclusively sdk/dwc functions. Should I append that to this PR, or make a second one after this clears?

Let's get this one in first. Especially because it seems like GitHub wants me to approve each test run. I think once you've made one contribution it should stop asking for my approval and just run them in future PRs, which should make things easier.

symbols/arm9/libs.yml
```diff
@@ -2188,11 +2398,14 @@ libs:
         EU: 0x207AFC4
         NA: 0x207AC2C
         JP: 0x207AF14
-    - name: OSi_DisplayExContent
```
UsernameFodder, line 2191: Even though this is a typo fix (I think?), let's still add an alias just in case anyone was relying on the old name. Don't want to break anyone's code.

CosneyMegundal:
This should be the last time...

symbols/overlay00.yml
```diff
@@ -154,12 +934,534 @@ overlay0:
...
+        EU: 0x22D38D0
+    - name: Cpsi_Md5_Calc
+      address:
+        EU: 0x22D3918
+    - name: Cpsi_Md5_Result
+      address:
+        EU: 0x22D39D0
+    - name: Cpsi_Sha1_Block_Helper
+      aliases:
+        - R
```
UsernameFodder: Aliases are resolved to real symbols by some downstream projects, so I'd prefer we keep this out of the global namespace. If you want to document it, you can put it in the description

UsernameFodder (changes requested):

symbols/overlay00.yml
CosneyMegundal: Ah ok. Just saw this.

CosneyMegundal:
If anything, the new description might fail formatting. Otherwise, this should be good.

CosneyMegundal:
Fixed the formatting, and actually got the right typo-ed function name this time.

UsernameFodder (approved):
Thanks!

PR #383, opened 2026-05-16
SDK/DWC: Overlay 2, and a bit more of Overlay 0

CosneyMegundal:
Let's try this again...

### DIFF
--- a/symbols/overlay00.yml
+++ b/symbols/overlay00.yml
@@ -3794,8 +3794,6 @@ overlay0:
     - name: Dwc_GetLastError
       address:
         EU: 0x22E0A50
-        NA: 0x22E0394
-        JP: 0x22E1B40
     - name: Dwc_GetLastErrorEx
       address:
         EU: 0x22E0A70
--- a/symbols/overlay00.yml
+++ b/symbols/overlay00.yml
@@ -3808,8 +3808,6 @@ overlay0:
     - name: Dwci_SetError
       address:
         EU: 0x22E0BA8
-        NA: 0x22E0394
-        JP: 0x22E1B40
     - name: Dwci_GetAllocateHeader
       address:
         EU: 0x22E0BC0
@@ -3918,6 +3916,8 @@ overlay0:
     - name: Dwc_CleanupInet
       address:
         EU: 0x22E1284
+        NA: 0x22E0A70
+        JP: 0x22E221C
     - name: Dwc_CleanupInetAsync
       address:
         EU: 0x22E1310
--- a/symbols/overlay00.yml
+++ b/symbols/overlay00.yml
@@ -3805,15 +3805,9 @@ overlay0:
         EU: 0x22E0B8C
     - name: Dwci_SetError
       address:
-        EU:
-          - 0x22E0BA8
-          - 0x22E1284
-        NA:
-          - 0x22E0394
-          - 0x22E0A70
-        JP:
-          - 0x22E1B40
-          - 0x22E221C
+        EU: 0x22E0BA8
+        NA: 0x22E0394
+        JP: 0x22E1B40
     - name: Dwci_GetAllocateHeader
       address:
         EU: 0x22E0BC0
--- a/symbols/overlay00.yml
+++ b/symbols/overlay00.yml
@@ -3808,6 +3808,8 @@ overlay0:
     - name: Dwci_SetError
       address:
         EU: 0x22E0BA8
+        NA: 0x22E0394
+        JP: 0x22E1B40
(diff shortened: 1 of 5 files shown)


CosneyMegundal:
Decided to rename a few functions that were a little too vague from my previous PR. overlay_00 progress is still ongoing, but overlay_02 functions should be fully documented now.

symbols/overlay00.yml
```diff
@@ -3456,6 +3913,9 @@ overlay0:
         EU: 0x22E10B0
         NA: 0x22E089C
         JP: 0x22E2048
+    - name: Dwc_CleanupInet
+      address:
+        EU: 0x22E1284
```
UsernameFodder, line 3918: Keep the NA/JP addresses from the old symbol

symbols/overlay00.yml
```diff
@@ -2345,7 +2736,7 @@ overlay0:
         EU: 0x22D80E4
         NA: 0x22D78D0
         JP: 0x22D907C
-    - name: Dwc_AC_GetApData
+    - name: Dwc_AC_GetApSpotInfo
```
UsernameFodder, line 2739: Where are you getting these names from? I assumed you had an external source for them, but the fact that you're changing some of these in a nontrivial way is making me question that assumption. Did you come up with this one yourself?

UsernameFodder (changes requested):

UsernameFodder:
Can you bump the `RELEASE_VERSION` variable in `.github/workflows/release.yml` to '0.10.3' since we're changing some of these names? It's been a while since I've bumped the version anyway, so it's probably overdue.

I also just realized we kept `SetThreadField0xB4` as an alias for `OS_SetThreadDestructor`. If we're bumping the version, we may as well delete that. Since it has a raw hex code, I consider it a placeholder and less important to keep.

symbols/overlay00.yml
CosneyMegundal: I do, I misidentified this particular symbol. They are from an external source.

symbols/overlay00.yml
CosneyMegundal: Fixed. I don't think I dropped any others, right?

symbols/overlay00.yml
CosneyMegundal: To be clear, some symbols in Explorers of Sky differ slightly from my source, likely due to differing versions of the source libraries, so I am doing some analysis. With that specific example, I found an xref that proved my prior analysis incorrect.

symbols/overlay00.yml
UsernameFodder: Not that I noticed

symbols/overlay00.yml
UsernameFodder: Okay, thanks for clarifying

UsernameFodder: I mainly ask because the community's naming convention is normal PascalCase with no underscores, so I try to reserve underscore-full conventions for system stuff like this.

symbols/overlay00.yml
```diff
@@ -3330,27 +3781,33 @@ overlay0:
...
+        EU: 0x22E0A70
+    - name: Dwc_ClearError
+      address:
+        EU: 0x22E0B6C
+    - name: Dwci_IsError
+      address:
+        EU: 0x22E0B8C
+    - name: Dwci_SetError
+      address:
+        EU: 0x22E0BA8
```
UsernameFodder, line 3808: You had the NA and JP versions before but you deleted them with your recent commit?

UsernameFodder (changes requested):

symbols/overlay00.yml
CosneyMegundal: I think I restored them on the wrong function. Dwc_GetLastError is apparently at that address in NA, so Dwci_SetError can't also be there.

UsernameFodder: The old `Dwc_CleanupInet` had the addresses [EU]0x22E0BA8, [NA]0x22E0394, [JP]0x22E1B40. Since the address [EU]0x22E0BA8 now belongs to `Dwci_SetError`, I think the other two should likewise be here, and not under `Dwc_GetLastError`?

@AnonymousRandomPerson added the NA/JP addresses. Unless those were wrong?

CosneyMegundal: hopefully I got it right this time. I don't have any additional/conflicting information about the EU/JP locations of these functions.

UsernameFodder (approved):

PR #386, opened 2026-05-19
Finishing off overlay_00...

CosneyMegundal:
This should be my last PR for a while. I may continue to check in and fix any other mistakes I've made, but this should be the last batch of new symbols for a while.

### DIFF
--- a/headers/functions/overlay00.h
+++ b/headers/functions/overlay00.h
@@ -19,7 +19,7 @@ void Soc_Startup(void);
 int Dwc_AC_Process(void);
 int CheckRCode_Gsi(int error, int cast_error);
 int socket_gsi(int domain, int family, int protocol);
-int close_gsi(int fd);
+int closesocket_gsi(int fd);
 int bind_gsi(int sockfd, const struct sockaddr_in* addr, int addr_len);
 int connect_gsi(int sockfd, const struct sockaddr_in* addr, int addr_len);
 int recv_gsi(int sockfd, void* buf, size_t size, int flags);
@@ -32,7 +32,7 @@ bool SetSockBlocking(int sockfd, bool blocking);
 int nextlongrand_gsi(uint32_t ctx);
 int longrand_gsi(void);
 void Util_RandSeed(uint32_t seed);
-int RandRangeOverlay0(int x, int y);
+int Util_RandInt(int x, int y);
 void SendPacket_Gsi(const char* identifier);
 void GpiEncodeString(const char* src, char* dst);
 
--- a/headers/functions/overlay00.h
+++ b/headers/functions/overlay00.h
@@ -8,10 +8,11 @@ int Soc_Socket(int domain, int family, int protocol);
 int Soc_Bind(int sockfd, const struct sockaddr_in* addr);
 int Soc_Connect(int sockfd, const struct sockaddr_in* addr);
 int Soc_Recv(int sockfd, void* buf, size_t size, int flags);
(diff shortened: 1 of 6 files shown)


symbols/arm9/libs.yml
```diff
@@ -4618,6 +4691,15 @@ libs:
         EU: 0x2085F6C
         NA: 0x2085BD4
         JP: 0x2085EBC
+    - name: "nan"
```
UsernameFodder: Not a fan of this name, the fact that it requires quotes isn't great. Could we come up with a better name for it?

symbols/arm9/libs.yml
```diff
@@ -4834,6 +4989,12 @@ libs:
         r1: format
         ...: variadic
         return: number of characters printed, excluding the null-terminator
+    - name: rand
```
UsernameFodder, line 4993: There's not a test for this, but I've been sort of manually enforcing it: Since `rand` and `srand` already exist in overlay 0, this name clash could cause problems for downstream projects that link against the binaries indiscriminately (i.e. `c-of-time`). For an important function like this, I'd like to avoid that. But I also think this new one in arm9 is more "important" than the one in overlay 0.

I see in overlay 0 you're also changing the names of `rand`/`srand` and keeping them as aliases, but in this case can we just omit those two aliases? It should be pretty safe since the symbols will still exist at the new addresses in arm9. But to be extra safe, let's also just bump the version again (sorry!)

UsernameFodder (changes requested):

symbols/arm9/libs.yml
CosneyMegundal: How about NotANumber?

symbols/arm9/libs.yml
CosneyMegundal: Done and done. If that'll be all, I have an omnibus to inspect. The Scotland Yard insist that I should be present while they re-investigate the evidence, for some reason...

symbols/arm9/libs.yml
UsernameFodder: Without knowing what this function does, I guess that sounds fine

headers/functions/overlay00.h
```diff
@@ -17,23 +17,23 @@ int Soc_Close(int fd);
...
-int SocketConnect(int sockfd, const struct sockaddr_in* addr, int addr_len);
-int SocketRecv(int sockfd, void* buf, size_t size, int flags);
-int SocketRecvFrom(int sockfd, void* buf, size_t size, int flags, struct sockaddr_in* addr,
-                   int addr_len);
-int SocketSend(int sockfd, const void* buf, size_t size, int flags);
-int SocketSendTo(int sockfd, const void* buf, size_t size, int flags,
-                 const struct sockaddr_in* addr, int addr_len);
-bool SocketSetBlocking(int sockfd, bool blocking);
-int DoRand(uint32_t ctx);
-int rand(void);
```
UsernameFodder, line 33: Oh, can we also move these two prototypes to the other header? Assuming they're the same

UsernameFodder (changes requested):

headers/functions/overlay00.h
CosneyMegundal: I suppose I can do that. It appears like the number and general context of the params is the same. not sure about signed vs. unsigned, if that matters

UsernameFodder: That's fine, I'll check it later to confirm

UsernameFodder (approved):

PR #391, opened 2026-05-25
Fill language addresses

AnonymousRandomPerson:


symbols/arm9/libs.yml
```diff
@@ -796,55 +796,71 @@ libs:
     - name: RetEu0x02073300
```
UsernameFodder: Can you do me a favor and rename these to use the NA addresses? That's the convention elsewhere in the repo. Don't worry about aliases

AnonymousRandomPerson: Done.

UsernameFodder (approved):

symbols/arm9/libs.yml
UsernameFodder: Thanks!

PR #394, opened 2026-05-31
Spring Cleaning: Overlay 19

HappyLappy1:
These are my old notes on overlay 19, which I've recently gone back and finished off. 

Bulk-imported and merged with my symbol import script, so I'd love feedback on how I did on that, since there will likely be more of these as I comb through the rest of my ghidra analysis.


HappyLappy1:
I imagine some of these fields need aliases, but I'd like to know which ones do and do not, given that about half of the ones that are flagged are UNK_POINTER_NA, Irdkwia's notes.

headers/data/overlay19.h
```diff
@@ -1,17 +1,34 @@
 #ifndef HEADERS_DATA_OVERLAY19_H_
 #define HEADERS_DATA_OVERLAY19_H_
 
+int16_t BAR_MIRACLE_EVENT_WEIGHTS[4];
```
UsernameFodder: This should have `extern`

headers/data/overlay19.h
```diff
@@ -1,17 +1,34 @@
...
 extern struct dungeon_id_16 BAR_UNLOCKABLE_DUNGEONS_TABLE[6];
-extern struct monster_id_16 BAR_RECRUITABLE_MONSTER_TABLE[108];
+extern struct monster_id_16 BAR_RECRUITABLE_GOURMET_MONSTER_TABLE[108];
 extern struct bar_item BAR_AVAILABLE_ITEMS[66];
+int16_t BAR_STAT_REACTION_STRINGS[3];
```
UsernameFodder: These need `extern`

headers/data/overlay19.h
```diff
@@ -1,17 +1,34 @@
...
 extern struct simple_menu_id_item BAR_SUBMENU_ITEMS_1[4];
 extern struct simple_menu_id_item BAR_SUBMENU_ITEMS_2[6];
+wchar_t BAR_NEW_LINE_CHAR;
+char BAR_ME_PLAY_TEXT_TAG[12];
+char DEBUG_MESSAGE_REQUEST_STR[23];
+char DEBUG_MESSAGE_REQUEST_AND_RETURN_STR[28];
+char DEBUG_MESSAGE_BEFORE_SET_STR[30];
+char DEBUG_MESSAGE_AFTER_SET_STR[29];
 extern undefined4 OV19_STATIC_INITIALIZER;
-
```
UsernameFodder, line 16: Keep the newline between the last declaration and the `#endif`

headers/functions/overlay19.h
```diff
@@ -2,8 +2,46 @@
 #define HEADERS_FUNCTIONS_OVERLAY19_H_
 
 struct bar_item* GetBarItem(enum item_id item_id);
-int GetRecruitableMonsterAll(void);
-int GetRecruitableMonsterList(void);
-int GetRecruitableMonsterListRestricted(void);
-
+void SetMiracleDrinkStatCountRange(int* param_1, int* param_2);
+void SetBadDrinkStatCountRange(int* param_1, int* param_2);
+unsigned short* GetBarItemResultWeights(enum item_id param_1);
```
UsernameFodder: This repo doesn't use `short` use the more explicit `(u)int16_t` instead

headers/types/ground_mode/enums.h
```diff
@@ -2254,6 +2254,110 @@ enum performance_progress_flag {
...
+    BAR_SELECT_MSG_SUSPEND = 12,
+    BAR_MSG_PRE_SUSPEND = 13,
+    BAR_SUSPEND_MSG = 14,
+    BAR_MSG_POST_SUSPEND = 15,
+    BAR_EGG_HERO_PARTNER = 16,
+    BAR_EGG_RECRUIT = 17,
+    BAR_GOURMET_HERO_PARTNER = 18,
+    BAR_GOURMET_RECRUIT = 19,
+    BAR_GOURMET_LEAVE = 20,
+    BAR_UNLOCK_DUNGEON = 21
```
UsernameFodder: Not mandatory, but if a script is doing this it'd be nice to include a trailing comma

headers/types/ground_mode/ground_mode.h
```diff
@@ -1025,4 +1027,78 @@ struct link_shop {
...
+    undefined1 unk_0x61;
+    undefined1 unk_0x62;
+    undefined1 unk_0x63;
+    int number_of_items;
+    struct item_id_16 bag_items[50];
+    struct item selected_item;
+    undefined1 unk_0xd2;
+    undefined1 unk_0xd3;
+    uint32_t team_member_index;
+    short selected_move_index;
```
UsernameFodder: Don't use `short`

symbols/overlay19.yml
```diff
@@ -23,45 +23,325 @@ overlay19:
         
         r0: item ID
         return: struct bar_item*
-    - name: GetRecruitableMonsterAll
```
UsernameFodder, line 26: This needs an alias, as does any other non-placeholder function being renamed

symbols/overlay19.yml
```diff
@@ -23,45 +23,325 @@ overlay19:
...
+        
+        return: uint16_t *
+    - name: BarGetRandomValidEggRecruit
       address:
         EU: 0x238AD04
         NA: 0x238A1C4
         JP: 0x238B724
       description: |-
-        Note: unverified, ported from Irdkwia's notes
+        Retrives a random valid egg species for an egg event at the cafe.
```
UsernameFodder:
```suggestion
        Retrieves a random valid egg species for an egg event at the cafe.
```

symbols/overlay19.yml
```diff
@@ -23,45 +23,325 @@ overlay19:
...
+        
+        r0: struct dungeon_id_8*
+        return: number of allowed dungeons
+    - name: BarTryClosePortraitBox
+      address:
+        EU: 0x238AFA8
+      description: |-
+        Seems to handle closing portrait boxes for the overlay. 
+        
+        No params or return.
```
UsernameFodder: The style convention is that the return value is counted as a param, so just "No params."

symbols/overlay19.yml
```diff
@@ -23,45 +23,325 @@ overlay19:
...
+        EU: 0x238D0DC
+      description: |-
+        Populates an item list of only valid bar items within spinda_cafe.
+        
+        No params or return.
+    - name: BarZeroInitPreprocessorFlagsAndReturnCase
+      address:
+        EU: 0x238D174
+      description: |-
+        Zero Inits many preprocessor_flag fields, but also sets the return_case for scripts to zero within spinda_cafe.
```
UsernameFodder:
```suggestion
        Zero inits many preprocessor_flag fields, but also sets the return_case for scripts to zero within spinda_cafe.
```

symbols/overlay19.yml
```diff
@@ -23,45 +23,325 @@ overlay19:
...
+        
+        No params or return.
+    - name: BarZeroInitBufferAndPreprocessorArgsZero
+      address:
+        EU: 0x238D1A4
+      description: |-
+        Zero initializes the preprocessor_args_0 and string_buffer_0 fields of spinda_cafe.
+        return: bool for if buffer was empty string
```
UsernameFodder, line 153: There should be a blank line between the description and the param block

symbols/overlay19.yml
```diff
@@ -23,45 +23,325 @@ overlay19:
...
+        Seems to set ACTOR_NPC_EVENT_MAIN to whoever the player ordered the drink for.
+        Seems to do other things before that though...
+        
+        return: undefined4
+    - name: BarZeroInitBufferAndPreprocessorArgsOne
+      address:
+        EU: 0x238D2E4
+      description: |-
+        Zero initializes the preprocessor_args_1 and string_buffer_1 fields of spinda_cafe.
+        return: bool for if buffer was empty string
```
UsernameFodder, line 168: Precede with blank line

symbols/overlay19.yml
```diff
@@ -23,45 +23,325 @@ overlay19:
...
         JP: 0x238F038
       length:
         NA: 0x8
       description: |-
-        4*0x2
-        
-        Note: unverified, ported from Irdkwia's notes
+        Weights for a "miracle drink event" to be an egg, a gourmet recruit, a dungeon unlock, or a regular miracle drink in that order.
+        By default, it's 10 weight for each.
+        type: int16_t[4]
```
UsernameFodder, line 355: The type thing is sort of a relic from before the data headers existed, so I don't mind whether or not it's there. But if it's there, it should be separated by a blank line just like function param blocks

UsernameFodder (changes requested):

headers/data/overlay19.h
HappyLappy1: Noted. that should be an easy fix for the script. (I'll also fix it by hand on this PR). Are there any data types that should NOT come with an extern? obviously all structs and enums, seems like intX_t as well and their variants. Is it just all extern? I probably should know this given how much I PR here...

headers/types/ground_mode/enums.h
HappyLappy1: A script is not doing this, no. I copied this mostly verbatim from ghidra's .h export, but I can add the comma if needed.

symbols/overlay19.yml
HappyLappy1: Right. There WERE blank lines in there, but I probably deleted one or two by mistake while inputting the descriptions by hand. It's very easy to do, unfortunately... Will fix all instances I can find.

symbols/overlay19.yml
HappyLappy1: Same as above, there *was* a blank line before human edits were made... Will fix.

headers/data/overlay19.h
HappyLappy1: Got it. That shouldn't be too hard to do. Will fix by hand this time though.

headers/functions/overlay19.h
HappyLappy1: I see. Is this a comprehensive list of all other "inexplicit" datatypes and their preferred conversions? I've had a lot of experience with getting these specific datatypes through tests in prsoa-debug...

// Pretty confident in these
(u)short -> (u)int16_t
(unsigned) byte -> (u)int8_t

// Not so sure about these ones.
(u)int -> (u)int32_t
(unsigned) undefined8 -> (u)int64_t
code -> void * 

// I'm pretty sure all of these are fine as-is, right?
void, undefined(4/2/1/), (u)char, bool, (u)int64_t, (u)int32_t, (u)int16_t, (u)int8_t, wchar_t

HappyLappy1: Also I'll fix this, by hand for now ofc, gonna have to think on the best way to programmatically handle this in the script.

symbols/overlay19.yml
HappyLappy1: Understood. For the time being I'm not going to script that, but I will ofc fix it by hand.

headers/types/ground_mode/enums.h
HappyLappy1: (I've added the commas)

symbols/overlay19.yml
HappyLappy1: Nvm there was a bug causing that. Fixed!

HappyLappy1:
I believe I've fixed everything, though I do have follow-up questions to anything I haven't resolved yet. Mostly about which types should and should not be preceded by "extern" or be converted to something more explicit.

symbols/overlay19.yml
```diff
@@ -93,36 +386,86 @@ overlay19:
...
       length:
         NA: 0x2C
-      description: "Note: unverified, ported from Irdkwia's notes"
-    - name: OVERLAY19_UNKNOWN_STRUCT__NA_238E1A4
+      description: |-
+        Text string ids for bad/miracle/good drink reactions to stat changes, in that order.
+        Ex: "Huh?[K] Whaaat?[K]" or "Shock!!!"
+        
+        type: int16_t[3]
+    - name: GOOD_DRINK_REACTION_STRINGS
```
UsernameFodder, line 408: Are the lengths for these known? If so, can we add them?

UsernameFodder (changes requested):

symbols/overlay19.yml
HappyLappy1: I've added it to everything but the raw strings. I will begrudgingly add lengths for the number of characters in each of those if you need them, but I feel less comfortable doing that because of the very obvious padding to four-align the start of each string.

headers/data/overlay19.h
UsernameFodder: It's all extern.

In C, a symbol (a variable or function) can be declared as either `static` or `extern`. `static` means it's only usable within the source file it's defined in. An `extern` declaration tells the compiler that it's defined somewhere else external to the declaration, which may or may not be in the same file.

Function declarations are `extern` by default, but can be made `static` if explicitly specified (not relevant to pmdsky-debug). Global variable declarations are `static` by default, but can be made `extern` if explicitly specified.

Since in pmdsky-debug, the headers don't actually _define_ anything (i.e. there are no function bodies and data symbols reference memory that exists somewhere else (in the game)), everything is declared as `extern`, either by default (as with functions) or explicitly (as with data). This tells compilers not to complain if code that includes the headers (like c-of-time code) references the symbols, and also what types everything is, but the actual definitions (the actual machine code or the actual data structures in memory) are associated with these declarations during the linking step.

headers/functions/overlay19.h
UsernameFodder: The conventions are mentioned in the headers README under the section titled "Use meaningful types". It's purely a convention for clarity, since the fixed-width types and undefined types are typedefs to the builtin types (like `short`), which you can see in `pmdsky.h`

HappyLappy1: I've looked over that, I think that resolves my confusion over (u)int being permissible, and leave anything of that form alone. Outside of (u)int, I'm going to script it so that "code" (ghidra's weird default type for function pointers) and undefined8 should be cast as above, since neither are actually defined in pmdsky-debug. I've only ever seen undefined8 in prsoa, but I do want to be prepared in case it shows up.

headers/types/ground_mode/ground_mode.h
UsernameFodder: This is still `short`

symbols/overlay19.yml
UsernameFodder: What do you mean padding at the start? Does code reference bytes starting at padding? Usually padding would happen at the end...

headers/types/ground_mode/ground_mode.h
HappyLappy1: Ah, my bad. I forgot to skim the types files for those

headers/functions/overlay19.h
UsernameFodder: Did I miss an instance of `code` or `undefined8` in this PR, or are you just aiming to be prepared?

Usually my go-to for an unknown function pointer would be `undefined*`, meaning a pointer to bytes of unknown meaning, which could be code. Don't use `void*` for something unknown; `void*` explicitly means something is known to be arbitrary or variadic, not an unknown (but single) type.

This repo doesn't define `undefined8` because the CPU architecture is 32-bit, so it's pretty rare to see 64-bit integers. We could define it if needed, but thus far `(u)int64_t` has been enough

HappyLappy1: None in this PR. Mostly just looking for suggestions on how to handle them if they come up, though these are real things I've run into in prsoa-debug, or in ghidra auto-analysis for "code".

symbols/overlay19.yml
HappyLappy1: I do mean at the end, bad wording on my part. between the end of the current and start of the next.

HappyLappy1: Because the start of each is four-aligned, padding is added to the end of each, which I'm not sure how I would count. I'm leaning towards "we don't touch it because it might be used for something else" on principle, even if that's unlikely.

UsernameFodder: You're already setting a length when you declare the arrays to be a certain length in the headers, whether you like it or not. The lengths in the symbol tables should correspond to those byte-lengths, as you would be from evaluating `sizeof` on the variable.

Generally for arrays, I would choose the length to ignore any padding, or include a single zero entry at the end if it makes sense for the data type, like with `char` (i.e., null-terminated strings) or some enum that can't normally be zero.

HappyLappy1: Fair enough. I've added lengths for the strings, including a null-terminated character. I'm dreading adding lengths to the python script, I might just have people (me if nobody else ends up using it) just handle lengths manually.

UsernameFodder (approved):

PR #397, opened 2026-06-04
Spring Cleaning: Swap Shop, Recycle Shop, Mission Rewards, and a few Other Things

HappyLappy1:
If there's one thing my python script can do right, it's extract functions I forgot I named... I have maybe one or two more of these PRs coming.


headers/data/arm9.h
```diff
@@ -232,6 +236,8 @@ extern void* ITEM_DATA_TABLE_PTRS[3];
 #if PMDSKY_VERSION == PMDSKY_EU
 extern char* ITEM_LANG_FILE_ARRAY[5];
 #endif
+
```
UsernameFodder: Remove the extra newline

headers/data/overlay20.h
```diff
@@ -20,5 +21,8 @@ extern struct window_params RECYCLE_WINDOW_PARAMS_10;
 extern struct window_params RECYCLE_WINDOW_PARAMS_11;
 extern struct simple_menu_id_item RECYCLE_MAIN_MENU_ITEMS_3[3];
 extern undefined4 OV20_STATIC_INITIALIZER;
-
```
UsernameFodder, line 23: Keep the newline at the end

headers/functions/arm9.h
```diff
@@ -276,6 +276,14 @@ void SetActiveKecleonShop(enum team_id team_id);
 int GetMoneyStored(void);
 void SetMoneyStored(int amount);
 void AddMoneyStored(int amount);
+enum monster_id GetEggSpecies(void);
+void SetEggSpecies(struct monster_id_16 param_1);
```
UsernameFodder: There's no point using the narrow-width enum structs as parameters when passed by value, just use the base enum

headers/functions/arm9.h
```diff
@@ -327,9 +336,36 @@ void ApplyGummiBoostsGroundMode(struct monster_id_16* monster_id, uint16_t* mons
...
+void FreeRecycleOfferItems(struct recycle_offer_items* param_1);
+enum item_id GetRecycleItemId(struct recycle_item** param_1);
+bool RecycleItemHasTradeTypePrizeTicket(struct recycle_item** param_1);
+int16_t GetRecycleItemBonusOdds(struct recycle_item** param_1);
+int CountTradedRecycleItems(struct recycle_item** param_1);
+bool RecycleShopTradeIsNonspecific(struct recycle_item** param_1);
+bool RecycleShopOfferExists(void);
+void ClearRecycleShopOffer(void);
+uint32_t GetGameStateRecycleCount(void);
+struct rank_8 GetRankForRecycleShop(void);
```
UsernameFodder: No point returning narrow-width enum structs either, just return the base enum

headers/functions/arm9.h
```diff
@@ -1199,8 +1251,14 @@ int GetSosMailCount(int param_1, bool param_2);
 bool IsMissionSuspendedAndValid(struct mission* mission);
 bool AreMissionsEquivalent(struct mission* mission1, struct mission* mission2);
 bool IsMissionValid(struct mission* mission);
+bool CheckMonsterForMissionType(enum mission_type type, int* subtype, enum monster_id target,
+                                bool param_4);
+bool CheckItemForMissionType(enum mission_type type, int* subtype, enum item_id param_3);
```
UsernameFodder: Surely `param_3` can just be named `item` or something

headers/functions/arm9.h
```diff
@@ -1221,34 +1281,63 @@ void DeleteAllPossibleDungeonsList(void);
...
+void FormatMissionHeader(char* buffer, struct mission_details* param_2);
+void FormatSpecialEpisodeMissionHeader(char* param_1);
 void AppendMissionSummary(char* main_buffer, char* temp_buffer, int window_id, int y_offset,
                           struct preprocessor_args* args, struct mission_details* details);
+void MakeMissionDetails(struct mission_details* param_1, char* buffer, undefined param_3);
+void MakeSpecialEpisodeMissionDetails(struct mission_details, undefined param_2);
+undefined PrintWonderMailSkyCode(undefined param_1, undefined param_2, undefined param_3,
+                                 undefined param_4);
+void AppendMissionObjective(char* param_1, int param_2, undefined4 param_3, uint32_t param_4,
+                            short param_5);
```
UsernameFodder: Don't use `short`

headers/functions/arm9.h
```diff
@@ -1221,34 +1281,63 @@ void DeleteAllPossibleDungeonsList(void);
...
 bool IsMonsterIllegalForMissions(enum monster_id monster_id);
 bool CanDungeonBeUsedForMission(enum dungeon_id dungeon_id);
+int8_t GetMissionRankWithCapAndModifiers(void);
+int8_t GetMissionRankWithCapAndModifiersAndCap(void);
 bool CanSendItem(enum item_id item_id, bool to_sky);
+void InitMissionReward(struct mission* param_1, struct mission_reward_data* param_2,
+                       undefined4 param_3, undefined4 param_4);
+void RollRandomItemReward(undefined4 param_1, undefined4 param_2, struct item_id_16* param_3);
+void GenerateMissionRewards(struct mission* MissionPtr, int IsCafeMission);
+undefined4 CheckDungeonMissionUnlockConditions(enum dungeon_id param_1);
```
UsernameFodder: We can probably name this param better?

headers/functions/overlay20.h
```diff
@@ -0,0 +1,24 @@
...
+void RecycleShopCloseAllWindows(void);
+undefined4 RecycleShopHandleTopScreenWindow(void);
+void RecycleShopTopScreenDrawItemStats(int param_1);
+void RecycleShopTopScreenDrawItemName(int window_id);
+void RecycleShopTopScreenDrawInfoWindow(int param_1, undefined4 param_2, undefined4 param_3,
+                                        undefined4 param_4);
+void DisplaySelectionTutorialWindow(int window_id, undefined4 param_2, undefined4 param_3,
+                                    undefined4 param_4);
+
+#endif
```
UsernameFodder: Add a trailing newline

headers/types/common/common.h
```diff
@@ -1447,4 +1344,224 @@ struct bitstream {
...
+    int8_t dialogue_box_window_id;
+    undefined field28_0xc7e;
+    undefined field29_0xc7f;
+};
+ASSERT_SIZE(struct diary_menu_main_menu_data, 3200);
+
+// mission_NPC_2 is misspelled to NCP02 in debug prints
+struct mission_result_and_client {
+    int16_t mission_result;
+    struct monster_id_16 mission_NPC_1;
```
UsernameFodder: Don't use caps in field names

headers/types/common/common.h
```diff
@@ -1447,4 +1344,224 @@ struct bitstream {
...
+    struct monster_id_16 mission_NPC_2;
+    // Padding, if these fields even exist.
+    undefined field_0x6;
+    undefined field_0x7;
+};
+ASSERT_SIZE(struct mission_result_and_client, 8);
+
+// This substruct is exactly 0x6 bytes long.
+#pragma pack(push, 2)
```
UsernameFodder: I don't think this should be necessary. What was failing?

headers/types/common/common.h
```diff
@@ -1447,4 +1344,224 @@ struct bitstream {
...
+    undefined field_0x5a;
+    undefined field_0x5b;
+    struct mission* mission_ptr;
+    struct mission_template* mission_template;
+    int8_t* rescue_mission_kind_ptr;
+};
+ASSERT_SIZE(struct mission_details, 104);
+
+// This struct is exactly 0x6 bytes long.
+#pragma pack(push, 2)
```
UsernameFodder: Definitely don't think this should be necessary

headers/types/ground_mode/ground_mode.h
```diff
@@ -1101,4 +1095,712 @@ struct spinda_cafe {
...
+
+struct recycle_unk_struct_0x84 {
+    int case_id;
+    int subcase_id;
+    int field2_0x8;
+    int field3_0xc;
+    int field4_0x10;
+    int8_t text_box_window_ids[3];
+    undefined fields_0x15[81];
+    undefined field7_0x68;
```
UsernameFodder, line 1214: Why are these separate from the condensed `fields_0x15` above?

headers/types/ground_mode/ground_mode.h
```diff
@@ -1101,4 +1095,712 @@ struct spinda_cafe {
...
+    undefined field8_0x69;
+    undefined field9_0x6a;
+    undefined field10_0x6b;
+    struct recycle_item* recycle_ite_ptr;
```
UsernameFodder:
```suggestion
    struct recycle_item* recycle_item_ptr;
```

headers/types/ground_mode/ground_mode.h
```diff
@@ -1101,4 +1095,712 @@ struct spinda_cafe {
...
+    int16_t matching_items_bag[4];
+    int16_t matching_items_storage[4];
+    int16_t num_items_to_trade;
+    undefined field15_0x82;
+    undefined field16_0x83;
+};
+ASSERT_SIZE(struct recycle_unk_struct_0x84, 132);
+
+struct recycle_unk_struct_0x340 {
```
UsernameFodder, line 1227: This definition is way too long just from the unknowns. Can we condense them into arrays?

headers/types/ground_mode/ground_mode.h
```diff
@@ -1101,4 +1095,712 @@ struct spinda_cafe {
...
+    undefined field12_0x1d;
+    undefined field13_0x1e;
+    undefined field14_0x1f;
+    struct window_extra_info window_extra_info;
+    struct window_input_ctx window_input_ctx;
+    char* option_states;
+    int simple_menu_result;
+    struct owned_item** owned_items_list_ptr;
+    int num_items_required;
+    void* function_1;
```
UsernameFodder: Are these genuinely generic (i.e., it can have different signatures at different times), or are you just not sure what the signatures are? Use `void*` for the former, `undefined*` for the latter

symbols/arm9.yml
```diff
@@ -3070,6 +3128,14 @@ arm9:
...
         r0: some struct that has species ID in it?
         return: max HP boost from exclusive items
+    - name: SwapShopFreeDoublePointer
+      address:
+        EU: 0x20115A0
+      description: |-
+        Seems to call MemFree on *param_1 and **param_1, with a safety check that the **param_1 isn't zero.
+        Only known XREF is to SwapShopInventoryManager, so a more specific purpose for this function may exist.
+        
+        r0: double pointer to free
```
UsernameFodder, line 3146: This doesn't match the headers. Which is right? I assume this is the correct one

symbols/arm9.yml
```diff
@@ -3221,6 +3287,190 @@ arm9:
...
+        return: 0 if the item is throwable, 1 if not.
+    - name: RecycleShopIsTradableItem
+      address:
+        EU: 0x20126D8
+      description: |-
+        Checks if an item is capable of being traded in the recycle shop, calling RecycleShopIsNotThrownItem.
+        Also returns if the item is a prize ticket from IsTicketItem, if r1 is not 0.
+        
+        r0: item_id
+        r1: 0 if prize tickets are tradable, 1 if not.
```
UsernameFodder, line 3435: Since this is known, please name it in the header

symbols/arm9.yml
```diff
@@ -3221,6 +3287,190 @@ arm9:
...
+        return: 0 if the item is not tradable, 1 if so.
+    - name: RecycleShopCountTradableItemsInBag
+      address:
+        EU: 0x2012704
+      description: |-
+        Count the number of tradable items in the current bag with RecycleShopIsTradableItem.
+        
+        r0: 0 if prize tickets are tradable, 1 if not.
```
UsernameFodder, line 3443: Name the param in the header

symbols/arm9.yml
```diff
@@ -3221,6 +3287,190 @@ arm9:
...
+        return: number of tradable items
+    - name: RecycleShopCountTradableItemsInStorage
+      address:
+        EU: 0x2012770
+      description: |-
+        Count the number of tradable items in storage with RecycleShopIsTradableItem.
+        
+        r0: 0 if prize tickets are tradable, 1 if not.
```
UsernameFodder, line 3451: Name the param in the header

symbols/arm9.yml
```diff
@@ -3221,6 +3287,190 @@ arm9:
...
+        return: number of valid recycle shop items
+    - name: IsRecycleTradeTypeValid
+      address:
+        EU: 0x20129FC
+      description: |-
+        Uses the following logic to check if a recycle trade type fits a trade filter.
+        return not ((filter != type) && ((filter != 0 || (type == 2))))
+        
+        r0: trade filter. 2 to whitelist offers, 1 to blacklist offers, 0 to blacklist offer and prize ticket trade types.
+        r1: trade_type
```
UsernameFodder, line 3473: Name the params in the header

symbols/arm9.yml
```diff
@@ -8858,6 +9179,15 @@ arm9:
...
         
         Returns: True for DUNGEON_CRYSTAL_LAKE and DUNGEON_5TH_STATION_CLEARING, as well as for DUNGEON_DEEP_STAR_CAVE_TEAM_ROGUE if the ground variable SIDE01_BOSS2ND is 0; false otherwise.
+    - name: MissionRewardValidateDungeonId
+      address:
+        EU: 0x204F490
+      description: |-
+        Checks script variables $DUNGEON_ENTER and $DUNGEON_ENTER_INDEX for "mission dungeon" and a valid non-dojo dungeon respectively.
+        If so, stores the dungeon id from $DUNGEON_ENTER_INDEX to a pointer.
+        
+        r0: dungeon_id_8* (return)
```
UsernameFodder: The convention in this repo for output pattern is

```suggestion
        r0: [output] dungeon_id_8*
```

symbols/arm9.yml
```diff
@@ -10831,6 +11195,16 @@ arm9:
...
         return: ground monster pointer
+    - name: GetRecruitMentryIdBySpecies
+      address:
+        EU: 0x205594C
+      description: |-
+        Seems to retrieve a regular recruit in the mentry's id with a specified monster_id, looping through the entire mentry starting from slot 5.
+        The second param can be used to skip a specified number of matches, though this appears to be zero in both known calls to this function.
+        
+        r0: monster_id
+        r1: number of target species to skip in the mentry
```
UsernameFodder, line 11241: Name the param in the header

symbols/arm9.yml
```diff
@@ -11487,6 +11891,31 @@ arm9:
...
+        EU: 0x205D384
+      description: |-
+        Seems to check if a monster is valid for a specific mission type and subtype.
+        Checks monster id range, some kind of legality check for purple kecleon, and if the monster is using its base form (IE: castform-sun would fail the base form check.)
+        Also checks body size for missions that use a monster, and are not MISSION_ARREST_OUTLAW, MISSION_CHALLENGE_REQUEST, or MISSION_TYPE_CONSUMABLE.
+        
+        r0: mission_type
+        r1: mission_subtype pointer
+        r2: monster_id
+        r3: 1 if the monster is allowed to be nothing, 0 if not.
```
UsernameFodder, line 11973: Name the param in the header

symbols/arm9.yml
```diff
@@ -11498,16 +11927,45 @@ arm9:
         r0: Pointer to something
         r1: Pointer to the struct where the data of the generated mission will be written to
         return: MISSION_GENERATION_SUCCESS if the mission was successfully generated, MISSION_GENERATION_FAILURE if it failed and MISSION_GENERATION_GLOBAL_FAILURE if it failed and the game shouldn't try to generate more.
+    - name: CountAndPopulateValidMissionTableMonsters
+      address:
+        EU: 0x205E328
+      description: |-
+        Reads a portion of the mission monster table from rescue.bin, counting and populating the valid monsters for the mission.
+        
+        r0: [secondary return] Pointer to the subset of valid monster_ids
```
UsernameFodder:
```suggestion
        r0: [output] Pointer to the subset of valid monster_ids
```

UsernameFodder: Name the params in the header

symbols/arm9.yml
```diff
@@ -11498,16 +11927,45 @@ arm9:
...
+        r1: Involved in the bounds of the min and max table indices.
+        return: number of valid monsters
     - name: IsMissionTypeSpecialEpisode
       address:
         EU: 0x205E5B4
         NA: 0x205E238
         JP: 0x205E538
       description: |-
-        Checks if a mission is for a Special Episode Transmission, which unlocks Special Episode 3. This specifically checks for a mission of type MISSION_SPECIAL_EPISODE and subtype 0x2.
+        Checks if a mission is for a Special Episode Transmission, which unlocks Special Episode 3. This specifically checks for a mission of type MISSION_CONSUMABLE and subtype 0x2.
```
UsernameFodder:
```suggestion
        Checks if a mission is for a Special Episode Transmission, which unlocks Special Episode 3. This specifically checks for a mission of type MISSION_TYPE_CONSUMABLE and subtype MISSION_SPECIAL_EPISODE.
```

headers/functions/arm9.h
```diff
@@ -1199,8 +1251,14 @@ int GetSosMailCount(int param_1, bool param_2);
 bool IsMissionSuspendedAndValid(struct mission* mission);
 bool AreMissionsEquivalent(struct mission* mission1, struct mission* mission2);
 bool IsMissionValid(struct mission* mission);
+bool CheckMonsterForMissionType(enum mission_type type, int* subtype, enum monster_id target,
+                                bool param_4);
+bool CheckItemForMissionType(enum mission_type type, int* subtype, enum item_id param_3);
 enum mission_generation_result GenerateMission(undefined* param_1, struct mission* mission_data);
+int CountAndPopulateValidMissionTableMonsters(undefined4* param_1, undefined4* param_2);
 bool IsMissionTypeSpecialEpisode(struct mission* mission);
+bool DoesMissionHaveTypeAndSubtype(struct mission, struct mission_type_8* type, int* subtype);
```
UsernameFodder:
```suggestion
bool DoesMissionHaveTypeAndSubtype(struct mission* mission, struct mission_type_8* type, int* subtype);
```

headers/functions/arm9.h
```diff
@@ -1199,8 +1251,14 @@ int GetSosMailCount(int param_1, bool param_2);
...
+bool AlreadyHasSimilarMission(struct mission_type_8* type, int8_t* subtype_8);
```
UsernameFodder: Should this be

```suggestion
bool AlreadyHasSimilarMission(struct mission_type_8* type, union mission_subtype* subtype_8);
```

?

headers/functions/arm9.h
```diff
@@ -1199,8 +1251,14 @@ int GetSosMailCount(int param_1, bool param_2);
 bool IsMissionSuspendedAndValid(struct mission* mission);
 bool AreMissionsEquivalent(struct mission* mission1, struct mission* mission2);
 bool IsMissionValid(struct mission* mission);
+bool CheckMonsterForMissionType(enum mission_type type, int* subtype, enum monster_id target,
```
UsernameFodder: Are you certain the subtype here and in the functions below are passed as a pointer to a 4-byte int, and not a pointer to a 1-byte value?

symbols/arm9.yml
```diff
@@ -11699,6 +12169,34 @@ arm9:
...
+    - name: GetMissionSpecificFixedRoom
+      address:
+        EU: 0x20602FC
+      description: |-
+        Returns a fixed room id for a mission based on the specified mission type and subtype.
+        For missions with multiple valid fixed rooms, uses SelectRandomFixedRoomInRange to select one at random.
+        
+        r0: mission_type
+        r1: mission_subtype
+        return: fixed room id
```
UsernameFodder: There's an `enum fixed_room_id`, should the function signature use that instead of `int`?

symbols/arm9.yml
```diff
@@ -11699,6 +12169,34 @@ arm9:
...
+    - name: SelectRandomFixedRoomInRange
+      address:
+        EU: 0x2060400
+      description: |-
+        Returns a random fixed room from a list for a specified range. Checks for "null-termination" if no ending table entry is given.
+        It seems that the TREASURE_MEMO_FIXED_ROOM_IDS table is frontloaded with 15 "easy" layouts, and only those are allowed before performance flag 9 is enabled. 
+        
+        r0: table of fixed room ids (int8_t pointer)
```
UsernameFodder: Should this be a `struct fixed_room_id_8` pointer?

HappyLappy1:
Accidentally merged something I shouldn't have yet.

symbols/arm9.yml
```diff
@@ -11733,6 +12231,16 @@ arm9:
...
         return: bool
+    - name: MatchMissionTemplateToMission
+      address:
+        EU: 0x2060CE8
+      description: |-
+        Seems to find a matching mission_template for the corresponding mission struct if one exists. 
+        Uses a table of validation functions ARM9_UNKNOWN_FUNCTION_TABLE__NA_20A3CF4, not all of which are known.
+        
+        r0: index of ARM9_UNKNOWN_FUNCTION_TABLE__NA_20A3CF4?
+        r1: mission pointer
```
UsernameFodder, line 12310: Can you fix up the header? There should be better types and names

symbols/arm9.yml
```diff
@@ -11745,20 +12253,109 @@ arm9:
...
+        r0: header buffer
+        r1: struct mission_details
+    - name: FormatSpecialEpisodeMissionHeader
+      address:
+        EU: 0x20611A0
+      description: |-
+        Constructs a "mission header" for a special episode transmission mission, shown in the job list on the bottom screen.
+        Uses a specific text string (15439 EU) to generate the header.
+        
+        r0: header buffer
```
UsernameFodder, line 12340: Name the param in the header

headers/functions/arm9.h
```diff
@@ -1221,34 +1281,63 @@ void DeleteAllPossibleDungeonsList(void);
...
                                        struct mission* mission);
+int MatchMissionTemplateToMission(struct mission* param_1, undefined4 param_2);
 void AppendMissionTitle(char* main_buffer, char* temp_buffer, struct preprocessor_args* args,
                         struct mission_details* details);
+void FormatMissionHeader(char* buffer, struct mission_details* param_2);
+void FormatSpecialEpisodeMissionHeader(char* param_1);
 void AppendMissionSummary(char* main_buffer, char* temp_buffer, int window_id, int y_offset,
                           struct preprocessor_args* args, struct mission_details* details);
+void MakeMissionDetails(struct mission_details* param_1, char* buffer, undefined param_3);
+void MakeSpecialEpisodeMissionDetails(struct mission_details, undefined param_2);
```
UsernameFodder: Give the first parameter a name

symbols/arm9.yml
```diff
@@ -11745,20 +12253,109 @@ arm9:
...
+        r0: buffer
+        r1: ?
+    - name: PrintWonderMailSkyCode
+      address:
+        EU: 0x2062044
+      description: |-
+        Seems to print a WMS code for a mission, based on context and live analysis.
+        
+        r0+: ?
+        return: ?
```
UsernameFodder: This is pointless. Just leave off the param block entirely if nothing is known.

symbols/arm9.yml
```diff
@@ -11745,20 +12253,109 @@ arm9:
...
+    - name: AppendMissionObjective
+      address:
+        EU: 0x2062144
+      description: |-
+        Generates the "Objective: ..." string for MakeMissionDetails.
+        
+        r0: buffer
```
UsernameFodder, line 12385: Add names and better types in the header

symbols/arm9.yml
```diff
@@ -11745,20 +12253,109 @@ arm9:
...
+        stack[0]: item_id_16 to embed in the string
+    - name: AppendMissionDungeonLocation
+      address:
+        EU: 0x2062358
+      description: |-
+        Generates the "Place: ..." string for MakeMissionDetails.
+        
+        r0: mission_details pointer
+        r1: buffer
+        r2: buffer size
```
UsernameFodder, line 12398: Add name in the header

symbols/arm9.yml
```diff
@@ -11745,20 +12253,109 @@ arm9:
...
+    - name: SumValidMissionCategoryWeights
+      address:
+        EU: 0x206260C
+      description: |-
+        Seems to tally up the total weights for each mission_category under a specified mission vendor.
+        
+        
+        r0: 0 for job board, 1 for outlaw board, 2 for cafe client, 3 for beach bottle
```
UsernameFodder: Can we add an enum for this?

symbols/arm9.yml
```diff
@@ -11745,20 +12253,109 @@ arm9:
...
+        
+        return: mission_template pointer
+    - name: LoadMissionTemplates
+      address:
+        EU: 0x2062B80
+      description: |-
+        Retrieves a pointer to the mission_template table from the rescue.bin file.
+        Does not check if the file has been loaded first.
+        
+        return: mission_template*
```
UsernameFodder, line 12422: Fix the type in the header

symbols/arm9.yml
```diff
@@ -11833,6 +12430,25 @@ arm9:
...
         r0: Dungeon ID
         return: True if the specified dungeon can be part of a mission
+    - name: GetMissionRankWithCapAndModifiers
+      address:
+        EU: 0x20630D8
+      description: |-
+        Calls GetMissionRank, then adds one to the rank if the mission requires escorting a client, or defeating an outlaw.
+        Before returning, caps the mission rank at 15 (9 stars).
+        
+        r0: dungeon_floor_pair pointer
```
UsernameFodder, line 12504: This completely mismatches the header? Which is right?

symbols/arm9.yml
```diff
@@ -11833,6 +12430,25 @@ arm9:
...
+        r1: mission_type
+        return: mission rank
+    - name: GetMissionRankWithCapAndModifiersAndCap
+      address:
+        EU: 0x2063118
+      description: |-
+        Calls GetMissionRankWithModifiersWithCeil, then redundantly caps the return again at 15 (9 stars).
+        
+        r0: dungeon_floor_pair pointer
```
UsernameFodder, line 12513: This also doesn't match the header. Which is right?

symbols/arm9.yml
```diff
@@ -11844,6 +12460,43 @@ arm9:
...
+        r3: undefined4
+    - name: RollRandomItemReward
+      address:
+        EU: 0x2063420
+      description: |-
+        Seems to select a random item for an item mission reward, calling RetrieveFromItemList2 until it succeeds.
+        
+        r0: undefined4
+        r1: undefined4
+        r2: mission reward item pointer
```
UsernameFodder, line 12547: Name the param in the header

headers/functions/arm9.h
```diff
@@ -1221,34 +1281,63 @@ void DeleteAllPossibleDungeonsList(void);
...
 bool IsMonsterMissionAllowedStory(enum monster_id monster_id);
 bool IsMonsterIllegalForMissions(enum monster_id monster_id);
 bool CanDungeonBeUsedForMission(enum dungeon_id dungeon_id);
+int8_t GetMissionRankWithCapAndModifiers(void);
+int8_t GetMissionRankWithCapAndModifiersAndCap(void);
 bool CanSendItem(enum item_id item_id, bool to_sky);
+void InitMissionReward(struct mission* param_1, struct mission_reward_data* param_2,
+                       undefined4 param_3, undefined4 param_4);
+void RollRandomItemReward(undefined4 param_1, undefined4 param_2, struct item_id_16* param_3);
+void GenerateMissionRewards(struct mission* MissionPtr, int IsCafeMission);
```
UsernameFodder: Param names should be `snake_case`

symbols/arm9.yml
```diff
@@ -11844,6 +12460,43 @@ arm9:
...
+        r0: mission pointer
+        r1: 1 if mission is from the cafe, 0 if not.
+    - name: CheckDungeonMissionUnlockConditions
+      address:
+        EU: 0x2063710
+      description: |-
+        Checks if a specified dungeon is eligible to be unlocked by a mission.
+        
+        r0: dungeon_id
+        return: 3 if the dungeon is not in MISSION_DUNGEON_UNLOCK_TABLE, 2 if the dungeon is open or entered, 1 if $SCENARIO_BALANCE_FLAG is not large enough, and 0 otherwise.
```
UsernameFodder, line 12563: It sounds like the return type should be an int, not `undefined4`

symbols/arm9.yml
```diff
@@ -11917,7 +12577,33 @@ arm9:
...
         
-        r0: script_entity_id
+        r0: actor_id
+    - name: SetActorEventMain
+      address:
+        EU: 0x2065EEC
+      description: |-
+        Sets ACTOR_NPC_EVENT_MAIN to a specific actor_id.
+        
+        r0: actor_id
```
UsernameFodder, line 12651: Name and type the param in the header

symbols/arm9.yml
```diff
@@ -11917,7 +12577,33 @@ arm9:
...
+    - name: SetRandomRequestNpcs1And2
+      address:
+        EU: 0x2065EFC
+      description: |-
+        Sets ACTOR_REQUEST_NPC01 and ACTOR_REQUEST_NPC02 to specific actor_ids.
+        Has no impact on ACTOR_REQUEST_NPC03.
+        
+        r0: ACTOR_REQUEST_NPC01 actor_id
```
UsernameFodder, line 12659: Name and type the params in the header

symbols/arm9.yml
```diff
@@ -11917,7 +12577,33 @@ arm9:
...
+        r1: ACTOR_REQUEST_NPC02 actor_id
+    - name: SetAllEventNpcs
+      address:
+        EU: 0x2065F28
+      description: |-
+        Sets ACTOR_EVENT_NPC01 through ACTOR_EVENT_NPC04 to specific actor_ids.
+        
+        r0: ACTOR_EVENT_NPC01 actor_id
```
UsernameFodder, line 12667: Name and type the params in the header

symbols/arm9.yml
```diff
@@ -13475,6 +14174,24 @@ arm9:
...
       description: "The prefactor to the argument of ClampedLn in the damage formula (FLV + DAMAGE_FORMULA_FLV_SHIFT), as a binary fixed-point number with 8 fraction bits (10)."
+    - name: OUTLAW_HIDEOUT_FIXED_ROOM_IDS
+      address:
+        EU: 0x20A1E80
+      length:
+        EU: 0x5
+      description: |-
+        Table of fixed room ids used by the MISSION_ARREST_OUTLAW and MISSION_OUTLAW_HIDEOUT type and subtype.
+        
+        type: uint8_t
```
UsernameFodder: This is wrong?

symbols/arm9.yml
```diff
@@ -13475,6 +14174,24 @@ arm9:
...
+    - name: CHALLENGE_NORMAL_FIXED_ROOM_IDS
+      address:
+        EU: 0x20A1E88
+      length:
+        EU: 0x5
+      description: |-
+        Table of fixed room ids used by the MISSION_CHALLENGE_REQUEST and MISSION_CHALLENGE_NORMAL type and subtype.
+        
+        type: uint8_t
```
UsernameFodder: Fix?

symbols/arm9.yml
```diff
@@ -13519,6 +14236,16 @@ arm9:
...
+    - name: TREASURE_MEMO_FIXED_ROOM_IDS
+      address:
+        EU: 0x20A1EDC
+      length:
+        EU: 0x1E
+      description: |-
+        Table of fixed room ids used by the MISSION_TREASURE_MEMO type.
+        The last 15 entries only show up after the postgame (performance flag 9).
+        
+        type: uint8_t
```
UsernameFodder: Fix?

symbols/arm9.yml
```diff
@@ -14782,6 +15513,14 @@ arm9:
...
         
         type: int16_t[30]
+    - name: MISSION_REWARD_STRUCT_MAIN_PTR
+      address:
+        EU: 0x20B0790
+      description: |-
+        Points to a mission_reward_struct_main struct, responsible for handling the overlay 11 portion of the mission reward sequence.
+        Is also used sparingly during the Spinda Cafe egg event.
+        
+        type: mission_reward_struct_main
```
UsernameFodder: This should be a pointer?

symbols/overlay10.yml
```diff
@@ -132,6 +132,20 @@ overlay10:
...
         ...: variadic
         return: number of characters printed, excluding the null-terminator
+    - name: CreateInventoryMenuOuter
+      address:
+        EU: 0x22BDDB4
+      description: |-
+        Seems to handle the creation of an inventory menu for the spinda cafe.
+        The fact that this lives in overlay_10 suggests it may serve a more general purpose
+        
+        r0: int32_t
```
UsernameFodder: `n_items` based on the header?

symbols/overlay10.yml
```diff
@@ -132,6 +132,20 @@ overlay10:
...
+        r1: int32_t
+        r2: int32_t
+        r3: uint8_t * 
+        stack[0]: void
```
UsernameFodder: `void*`. Also wondering if this is genuinely a void pointer

symbols/overlay11.yml
```diff
@@ -189,6 +189,27 @@ overlay11:
...
+        Returns control to the scripting engine for a message_menu, with case id 0.
+        
+        No params.
+    - name: ReturnScriptMenuResult
+      address:
+        EU: 0x22E77CC
+      description: |-
+        Returns control to the scripting engine for a message_menu, with a specified case_id.
+        
+        r0: case id to pass to the scripting engine
```
UsernameFodder, line 205: Name the param in the header

symbols/arm9.yml
```diff
@@ -14676,6 +15403,10 @@ arm9:
       length:
         EU: 0x14
       description: An array containing the names of each language file. Only exists in EU.
+    - name: SYNTH_BIN_FILE_PTR
+      address:
+        EU: 0x20AFFA0
```
UsernameFodder, line 15469: Add length of 4

symbols/arm9.yml
```diff
@@ -14782,6 +15513,14 @@ arm9:
         Note: unverified, ported from Irdkwia's notes
         
         type: int16_t[30]
+    - name: MISSION_REWARD_STRUCT_MAIN_PTR
+      address:
+        EU: 0x20B0790
```
UsernameFodder, line 15581: Add length of 4

symbols/overlay11.yml
```diff
@@ -1830,6 +1930,54 @@ overlay11:
       length:
         NA: 0x10
       description: Default window_params for a team_info_box.
+    - name: SWAP_SHOP_ONE_STAR_STR
+      address:
+        EU: 0x2323AD8
```
UsernameFodder, line 1935: Add lengths for all these

symbols/overlay20.yml
```diff
@@ -12,16 +12,150 @@ overlay20:
...
+        Calls RecycleShopCommonEntryPoint(0).
+        
+        return: undefined4
+    - name: RecycleShopEntryPointResume
+      address:
+        EU: 0x238AC90
+      description: |-
+        The entry point for the overlay, when called by MENU_RECYCLE_SHOP2, resuming operations of the overlay.
+        Calls RecycleShopCommonEntryPoint(0).
+
```
UsernameFodder, line 30: Remove extra newline

symbols/overlay20.yml
```diff
@@ -12,16 +12,150 @@ overlay20:
...
+      description: |-
+        The shared frame_update of both recycle shop overlay_load_entries, managing operations of the overlay each frame.
+        
+        return: undefined4
+    - name: RecycleShopCommonEntryPoint
+      address:
+        EU: 0x238BDD4
+      description: |-
+        Effectively the shared entry_point of both recycle shop overlay_load_entry, called immediately by both real entry_points to the overlay.
+
```
UsernameFodder, line 51: Remove extra newline

symbols/overlay20.yml
```diff
@@ -12,16 +12,150 @@ overlay20:
...
+        
+        
+        r0: 1 if resuming, 0 if first entry.
+        return: undefined4
+    - name: PrintWynautTextbox
+      address:
+        EU: 0x238BE80
+      description: |-
+        Seems to display a textbox with wynaut's default portrait and a specified text string id.
+
```
UsernameFodder, line 59: Remove extra newline

headers/functions/overlay20.h
```diff
@@ -0,0 +1,24 @@
+#ifndef HEADERS_FUNCTIONS_OVERLAY20_H_
+#define HEADERS_FUNCTIONS_OVERLAY20_H_
+
+undefined4 RecycleShopEntryPointFirst(void);
+undefined4 RecycleShopEntryPointResume(void);
+void RecycleShopDestructor(void);
+undefined4 RecycleShopFrameUpdate(void);
+undefined4 RecycleShopCommonEntryPoint(bool resuming_menu);
+void PrintWynautTextbox(int strind_id);
```
UsernameFodder:
```suggestion
void PrintWynautTextbox(int string_id);
```

symbols/overlay20.yml
```diff
@@ -12,16 +12,150 @@ overlay20:
...
+        Seems to display a textbox with wynaut's default portrait and a specified text string id.
+        
+        
+        r0: string_id
+    - name: PrintWynautSimpleMenu
+      address:
+        EU: 0x238BEDC
+      description: |-
+        Seems to display a simple_menu with wynaut's default portrait, a specified text string id, and a few simple_menu_id_item pointers.
+
```
UsernameFodder, line 66: Remove extra newline

symbols/overlay20.yml
```diff
@@ -12,16 +12,150 @@ overlay20:
...
+        r0: string_id
+    - name: PrintWynautSimpleMenu
+      address:
+        EU: 0x238BEDC
+      description: |-
+        Seems to display a simple_menu with wynaut's default portrait, a specified text string id, and a few simple_menu_id_item pointers.
+        
+        
+        r0: struct recycle_shop_main
+        r1: string_id
```
UsernameFodder, line 68: This should be an `int` in the header if it's known?

symbols/overlay20.yml
```diff
@@ -12,16 +12,150 @@ overlay20:
...
+      description: |-
+        Seems to generally manage the top screen window during item selection, assigning the below callbacks to some window creation function.
+        
+        return: undefined4
+    - name: RecycleShopTopScreenDrawItemStats
+      address:
+        EU: 0x238C40C
+      description: |-
+        Seems to display the trade info and stats for the item the cursor is currently on on the top screen window.
+
```
UsernameFodder, line 111: Remove extra newline

symbols/overlay20.yml
```diff
@@ -12,16 +12,150 @@ overlay20:
...
+        Seems to display the trade info and stats for the item the cursor is currently on on the top screen window.
+        
+        
+        r0: int
+    - name: RecycleShopTopScreenDrawItemName
+      address:
+        EU: 0x238C580
+      description: |-
+        Seems to print the item the cursor is currently on to the top screen window. 
+
```
UsernameFodder, line 118: Remove extra newline

symbols/overlay20.yml
```diff
@@ -12,16 +12,150 @@ overlay20:
...
+        
+        r0: int
+    - name: RecycleShopTopScreenDrawItemName
+      address:
+        EU: 0x238C580
+      description: |-
+        Seems to print the item the cursor is currently on to the top screen window. 
+        
+        
+        r0: int
```
UsernameFodder: Header says this is a window ID?

symbols/overlay20.yml
```diff
@@ -12,16 +12,150 @@ overlay20:
...
+        r1: undefined4
+        r2: undefined4
+        r3: undefined4
+    - name: DisplaySelectionTutorialWindow
+      address:
+        EU: 0x238DA60
+      description: |-
+        Displays a side window explaining how to multi-select items for non-specific recycling.
+        
+        r0: int32_t
```
UsernameFodder: Header says this is a window ID?

symbols/overlay20.yml
```diff
@@ -12,16 +12,150 @@ overlay20:
...
+        
+        
+        r0: int
+    - name: RecycleShopTopScreenDrawInfoWindow
+      address:
+        EU: 0x238C5F4
+      description: |-
+        Seems to display the bottom window in the top screen display of the recycle shop, which explains how to read the top window.
+        
+        r0: int32_t
```
UsernameFodder: Header doesn't say it, but wondering if this is also a window ID?

symbols/overlay20.yml
```diff
@@ -67,14 +201,18 @@ overlay20:
...
-    - name: OVERLAY20_UNKNOWN_TABLE__NA_238D014
+    - name: RECYCLE_SHOP_S14P02A_BGP_FILE_STR
       address:
         EU: 0x238DB54
         NA: 0x238D014
         JP: 0x238E570
       length:
-        NA: 0x14
-      description: "Note: unverified, ported from Irdkwia's notes"
+        NA: 0x11
```
UsernameFodder, line 204: Add length

symbols/overlay20.yml
```diff
@@ -209,39 +347,61 @@ overlay20:
...
         JP: 0x4
-      description: "Note: unverified, ported from Irdkwia's notes"
-    - name: OVERLAY20_UNKNOWN_POINTER__NA_238D124
+      description: |-
+        A permanent pointer to the recycle_shop_main struct.
+        
+        type: recycle_shop_main*
+    - name: RECYCLE_SHOP_UNK_STRUCT_SIZE_84H_PTR
+      aliases:
+        - OVERLAY20_UNKNOWN_POINTER__NA_238D124
```
UsernameFodder: Don't add an alias for an unknown pointer

symbols/overlay20.yml
```diff
@@ -209,39 +347,61 @@ overlay20:
...
       length:
+        EU: 0x4
         NA: 0x4
         JP: 0x4
-      description: "Note: unverified, ported from Irdkwia's notes"
-    - name: OVERLAY20_UNKNOWN_POINTER__NA_238D128
+      description: |-
+        Seems to be a pointer to a struct of unknown purpose with size 0x84.
+        
+        type: recycle_shop_struct_0x84*
```
UsernameFodder: This name doesn't match the header name. Which is right?

symbols/overlay20.yml
```diff
@@ -209,39 +347,61 @@ overlay20:
...
+    - name: RECYCLE_SHOP_UNK_STRUCT_SIZE_340H_PTR
+      aliases:
+        - OVERLAY20_UNKNOWN_POINTER__NA_238D128
```
UsernameFodder: Remove alias

symbols/overlay20.yml
```diff
@@ -209,39 +347,61 @@ overlay20:
...
       length:
+        EU: 0x4
         NA: 0x4
         JP: 0x4
-      description: "Note: unverified, ported from Irdkwia's notes"
-    - name: OVERLAY20_UNKNOWN_POINTER__NA_238D12C
+      description: |-
+        Seems to be a pointer to a struct of unknown purpose with size 0x340.
+        
+        type: recycle_shop_struct_0x340*
```
UsernameFodder: Name doesn't match header, which is right?

symbols/overlay20.yml
```diff
@@ -209,39 +347,61 @@ overlay20:
...
+    - name: RECYCLE_SHOP_UNK_STRUCT_SIZE_1CCH_PTR
+      aliases:
+        - OVERLAY20_UNKNOWN_POINTER__NA_238D12C
```
UsernameFodder: Remove alias

symbols/overlay20.yml
```diff
@@ -209,39 +347,61 @@ overlay20:
...
         JP: 0x238E68C
       length:
+        EU: 0x4
         NA: 0x4
         JP: 0x4
-      description: "Note: unverified, ported from Irdkwia's notes"
+      description: |-
+        Seems to be a pointer to a struct of unknown purpose with size 0x1CC.
+        
+        type: recycle_shop_struct_0x1cc*
```
UsernameFodder: Name doesn't match header, which is right?

symbols/overlay26.yml
```diff
@@ -11,20 +11,304 @@ overlay26:
...
+        No params.
+    - name: ExitDungeon1FrameUpdate
+      address:
+        EU: 0x238B600
+      description: |-
+        Seems to get called on exiting a dungeon. Might be exclusive to whether the dungeon is in mission mode?
+        Is a frame_update for an overlay_load_entry.
+        
+        return: undefined4
+    - name: UnkOv26EntryPoint0238b754EU
```
UsernameFodder: Put the version before the raw address value in these

```suggestion
    - name: UnkOv26EntryPointEu0238b754
```

symbols/overlay26.yml
```diff
@@ -11,20 +11,304 @@ overlay26:
...
         JP: 0x238C380
       length:
-        NA: 0x8C
+        EU: 0x6
+        NA: 0x6
+      description: |-
+        A singular mission_reward_dialogue_sequence struct for treasure hoard missions.
+        Is fed sequentially to MissionRewardFrameUpdate to either produce a textbox, or conclude the dialogue. 
+        
+        type: mission_reward_dialogue_sequence
```
UsernameFodder: These are wrong, should be arrays?

symbols/overlay26.yml
```diff
@@ -11,20 +11,304 @@ overlay26:
...
+        EU: 0xC
+        NA: 0xC
+      description: |-
+        A pair of mission_reward_dialogue_sequence structs used for missions of type 11.
+        Is fed sequentially to MissionRewardFrameUpdate to either produce a textbox, or conclude the dialogue.
+        
+        type: mission_reward_dialogue_sequence
+    - name: DEBUG_CONQUEST_COUNT_STR
+      address:
+        EU: 0x238B9EC
```
UsernameFodder, line 293: Add lengths for these

symbols/overlay26.yml
```diff
@@ -37,43 +321,47 @@ overlay26:
...
         JP: 0x4
       description: Static initializer for overlay 26.
-    - name: OVERLAY26_UNKNOWN_POINTER__NA_238AF60
+    - name: MISSION_REWARD_OVERLAY_STRUCT_PTR
       address:
         EU: 0x238BAA0
         NA: 0x238AF60
         JP: 0x238C4C0
-      length:
-        NA: 0x4
```
UsernameFodder, line 46: Keep length?

symbols/overlay26.yml
```diff
@@ -37,43 +321,47 @@ overlay26:
...
+        A pointer to a mission_reward_struct_overlay struct, responsible for the operations of part of the mission reward sequence.
+        
+        type: mission_reward_struct_overlay*
+    - name: APPLY_RULE_DUNGEON_EFFECTS_STRUCT_PTR
       address:
         EU: 0x238BAA4
         NA: 0x238AF64
         JP: 0x238C4C4
-      length:
-        NA: 0x4
```
UsernameFodder, line 54: Keep length?

symbols/overlay26.yml
```diff
@@ -37,43 +321,47 @@ overlay26:
...
+        A pointer to a struct of unknown size, responsible for the operations of the ApplyRuleDungeonEffects* overlay_load_entry.
+        
+        type: undefined*
+    - name: EXIT_DUNGEON_1_STRUCT_PTR
       address:
         EU: 0x238BAA8
         NA: 0x238AF68
         JP: 0x238C4C8
-      length:
-        NA: 0x4
```
UsernameFodder, line 62: Keep length?

symbols/overlay26.yml
```diff
@@ -37,43 +321,47 @@ overlay26:
...
-      description: "Note: unverified, ported from Irdkwia's notes"
-    - name: OVERLAY26_UNKNOWN_POINTER5__NA_238AF70
+      description: "Note: unverified, ported from Irdkwia's notes. Used exclusively by the UnkOv26EntryPoint_* structs."
+    - name: EXIT_DUNGEON_2_STRUCT_PTR
       address:
         EU: 0x238BAB0
         NA: 0x238AF70
         JP: 0x238C4D0
-      length:
-        NA: 0x4
```
UsernameFodder, line 78: Keep length?

UsernameFodder (changes requested):
There's a lot of cleanup needed here... If you want to help me out you can do a review pass on your own changes before I go through them

HappyLappy1:
Got it, will look over these when I get a chance.

headers/functions/arm9.h
HappyLappy1: Ah yeah, it is a 1 byte pointer. This particular function doesn't actually use the param, but the struct it points to definitely has it as 1 byte.

headers/functions/arm9.h
HappyLappy1: yeah that seems likely. I'll change any others I find to use the union. I forgot that existed...

headers/types/common/common.h
HappyLappy1: I'll run it on this repo to double-check, but if I recall, structs that aren't four-aligned end up failing the implicit padding check?

headers/types/ground_mode/ground_mode.h
HappyLappy1: I'll leave comment to clarify, but I believe those 81 bytes to be fully unused. Unless you'd prefer I just lump in the remaining 4 fields?

headers/types/ground_mode/ground_mode.h
HappyLappy1: They specifically point to addresses of functions. Would those be considered undefined, even if we know exactly what they are?

HappyLappy1: I don't know c well enough to tell if pointing to functions of variable signatures, lengths, etc. are generic in this context, but I feel it would be weird to claim we don't know what they are, yet to be defined.

symbols/arm9.yml
HappyLappy1: The symbol is right, will fix the header.

symbols/arm9.yml
HappyLappy1: Oh yeah, it definitely should. Will fix.

symbols/arm9.yml
HappyLappy1: yup, will fix the data types as well

symbols/arm9.yml
HappyLappy1: The symbol is right. Fixed header.

symbols/arm9.yml
HappyLappy1: The symbol is right. Fixed.

symbols/arm9.yml
HappyLappy1: the type is wrong? Or the data is wrong? I've fixed the type, but the fixed room ID looks pretty consistent to me. 0xA0-0xA4 are here in the EU version, and treated like fixed_room_id_8s. 
https://github.com/UsernameFodder/pmdsky-debug/blob/272938c724a44dade2fb060624d154d8ea91e1d2/headers/types/dungeon_mode/enums.h#L1164

symbols/overlay10.yml
HappyLappy1: It doesn't seem like it, no. Fixed.

symbols/overlay20.yml
HappyLappy1: fixed

symbols/overlay20.yml
HappyLappy1: That seems likely, all three of these are passed as arguments to the same function. param_1 is probably a window_id consistently between the three.

symbols/overlay20.yml
HappyLappy1: I'm going with the header name. It's a new struct either way, I just managed to name it two different things somehow...

headers/types/common/common.h
HappyLappy1: I appear to be mistaken, as usual. It works fine without, I've removed them.

HappyLappy1:
I think I got everything. For some reason the github comments aren't marking themselves "Outdated", but my fixes seem to have gone through.

headers/types/common/common.h
```diff
@@ -1447,4 +1344,218 @@ struct bitstream {
...
+};
+ASSERT_SIZE(struct mission_substruct_1, 28);
+
+struct mission_reward_data {
+    struct monster_id_16 client;
+    struct mission_reward_type_8 mission_reward_type;
+    undefined field2_0x3;
+    int outlaw_leader_level;
+    int amount_money;
+    undefined4 num_items_to_roll;
```
UsernameFodder: If this is a 4-byte number, just pick signed or unsigned, it doesn't matter that much.

headers/types/ground_mode/ground_mode.h
UsernameFodder: I'd rename the field to `_unused` or something

headers/types/ground_mode/ground_mode.h
```diff
@@ -1101,4 +1095,196 @@ struct spinda_cafe {
...
+    undefined field10_0x19;
+    undefined field11_0x1a;
+    undefined field12_0x1b;
+    int arm9_unk_field; //  Arm9LoadUnkFieldNa0x2029EC8
+    int is_0xd;
+    int16_t text_string_id;
+    undefined field16_0x26;
+    undefined field17_0x27;
+    int field18_0x28;
+    undefined field_0x2c[100];
```
UsernameFodder:
```suggestion
    undefined fields_0x2c[100];
```

headers/types/ground_mode/ground_mode.h
```diff
@@ -1101,4 +1095,196 @@ struct spinda_cafe {
...
+    int field119_0x90;
+    undefined field_0x94[32];
```
UsernameFodder:
```suggestion
    undefined fields_0x94[32];
```

headers/types/ground_mode/ground_mode.h
```diff
@@ -1101,4 +1095,196 @@ struct spinda_cafe {
...
+    struct window_input_ctx window_input_ctx;
+    undefined field153_0x1ac[388];
```
UsernameFodder:
```suggestion
    undefined fields_0x1ac[388];
```

headers/types/ground_mode/ground_mode.h
UsernameFodder: If you know the exact signature of the function, then define a function pointer type. Otherwise, we _don't_ know what they are specifically, hence `undefined*`

symbols/arm9.yml
```diff
@@ -2885,6 +2885,64 @@ arm9:
...
+        If no egg is present, returns zero.
+        
+        return: monster_id
+    - name: SetEggSpecies
+      address:
+        EU: 0x201083C
+      description: |-
+        Seems to set the species of the egg in Chansey Daycare to GAME_STATE_VALUES.
+        
+        r0: struct monster_id_16
```
UsernameFodder: Fix

headers/functions/arm9.h
```diff
@@ -323,13 +332,40 @@ int ApplyNectarBoostToGroundMonster(struct ground_monster* ground_monster,
 bool IsMonsterAffectedByGravelyrockGroundMode(struct ground_monster* ground_monster);
 int ApplyGravelyrockBoostToGroundMonster(struct ground_monster* ground_monster,
                                          int* attempted_iq_boost_out);
-void ApplyGummiBoostsGroundMode(struct monster_id_16* monster_id, uint16_t* monster_iq,
+void ApplyGummiBoostsGroundMode(enum monster_id* monster_id, uint16_t* monster_iq,
```
UsernameFodder: Are you sure about this change?

The reason you don't pass the narrow enum structs by value (`MyFunction(struct some_enum_16 val)`) is because the actual asm always passes parameters through 4-byte registers/stack entries, so even if the caller is using a struct, its value will get widened when it gets passed into the function.

With pointer parameters (`MyFunction(struct some_enum_16* ptr)`), it's different. In the asm, the actual _pointer_ value (the 4-byte address) is what goes in the register/stack, while the actual data the pointer points _to_ can be any width. E.g., it could be a 2-byte field within a parent structure. So with pointer parameters, you actually need to think about the data structure behind the pointer.

The same question applies to all the functions taking enums by pointer, but this one especially since it's a change to an existing signature.

symbols/arm9.yml
```diff
@@ -3221,6 +3287,190 @@ arm9:
...
+        Retrieves GAME_STATE_VALUES->recycle_count. May differ from the recycle count at ADVENTURE_LOG_PTR->nb_recycled.
+        
+        return: Recycle count
+    - name: GetRankForRecycleShop
+      address:
+        EU: 0x2011FF0
+      description: |-
+        Retrieves GAME_STATE_VALUES->rank. Seems to exclusively be used for recycle shop prize ticket unlocks, but does correlate with guild rank.
+        
+        return: rank_8
```
UsernameFodder: Fix

headers/functions/arm9.h
```diff
@@ -1155,15 +1203,19 @@ void SetActiveTeam(enum team_id team_id);
 struct team_member* GetActiveTeamMember(int roster_idx);
 int GetActiveRosterIndex(int member_idx);
 int TryAddMonsterToActiveTeam(int member_idx);
+int GetAppointedLeaderMemberIdx(void);
 void RemoveActiveMembersFromMainTeam(void);
 void SetTeamSetupHeroAndPartnerOnly(void);
 void SetTeamSetupHeroOnly(void);
 int GetPartyMembers(uint16_t* party_members);
+int GetAdventureNpcIds(enum monster_id* param_1);
+int GetUnitNpcIds(enum monster_id* param_1);
```
UsernameFodder: The headers say these are `monster_id_16` pointers. Which is it?

headers/functions/overlay10.h
```diff
@@ -12,6 +12,8 @@ bool CheckInventoryMenuField0x1A0(int window_id);
 uint8_t PopInventoryMenuField0x1A3(int window_id);
 void UpdateInventoryMenu(struct window* window);
 bool IsInventoryMenuState3(int window_id);
+void CreateInventoryMenuOuter(int n_items, int param_2, int param_3, uint8_t* param_4,
+                              undefined* param_5, update_window_fn_t* param_6, uint32_t param_7);
```
UsernameFodder: Function pointers are already pointers themselves, so `update_window_fn_t*` is a double pointer. Is that correct?

headers/functions/arm9.h
```diff
@@ -276,6 +276,14 @@ void SetActiveKecleonShop(enum team_id team_id);
 int GetMoneyStored(void);
 void SetMoneyStored(int amount);
 void AddMoneyStored(int amount);
+enum monster_id GetEggSpecies(void);
+void SetEggSpecies(enum monster_id param_1);
```
UsernameFodder: Can you go through and name all the well-typed parameters that are trivial to name in all the headers? E.g., things like enums and known structs can be named after the name of the enum/struct.

UsernameFodder (changes requested):

headers/types/ground_mode/ground_mode.h
HappyLappy1: Do we really not have one already? I could swear half of our window types contain one. Is "typedef undefined* code" sufficient?

HappyLappy1: Oh wait, exact signature? You mean like the params of the function that might live in the slot? No I don't know that.

UsernameFodder: Yes, the exact signature. You can see examples of other function pointer typedefs in the repo to see what I mean

symbols/arm9.yml
UsernameFodder: I just meant the type described didn't match the header

UsernameFodder: Oh actually I was more thinking about the fact it should've been an array. Anyway, it seems like it's fixed.

headers/functions/overlay10.h
HappyLappy1: The param is passed directly into CreateInventoryMenu. It should be a double pointer if that function is also correct.

headers/functions/arm9.h
HappyLappy1: Honestly no. I probably missed the "*" when going through and correcting the params. Reverted.

HappyLappy1:
Alright, I think I got everything. Is the naming params thing a new standard for PRs, or has it always been this way and I've never noticed?

UsernameFodder:
> Alright, I think I got everything. Is the naming params thing a new standard for PRs, or has it always been this way and I've never noticed?

It has always been this way if it's known what the parameters are, which is the case for many of them here. Usually the only cases where the params use placeholder names is when the type is undefined or a primitive type only known from type propagation. If you know it's a specific struct or enum type, you pretty much known what it is enough to name it, because what else could it be?

headers/functions/overlay10.h
UsernameFodder: `CreateInventoryMenu` doesn't have a documented `update_window_fn_t` param, so I don't know where that comes from

headers/functions/overlay11.h
```diff
@@ -158,13 +161,24 @@ int CreateTeamInfoBox(void);
...
 void CloseTopGroundMenu(void);
 int UpdateTopGroundMenu(void);
+undefined4 GroundBagOthersMenuEntryPoint(void);
+void GroundBagOthersMenuDestructor(void);
+undefined4 GroundBagOthersMenuFrameUpdate(void);
+undefined4 InitDiaryAndFaqBoardMenu(undefined4 param_1);
+void FreeDiaryAndFaqBoardMenuMenu(void);
+undefined4 DiaryAndFaqBoardMenuFrameUpdate(void);
 bool IsBagNotEmpty(void);
+bool IsGroundMonsterAtMaxLevel(struct ground_monster* param_1);
```
UsernameFodder: Name the param

headers/functions/overlay19.h
```diff
@@ -9,7 +9,7 @@ uint16_t* BarGetDrinkEventWeights(void);
 enum monster_id BarGetRandomValidEggRecruit(void);
 enum monster_id BarGetValidEggGiverFromGourmetTable(void);
 enum monster_id BarGetRandomValidRecruitFromGourmetTable(void);
-int BarGetUnlockableDungeonPtr(struct bar_dungeon_unlock** param_1);
+int BarGetUnlockableDungeonPtr(struct dungeon_unlock_entry** param_1);
```
UsernameFodder: Name the param?

UsernameFodder (changes requested):

headers/functions/overlay10.h
HappyLappy1: ...indeed it doesn't. I'm going to mark it undefined, and if I ever figure out what's going on here I'll correct it.

HappyLappy1:
Ok, that *should* be good for at least another review, if it isn't ready for a merge, ofc.

headers/functions/arm9.h
```diff
@@ -797,17 +835,21 @@ int ShowKeyboard(int menu_type, char* buffer1, int param_3, char* buffer2);
...
 char* GetCheckIqMenuSkillString(char* buf, int iq_entry_idx);
+void PlayMissionClearBgm(void);
 bool GetNotifyNote(void);
 void SetNotifyNote(bool flag);
 void InitSpecialEpisodeHero(void);
 void EventFlagBackupVeneer(void);
 void InitMainTeamAfterQuiz(void);
 void InitSpecialEpisodePartners(void);
 void InitSpecialEpisodeExtraPartner(void);
+void AssignSpecialEpisodePc(int team_member_id, struct special_episode_pc* param_2);
```
UsernameFodder: Name param_2?

symbols/arm9.yml
```diff
@@ -7546,6 +7858,15 @@ arm9:
         Implements SPECIAL_PROC_0x4 (see ScriptSpecialProcessCall).
         
         No params.
+    - name: AssignSpecialEpisodePc
+      address:
+        EU: 0x2048DE0
+      description: |-
+        Seems to be responsible for creating special episode ground_monster for special episodes.  
+        
+
```
UsernameFodder: Remove second newline

headers/functions/arm9.h
```diff
@@ -1229,34 +1291,69 @@ void DeleteAllPossibleDungeonsList(void);
...
+struct mission_template* MatchMissionTemplateToMission(undefined4 table_index,
+                                                       struct mission* mission);
 void AppendMissionTitle(char* main_buffer, char* temp_buffer, struct preprocessor_args* args,
                         struct mission_details* details);
+void FormatMissionHeader(char* buffer, struct mission_details* mission_details);
+void FormatSpecialEpisodeMissionHeader(char* buffer);
 void AppendMissionSummary(char* main_buffer, char* temp_buffer, int window_id, int y_offset,
                           struct preprocessor_args* args, struct mission_details* details);
+void MakeMissionDetails(struct mission_details* mission_details, char* buffer, undefined param_3);
+void MakeSpecialEpisodeMissionDetails(struct mission_details details, undefined param_2);
```
UsernameFodder: Is the first param type right? The description says it's a buffer? Passing `struct mission_details` by value seems suspect at any rate...

UsernameFodder (changes requested):
Hoping this is the last round

headers/functions/arm9.h
HappyLappy1: yeah that doesn't look quite right... Will check it out.

HappyLappy1: Fixed.

UsernameFodder (approved):

PR #399, opened 2026-06-11
Add NitroSDK symbols

AnonymousRandomPerson:
Symbols added to the decomp by @SethBarberee

### DIFF
--- a/symbols/overlay00.yml
+++ b/symbols/overlay00.yml
@@ -3801,9 +3801,7 @@ overlay0:
         EU: 0x22DC4B8
         NA: 0x22DBCA4
         JP: 0x22DD450
-    - name: Dwci_Auth_StartThread
-      aliases:
-        - Dwci_Auth_OS_CreateThread
+    - name: Dwci_Auth_OS_CreateThread
       address:
         EU: 0x22DC5DC
         NA: 0x22DBDC8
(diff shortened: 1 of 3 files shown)


symbols/overlay00.yml
```diff
@@ -3802,6 +3802,8 @@ overlay0:
         NA: 0x22DBCA4
         JP: 0x22DD450
     - name: Dwci_Auth_StartThread
+      aliases:
+        - Dwci_Auth_OS_CreateThread
```
SethBarberee: These 3 aliases are actually an error on my side from the original PR. I did a search and replace for StartThread with OS_CreateThread and it touched these files as well.

The current names are the correct ones.

SethBarberee (changes requested):

symbols/overlay00.yml
AnonymousRandomPerson: Gotcha, I'll replace the original names. The symbol names haven't been around that long, so maintaining compatibility should be less of an issue for those.

UsernameFodder: I don't think that's what he meant? I think what he's saying is that we shouldn't touch these symbols

SethBarberee: Sorry if it wasn't clear enough but @UsernameFodder is correct. The original names in the debug repo are correct.

AnonymousRandomPerson: Oops, I misunderstood. I reverted the symbol names and removed the aliases. I'll fix the names in the decomp as well.

UsernameFodder (approved):

PR #404, opened 2026-06-20
Duskull, Kecleon, Kangaskhan Overlays + Peripherals

HappyLappy1:


assidion2 (commented):
You can also probably document the inner function FUN_0200d398 as MaybeGetFormattedItemName

params 1-3: same as in your functions
param 4: should_color_name (bool)
param 5: space_align_non_sticky_items (bool, whether to add an extra space before non sticky items to align them with sticky ones)

symbols/overlay11.yml
```diff
@@ -1659,13 +1659,139 @@ overlay11:
...
+        
+        No params.
+    - name: LoadXatuAppraisalOverlay
+      address:
+        EU: 0x230D754
+      description: |-
+        Seems to handle loading overlay 24 for Xatu's appraisal.
+        
+        No params.
+    - name: LoadMissionRewardOverlay
```
UsernameFodder, line 1767: This needs an alias

symbols/overlay15.yml
```diff
@@ -25,43 +90,152 @@ overlay15:
...
       length:
+        EU: 0x10
         NA: 0x10
-      description: "Note: unverified, ported from Irdkwia's notes"
-    - name: BANK_WINDOW_PARAMS_2
+      description: |-
+        Seems to be fully unused. May be a leftover from Time/Darkness or Rescue Team?
+        
+        type: struct window_params
+    - name: BANK_GOLD_STATUS_WINDOW_PARAMS_2
```
UsernameFodder: Does this still need the `_2` at the end?

headers/functions/overlay15.h
```diff
@@ -0,0 +1,13 @@
+#ifndef HEADERS_FUNCTIONS_OVERLAY15_H_
+#define HEADERS_FUNCTIONS_OVERLAY15_H_
```
UsernameFodder, line 2: Add a blank line after this


```suggestion
#define HEADERS_FUNCTIONS_OVERLAY15_H_

```

headers/types/ground_mode/ground_mode.h
```diff
@@ -1287,4 +1287,202 @@ struct recycle_unk_struct_0x1cc {
...
+};
+ASSERT_SIZE(struct duskull_bank, 236);
+
+struct kecleon_shop_shared_struct {
+    bool is_purple_kec;
+    undefined field_0x1;
+    undefined field_0x2;
+    undefined field_0x3;
+    int32_t
+        main_case_id; // As the menu is better understood, a subcase should be made for this field.
```
UsernameFodder: Move the comment to the line above so there's no awkward line break

symbols/overlay11.yml
```diff
@@ -2111,6 +2237,20 @@ overlay11:
       length:
         EU: 0x8
       description: Host pointers to multiple structures used for performing internal Swap Shop checks.
+    - name: SPINDA_CAFE_OVERLAY_STATUS
+      address:
```
UsernameFodder, line 2243: Add lengths for data symbols of known types

UsernameFodder (changes requested):

symbols/overlay15.yml
HappyLappy1: Nope, that's my bad.

symbols/overlay11.yml
```diff
@@ -2111,6 +2239,24 @@ overlay11:
...
+        EU: 0x23258EC
+      length:
+        EU: 0x4
```
UsernameFodder, line 2246: `bool` has a length of 1

symbols/overlay11.yml
```diff
@@ -2111,6 +2239,24 @@ overlay11:
...
+      description: |-
+        Tracks whether or not the spinda cafe overlay is suspended. 1 if it is inactive, 0 if it is active.
+        
+        type: bool
+    - name: RECYCLE_SHOP_OVERLAY_STATUS
+      address:
+        EU: 0x23258F4
+      length:
+        EU: 0x4
```
UsernameFodder, line 2255: If `bool`,

```suggestion
        EU: 0x1
```

symbols/overlay23.yml
```diff
@@ -12,32 +12,131 @@ overlay23:
...
         JP: 0x238E8A8
       length:
+        EU: 0xC
         NA: 0xC
-      description: "Note: unverified, ported from Irdkwia's notes"
+      description: |-
+        Exclusively passed into MaybeGetUncoloredFormattedItemName and MaybeGetColoredFormattedItemName.
+        The exact purpose of this struct is unknown, but the only "true" bool in the table is for "Other" items.     
+        
+        type: bool
```
UsernameFodder: This should be an array?

symbols/arm9.yml
```diff
@@ -7914,6 +8068,162 @@ arm9:
...
+        
+        No params.
+    - name: InitUnkStorageStruct0x410
+      address:
+        EU: 0x2045714
+      description: |-
+        Allocates a currently unknown struct of size 0x410 at UNK_STORAGE_STRUCT_0X410_PTR. May be a state case structure of some kind.
+        Likely used by Kangaskhan Storage, or some other storage interaction menu.
+        
+        r0: undefined2
```
UsernameFodder: Header says this is a pointer? Which is right?

symbols/overlay11.yml
```diff
@@ -1659,13 +1659,141 @@ overlay11:
...
+        Seems to handle loading overlay 21 for the croagunk swap shop.
+        
+        No params.
+    - name: LoadKecleonShopOverlay
+      address:
+        EU: 0x230D6F8
+      description: |-
+        Seems to handle loading overlay 22 for the kecleon brothers shop. Is used by both colors of Kecleon.
+        
+        r0: 1 for purple kecleon, 0 for green kecleon.
```
UsernameFodder, line 1745: If this is confidently known, give the param a name in the header

symbols/overlay11.yml
```diff
@@ -1659,13 +1659,141 @@ overlay11:
...
+    - name: LoadKangaskhanOverlay
+      address:
+        EU: 0x230D720
+      description: |-
+        Seems to handle loading overlay 23 for Kangaskhan storage, as well as Kangaskhan rock storage. 
+        
+        r0: 2 for kangaskhan rock, 0 otherwise?
```
UsernameFodder, line 1752: The header says this is a `bool`, but a `bool` can't take the value 2. So is it an int instead?

symbols/overlay22.yml
```diff
@@ -12,7 +12,307 @@ overlay22:
...
+        No params.
+    - name: GreenKecleonGetItemNameStringByIndex
+      address:
+        EU: 0x238B180
+      description: |-
+        Retrieves the item name string for a shop item in the green kecleon shop by index.
+        
+        r0: unused
+        r1: int
+        r2: uint
```
UsernameFodder: The header says this is a `uint32_t*`, which is right?

symbols/overlay22.yml
```diff
@@ -12,7 +12,307 @@ overlay22:
...
+        No params.
+    - name: PurpleKecleonGetItemNameStringByIndex
+      address:
+        EU: 0x238B7CC
+      description: |-
+        Retrieves the item name string for a shop item in the purple kecleon shop by index.
+        
+        r0: unused
+        r1: int
+        r2: uint
```
UsernameFodder: The header says this is a `uint32_t*`, which is right?

symbols/overlay23.yml
```diff
@@ -12,32 +12,131 @@ overlay23:
...
+        No params.
+    - name: KangaskhanStorageShowDialogueAndPortraitIfNotRock
+      address:
+        EU: 0x238DC20
+      description: |-
+        Seems responsible for handling dialogue from kangaskhan.
+        Does nothing if "KANGASKHAN_STORAGE_MENU_PTR->is_kanga_rock" is true.
+        
+        r0: window_id
+        r1: preprocessor_flags pointer
```
UsernameFodder: Header says this is a value-type `struct preprocessor_flags`. Which is right? I would assume a pointer?

UsernameFodder (changes requested):

symbols/overlay11.yml
HappyLappy1: If the bool is loaded as a word, should I just make it an int? It's clearly only ever zero or one. Is there a "wide bool" type I could use instead?

UsernameFodder: In that case, just using an `int` is fine

symbols/overlay22.yml
HappyLappy1: header for both.

symbols/overlay23.yml
HappyLappy1: I think I made a similar assumption making the symbol, but it calls ShowStringIdInDialogueBox, which appears to be a preprocessor_flags value, not pointer. Will fix symbol.
https://github.com/UsernameFodder/pmdsky-debug/blob/2984c8936738463e908a884ace8a9ff5c59a7c9e/headers/functions/arm9.h#L747

HappyLappy1: Fixed.

symbols/arm9.yml
HappyLappy1: Header is right, my bad.

symbols/overlay23.yml
UsernameFodder: Oh I forgot there was a 2-byte flags struct alongside the bigger args struct. Yeah value makes sense here.

UsernameFodder (approved):
Sorry this took so long, I've been pretty busy.

PR #405, opened 2026-06-22
Some more OAM + misc.

assidion2:


headers/types/files/wan.h
```diff
@@ -103,25 +103,28 @@ ASSERT_SIZE(struct wan_header, 10);
...
+    uint8_t obj_mode : 2;
+    uint8_t likely_padding_1 : 1;
     bool is_mosaic : 1;
-    uint8_t shape_indice : 2; // as used in OBJ attribute 1
+    uint8_t shape_indice : 2; // as used in OBJ attribute 0
 
     // 2 bytes
     uint16_t offset_x : 9; // The value to be used is this - 256, which can end up negative.
     bool likely_padding_2 : 1;
-    bool unk5 : 1;
```
UsernameFodder, line 120: Can we keep this the same? I don't like this counter numbering scheme for this reason (figuring out something means there will either be gaps or the numbers will have to shift up), but since it was already this way, I'd rather prioritize keeping the names stable over removing gaps in the numbering.

UsernameFodder (changes requested):

UsernameFodder (approved):

PR #406, opened 2026-06-26
(many) more status duration symbols

Xaklor:
added status duration symbols for most undocumented status effects. all have been verified working correctly after hex editing them on an NA rom. the full list of status effects is:

- aqua ring
- embargo
- miracle eye
- magnet rise
- curse
- snatch
- leech seed
- wish
- player petrify (duration differs if the player is affected)
- petrify
- set damage (the future sight/doom desire one)
- focus energy
- cowering
- protect
- taunt
- invisible
- mirror coat
- destiny bond
- encore
- infatuated
- mobile (mobile orb)
- blinker
- cross eyed
- eyedrop
- muzzled
- conversion 2
- vital throw
- slip (slip seed)
- frozen
- burn
- poisoned
- dropeye
- badly poisoned
- confused
- constriction (whirlpool-like moves)


symbols/overlay10.yml
```diff
@@ -1996,6 +2009,71 @@ overlay10:
...
+      address:
+        EU: 0x22C50D4
+        NA: 0x22C477C
+        JP: 0x22C5E64
+      length:
+        EU: 0x4
+        NA: 0x4
+        JP: 0x4
+      description: |-
+        The turn range for the Miracle Eye status, [127, 127).
```
assidion2: The value 0x7F (or 127) actually means the status is infinite, so you can probably add a note in parentheses saying (infinite) like for the already defined sleeplessness one.

Xaklor: I did know that 0x7F/127 means infinite, but I didn't notice that sleeplessness has a note attached to it that mentions it. that's a good idea, so I'll go add it now

UsernameFodder (approved):

PR #407, opened 2026-06-28
Correct TryRecruit, some more palettes, misc.

assidion2:


symbols/arm9.yml
```diff
@@ -6159,25 +6182,29 @@ arm9:
...
         EU: 0x202A1BC
         NA: 0x2029EC8
         JP: 0x202A220
       description: |-
-        Note: unverified, ported from Irdkwia's notes
+        Loads the index of the previous menu item the user had selected on the menu referenced by the given ID.
         
-        r0: id
-    - name: Arm9StoreUnkFieldNa0x2029ED8
+        r0: menu id
```
UsernameFodder, line 6195: Is this also supposed to return the item index? (The header says there's a return value)

UsernameFodder (changes requested):

symbols/arm9.yml
assidion2: Yes, fixed

UsernameFodder (approved):

PR #412, opened 2026-07-14
LICK_PARALYZE_CHANCE Corrections

Caitemis:
Changed LICK_PARALYZE_CHANCE's description to reflect its actual value of 15%, as well as edited the description and name of DoMoveDamageParalyze10 (now DoMoveDamageParalyze15) to match.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -40,7 +40,7 @@ jobs:
       - check-headers
       - check-symbol-header-sync
     env:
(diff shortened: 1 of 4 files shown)


UsernameFodder (changes requested):
Can you bump the `RELEASE_VERSION` variable in `.github/workflows/release.yml` to `0.10.5` since we're deleting the old symbol name?

UsernameFodder (approved):

PR #416, opened 2026-08-17
Fully document PlayEffectAnimationEntity + misc.

assidion2:


symbols/overlay10.yml
```diff
@@ -160,16 +160,19 @@ overlay10:
         
         r0: sprite index in wan_table to load from
         r1: 0 for 8bpp, 1 for 4bpp
-    - name: GetEffectAnimationField0x19
+        r2: starting palette number for 4bpp
+    - name: GetEffectAnimationWanOffset
+      aliases:
+        - GetEffectAnimationField0x19
```
UsernameFodder: Don't think it's worth keeping this alias. Can you bump the version to 0.10.6 instead?

UsernameFodder (changes requested):

symbols/overlay10.yml
assidion2: Where do I do that?

UsernameFodder: In `.github/workflows/release.yml` there's an environment variable called `RELEASE_VERSION`

UsernameFodder (approved):

PR #422, opened 2026-09-12
Move function notes

itred12:
Made some observations that allowed me to semi-confidently prove that GetNbMoves and GetMovesetIdx both take a `ground_move` pointer as their previously "undefined" argument. Clarified comments around these two functions to go with it.

(also, sorry about the commits being a bit wonky– I didn't realize I branched off of my UniqueStarterMoves branch instead of main until I went to make this pr, but the branch in question was merged regardless yesterday, so I *think* it should be okay. I'll be more careful next time)


UsernameFodder (approved):
Thanks! I'm not _too_ picky about commit history since not all contributors are well versed with Git, and PRs are squash-merged anyway. But I appreciate the attention to detail! A neat history does make reviewing easier.

PR #423, opened 2026-09-12
Starter move struct

itred12:
Upon recommendation, defined a struct for the array of moves read by AddUniqueStarterMoveTo, provided its (only NA) address, and fully documented both of these, including some very interesting notes about the choice of starters given unique moves in vanilla.

To avoid being too verbose, it may be worth removing the comment listing the moves given from AddUniqueStarterMovesTo, and instead only have it attached to this array.

I think it may also be worth researching whether these moves will "stick"– as in, if they're forgotten by any method, if they can be re-learned at the Electrivire Link Shop. The assembly behind that specific part of the game is still dizzying to me, so I haven't manged to glean too much regarding that myself.


symbols/arm9.yml
```diff
@@ -15725,6 +15725,28 @@ arm9:
...
+        NA: 0x20A3258
+      description: |-
+        An array of eight unique moves given to specific starter pokemon, which their lines cannot normally learn.
+        
+        The ninth element is a terminator, containing a blank monster and move ID.
+        
+        The moves are as follows:
+        - Vulpix: Faint Attack
+        - Eevee: Flail
+        - Phanphy: Ancient Power
```
AnonymousRandomPerson: Nitpick: it's spelled "Phanpy", and "AncientPower" is one word in Gen 4.

itred12: Ah, my bad. Good catch!

symbols/arm9.yml
```diff
@@ -17015,4 +17032,4 @@ arm9:
         EU: 0x20B3B68
         NA: 0x20B3224
         JP: 0x20B4A98
-      description: "[Runtime] The current PRNG sequence number for the system PRNG. See rand for more information on how the system PRNG works."
+      description: "[Runtime] The current PRNG sequence number for the system PRNG. See rand for more information on how the system PRNG works."
```
UsernameFodder, line 17035: Keep the newline at the end of the file

symbols/arm9.yml
```diff
@@ -11556,25 +11556,18 @@ arm9:
         r0: team_member pointer
         r1: ground_monster pointer
         r2: team_member index
-    - name: AddUniqueStarterMoveTo
+    - name: AddStarterEggMove
       address:
         NA: 0x2053568
       description: |-
-        Grants certain starter Pokemon a unique move that they cannot normally learn.
+        Grants certain starter pokemon an egg move from their line at the beginning of the game, adding it to the first available slot.
```
UsernameFodder, line 11563: Nit: I won't be picky about the accent, but "Pokemon" should still be capitalized

headers/types/common/common.h
```diff
@@ -1758,4 +1758,13 @@ struct effect_control {
 };
 ASSERT_SIZE(struct effect_control, 10148);
 
+// Stores a pair of monster_id_16 to move_id_16, used for granting specific starter pokemon an egg move from their line.
```
UsernameFodder, line 1761: This line is too long, it should be broken up into multiple lines.

In the future if you aren't sure what the formatter is complaining about, you can always download the properly formatted file from the [`format` job](https://github.com/UsernameFodder/pmdsky-debug/actions/runs/34718711905/job/103620518045?pr=423) that runs if the corresponding `format-check` fails (this is true for both symbols and headers).

symbols/arm9.yml
```diff
@@ -11556,25 +11556,18 @@ arm9:
         r0: team_member pointer
         r1: ground_monster pointer
         r2: team_member index
-    - name: AddUniqueStarterMoveTo
+    - name: AddStarterEggMove
```
UsernameFodder, line 11559: Keep the old name as an alias, as suggested by the [`compat-check` job](https://github.com/UsernameFodder/pmdsky-debug/actions/runs/34718711900/job/103620465579?pr=423)

UsernameFodder (changes requested):

symbols/arm9.yml
```diff
@@ -15725,6 +15718,30 @@ arm9:
         NA: 0x4B0
         JP: 0x4B0
       description: Contains information about the sprite size and sprite file size of each monster
+    - name: STARTER_MOVES_ARRAY
```
UsernameFodder, line 15721: Per the C type, the length of this symbol is known, so please include it below the address

UsernameFodder (changes requested):

headers/types/common/common.h
itred12: Thanks! I'll keep that in mind

UsernameFodder (approved):

PR #426, opened 2026-09-18
Chansey daycare docs

itred12:
Might be a little messy, been working on this furiously for a while. Based my documentation and structure around that for existing merchants (i.e., Electivire link shop)

Will definitely return to this later


headers/types/ground_mode/enums.h
```diff
@@ -2393,6 +2393,48 @@ enum bank_subcase {
     SBANK_UPDATE_GOLD_STATUS = 13,
 };
 
+typedef enum daycare_menu_state { // Used to manage switch-cases in Chansey Daycare
```
UsernameFodder: This repo doesn't use `typedef`-style enums

headers/types/ground_mode/enums.h
```diff
@@ -2393,6 +2393,48 @@ enum bank_subcase {
...
+    DAYCARE_INIT = 0,
```
UsernameFodder, line 2397: Enum labels are global, so they should all have a distinguishing prefix in the names. E.g., something like `DAYCARE_`

symbols/overlay24.yml
```diff
@@ -99,11 +240,15 @@ overlay24:
         NA: 0x4
         JP: 0x4
       description: Static initializer for overlay 24.
-    - name: OVERLAY24_UNKNOWN_POINTER__NA_238C600
+    - name: CHANSEY_DAYCARE_STRUCT_POINTER
+      aliases:
+        - OVERLAY24_UNKNOWN_POINTER__NA_238C600
```
UsernameFodder: We don't need aliases for these obvious placeholder names. In this case the check can be ignored

headers/types/ground_mode/ground_mode.h
```diff
@@ -1503,4 +1503,523 @@ struct special_actors {
...
+    int8_t portrait_window_id; // Created by Rename Structure Field action
+    undefined field509_0x28e;
+    int unk_menu_id; // Created by Rename Structure Field action
+    undefined field511_0x293;
+    struct portrait_params portrait_params;
+    int field513_0x2a4;
+};
+ASSERT_SIZE(struct chansey_daycare, 688);
+
+#endif
```
UsernameFodder:
```suggestion
#endif

```

Add a newline at the end

headers/types/ground_mode/ground_mode.h
```diff
@@ -1503,4 +1503,523 @@ struct special_actors {
...
+    undefined field46_0x39;
+    undefined field47_0x3a;
+    undefined field48_0x3b;
+    undefined field49_0x3c;
+    undefined field50_0x3d;
+    undefined field51_0x3e;
+    undefined field52_0x3f;
+    undefined field53_0x40;
+    undefined field54_0x41;
+    char* baby_name;
```
UsernameFodder: Are you sure this is correct? Pointers are 4 bytes wide and typically aligned to a position that's a multiple of 4 (which 0x42 is not). The compiler inserts 2 bytes of padding by default so that its position is actually 0x44. Typically this means there's a mistake here, and it's why the check is failing.

A similar thing applies to `unk_buffer2`

headers/types/ground_mode/ground_mode.h
```diff
@@ -1503,4 +1503,523 @@ struct special_actors {
...
+    undefined field335_0x15d;
+    undefined field336_0x15e;
+    undefined field337_0x15f;
+    undefined field338_0x160;
+    undefined field339_0x161;
+    undefined field340_0x162;
+    undefined field341_0x163;
+    undefined field342_0x164;
+    undefined field343_0x165;
+    char* unk_buffer2; // Created by Rename Structure Field action
```
UsernameFodder: I prefer not to use ordinals for unknowns like this (1, 2, 3, ...) because when you figure out what one of them means, it's awkward to decide what to do with the rest. Usually using the field offset is cleaner

UsernameFodder (changes requested):

headers/types/ground_mode/enums.h
itred12: Ah, whoops. I grabbed it from what I filled out in Ghidra without double-checking. My bad!

headers/types/ground_mode/ground_mode.h
itred12: Hm... I'll double-check my notes.

headers/types/ground_mode/ground_mode.h
```diff
@@ -1503,4 +1503,525 @@ struct special_actors {
...
+    undefined field48_0x3b;
+    undefined field49_0x3c;
+    undefined field50_0x3d;
+    undefined field51_0x3e;
+    undefined field52_0x3f;
+    undefined field53_0x40;
+    undefined field54_0x41;
+    // The first ten characters of this buffer(?) are strncpy'd into the hatched monster's name.
+    // (In state MAYBE_GIVEEGGSTATBONUS)
+    char unk_strncpyied_into_hatched_mon_name_0x42;
```
UsernameFodder, line 1564: It sounds like this should be an array and subsume some if the fields below it?

headers/types/ground_mode/ground_mode.h
```diff
@@ -1503,4 +1503,525 @@ struct special_actors {
...
+    undefined field335_0x15d;
+    undefined field336_0x15e;
+    undefined field337_0x15f;
+    undefined field338_0x160;
+    undefined field339_0x161;
+    undefined field340_0x162;
+    undefined field341_0x163;
+    undefined field342_0x164;
+    undefined field343_0x165;
+    char* unk_keyboardrelated_char_0x166; // Created by Rename Structure Field action
```
UsernameFodder, line 1853: I still find this pointer to be suspect. Is it an inline array instead?

headers/types/ground_mode/ground_mode.h
```diff
@@ -1503,4 +1503,525 @@ struct special_actors {
...
+    // (Created by Rename Structure Field action) Potentially main window?
+    int8_t dialogue_window_id;
+    int8_t portrait_window_id; // Created by Rename Structure Field action
+    undefined field509_0x28e;
+    int unk_menu_id; // Created by Rename Structure Field action
+    undefined field511_0x293;
+    struct portrait_params portrait_params;
+    int field513_0x2a4;
+};
+ASSERT_SIZE(struct chansey_daycare, 688);
```
UsernameFodder, line 2025: This assertion is failing. I can't tell you where the mistake is though. Please double-check the fields?

UsernameFodder (changes requested):

headers/types/ground_mode/ground_mode.h
itred12: Will do

